PR #2, opened 2020-11-06
Merge COMPASS contents from MPAS-Dev/MPAS-Model/ocean/develop

xylar:
This merge brings in the full history of https://github.com/MPAS-Dev/MPAS-Model/tree/ocean/develop/testing_and_setup/compass


xylar:
@mark-petersen, I'm just hoping of something of a sanity check from you.  For example, that the first commits and last commits make sense and aren't missing anything.  And that we're not changing files we don't want to be.  Wouldn't hurt to run a test, but maybe not critical either.

mark-petersen:
@xylar this is great!!! I ran from the head of this branch. I only needed to add a tag:
```
git tag -m "test" testTag
```
and then everything worked. I tested on grizzly with `compass_0.1.11`:
```
list_testcases.py
setup_testcase.py
manage_regression_suite.py
```
and ran the nightly regression suite with the head of MPAS-O `ocean/develop`. 

Only two requests. 
1. The head has four commits from MPAS-Dev/MPAS-Model#740, would be nice to have those as a merge into the trunk here.

2. All passed nightly regression suite, except these links that still had the prepended directories:
```
git grep -in testing_and_setup
ocean/soma/32km/default/config_analysis.xml:3:  
   <add_link source_path="mpas_model" source="testing_and_setup/compass/ocean/soma/32km/default/analysis/check_particle_sampling.py" dest="check_particle_sampling.py"/>
ocean/ziso/20km/default/config_forward.xml:8:   
   <add_link source_path="mpas_model" source="testing_and_setup/compass/ocean/scripts/LIGHTparticles/make_particle_file.py" dest="make_particles.py"/>
```
if you could remove the `testing_and_setup/compass` after the merge above.

xylar:
@mark-petersen, thanks very much for testing.  I hadn't tried the `ziso` and `soma` tests, so it's great that you uncovered that issue.  I'll do a separate PR to fix those, as you suggest.

I'll add the missing commits from `ocean/develop` right now.  Not sure how those aren't included.

xylar:
@mark-petersen, I show both commits from https://github.com/MPAS-Dev/MPAS-Model/pull/740 as being the last two commits in this PR.  Could you tell me what you think is missing?  Merge commits don't get saved in the process of a `git filter` call as far as I can tell.

mark-petersen:
This is what I see:
```
git log --graph --oneline -n 12 xylar/ocean/develop
* 7496357 (HEAD -> ocean/develop, tag: testTag, xylar/ocean/develop) Move high res 5 degrees further north
* f6e7b65 Add a new version of the EC15to60 mesh
* fb0ab84 Finishes removal of gm_stratification_ratio
* 8ce3ac1 Change flag names, as discussed with Luke
*   62eea73 Merge PR #688 'xylar/ocean/update_compass_env' into ocean/develop
|\
| * b07471e Update to version 0.1.11 of the compass environment
* | d5c06d5 Change block tests to partition tests
|/
*   68469eb Merge PR #660 'xylar/ocean/fix_QU240_test' into ocean/develop
|\
| * ae9a47f Fix validation for QU240 test step
* | dbceb32 Fix bathymetry links
* | aaa6f56 Fix tabbing
* | 2a17a9b Fixes formatting
```
I just didn't understand why the last four commits are on the trunk, rather than showing a merge into the trunk. It's not super important, just nice if it's a matter of rerunning `git filter`. Maybe it left out the last two merges for some reason.

xylar:
@mark-petersen, I see. There does seem to be a fairly complicated procedure here: https://stackoverflow.com/a/23901136/7728169 that would preserved the merge commits.  I'm not sure I want to do that, particularly because I did not do that for `master` and `develop`.  Also, it would make it hard to move other branches over from `MPAS-Model` in the future.

mark-petersen (approved):
Please merge. Request above is not important enough for the trouble. Thanks!

xylar:
Thanks, @mark-petersen!

PR #4, opened 2020-11-06
Add submodules for ocean/develop and landice/develop

xylar:
This merge adds submodules for [ocean/develop](https://github.com/MPAS-Dev/MPAS-Model/tree/ocean/develop) and [landice/develop](https://github.com/MPAS-Dev/MPAS-Model/tree/landice/develop).  These submodules will be maintained as the official versions of MPAS-Ocean and MALI that COMPASS is intended to be compatible with at a given time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,10 @@
 command_history
 local.config
+
+# exclude any new worktrees added
+MPAS-Model/*
+
+# except the ones we have on purpose
+!MPAS-Model/ocean/develop
+!MPAS-Model/landice/develop
+
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "MPAS-Model/ocean/develop"]
+	path = MPAS-Model/ocean/develop
+	url = git@github.com:MPAS-Dev/MPAS-Model.git
+[submodule "MPAS-Model/landice/develop"]
+	path = MPAS-Model/landice/develop
+	url = git@github.com:MPAS-Dev/MPAS-Model.git


xylar:
@mark-petersen and @matthewhoffman, this should be really simple to review.  Just make sure you agree that the submodules are indeed set to the current `ocean/develop` and `landice/develop`.

.gitignore
```diff
@@ -1,2 +1,10 @@
 command_history
 local.config
+
+# exclude any new worktrees added
+MPAS-Model/*
```
mark-petersen, line 5: This is really nice. It didn't sink in that we could have multiple worktrees at this level within a compass repo. Thanks for thinking of this as a workflow.

mark-petersen (approved):
I tested this by compiling within the submodule,
`MPAS-Model/ocean/develop`
and altering `general.config.ocean` to point there. It all works fine.

Could you update the ocean/develop submodule to the head 53837597? We merged a nice performance improvement this morning. 

Going forward, we can see how often we want to update. If we update with MPAS namelist change PRs, I think it will be every week or two. But we are well practiced at submodules, so I think it will be smooth.

xylar:
> Could you update the ocean/develop submodule to the head 53837597? We merged a nice performance improvement this morning.

Done.

> Going forward, we can see how often we want to update. If we update with MPAS namelist change PRs, I think it will be every week or two. But we are well practiced at submodules, so I think it will be smooth.

Yeah, I don't have a feel yet for how often we need to update.  Eventually, I want to have a check in the other direction, at least for MPAS-Ocean so we can mark which `compass` version is (or versions are) compatible with a given checkout of MPAS-Ocean.  But for now I think we will just have to get a feel for it.

matthewhoffman (approved):

xylar:
@matthewhoffman and @mark-petersen, thanks again!

PR #5, opened 2020-11-06
Add the recipe for the compass conda metapackage

xylar:
This is from https://github.com/MPAS-Dev/MPAS-Tools/tree/master/compass

### DIFF
--- a/recipe/LICENSE
+++ b/recipe/LICENSE
@@ -1,0 +1,39 @@
+Copyright (c) 2013-2020,  Los Alamos National Security, LLC (LANS) (Ocean: LA-CC-13-047;
+Land Ice: LA-CC-13-117) and the University Corporation for Atmospheric Research (UCAR).
+
+All rights reserved.
+
+LANS is the operator of the Los Alamos National Laboratory under Contract No.
+DE-AC52-06NA25396 with the U.S. Department of Energy.  UCAR manages the National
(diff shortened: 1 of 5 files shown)


mark-petersen:
Thanks for saving the rest of us from these details. I need to learn more about conda packages, so this is a nice example.

I'm confused about how the conda package relates to the actual code, the repos, and the names. Right now, `compass_0.1.11` lives in MPAS-Tools. So that will be removed? Will this new compass package include `MpasMeshConverter.x` and other codes in MPAS-Tools, or will those split off?

xylar:
@mark-petersen, 

> Thanks for saving the rest of us from these details. I need to learn more about conda packages, so this is a nice example.

Sure, happy to do it.  But I think you would benefit from learning some of these details.

> I'm confused about how the conda package relates to the actual code, the repos, and the names. Right now, compass_0.1.11 lives in MPAS-Tools. So that will be removed? 

Yes, for sure.  The recipe is for a "metapackage" that contains all of the conda tools that COMPASS needs.  It couldn't live in MPAS-Model (because of the issue with how frequently it needs to be updated) but it doesn't really belong in MPAS-Tools, now that there is a separate COMPASS repo.

> Will this new compass package include MpasMeshConverter.x and other codes in MPAS-Tools, or will those split off?

The metapackage includes `mpas_tools` as one of its packages.  `MpasMeshConverter.x` is part of `mpas_tools`.

PR #13, opened 2020-11-06
Add initial condition from EN4 1900 to EC30to60E2r3 test case

xylar:
This will update the 30to60 test case to allow for use of the Met Office EN4 1900 state estimate as the ocean initial condition.

This PR replaces https://github.com/MPAS-Dev/MPAS-Model/pull/671


xylar:
@vanroekel, this needs to be merged in to document the EN4 1900 initial condition for the EC30to60E2r3 mesh.  I think that may be less pressing now than WC is looking into the EC15to60E2r4 mesh, but still...

vanroekel (approved):
Approved by visual inspection and testing on grizzly.

The new structure looks great.  I like not having to setup init and spinup a lot.  Makes it much clearer.  Thanks @xylar!

mark-petersen (approved):
Spun up for 7 days successfully, here:
```
/lustre/scratch3/turquoise/mpeterse/runs/201112_EC60to30/ocean/global_ocean/EC60to30/spin_up_EN4_1900/prep_spin_up1
```
image

xylar:
@vanroekel and @mark-petersen, thanks for the reviews!

PR #15, opened 2020-11-06
New Mesh: ECwISC30to60kmL60E3SMv2r01 and add 1900 state estimate initial condition 

xylar:
This merge updates the ECwISC30to60 test case to allow for use of the Met Office EN4 1900 state estimate as the ocean initial condition.  It attempts to create a mesh with ice-shelf cavities that is as similar as possible to https://github.com/MPAS-Dev/MPAS-Model/pull/669 (EC30to60kmL60E3SMv2r03) with the new IC from #13 

This PR replaces https://github.com/MPAS-Dev/MPAS-Model/pull/694


xylar:
To do:

- [ ] rebase
- [ ] update PR reference in metadata

xylar:
@darincomeau, @vanroekel, @jonbob and @mark-petersen, this is the PR for the ECwISC30to60E2r1 mesh that I think @darincomeau and @jonbob have already been using.  Assuming there are no changes needed, Could each of you approve this?

@vanroekel, you were involved in discussion of this mesh, but if WC isn't going to use it, maybe you don't need to approve.  Your call.

jonbob (approved):
As far as I know, the testing of this mesh has all been good

darincomeau (approved):

vanroekel (approved):
Looks great.  Approving based on visual inspection and prior testing.

mark-petersen:
These two files are added to the base_mesh subdirectory:
```
arctic_high_res_region.geojson
north_mid_res_region.geojson
```
when this case is run in compass:
```
  83: -o ocean -c global_ocean -r EC60to30 -t build_mesh
```
They are both bad links, and apparently not used as `run.py` completes successfully.

I'm guessing we just need to remove these two lines:
```
ocean/global_ocean/config_files/config_base_mesh.xml:6: 
 <add_link source_path="script_test_dir" source="arctic_high_res_region.geojson" dest="arctic_high_res_region.geojson"/>
ocean/global_ocean/config_files/config_base_mesh.xml:5: 
  <add_link source_path="script_test_dir" source="north_mid_res_region.geojson" dest="north_mid_res_region.geojson"/>
```
They probably need to be added to individual cases instead here.
```
ocean/global_ocean/EC15to60/build_mesh/build_base_mesh.py:41:    
   fc = read_feature_collection('north_mid_res_region.geojson')
ocean/global_ocean/SO60to10wISC/init/build_base_mesh.py:73:
    fc = read_feature_collection('north_mid_res_region.geojson')
ocean/global_ocean/SO60to10wISC/init/config_base_mesh.xml:6: 
   <add_link source_path="script_test_dir" source="north_mid_res_region.geojson" dest="north_mid_res_region.geojson"/>
```
so that would belong is a separate PR. Never mind - I'll do that later today.

xylar:
> These two files are added to the base_mesh subdirectory

Oh, shoot! I forgot to fix that.  It's from the EC15to60E2r4 mesh PR (https://github.com/MPAS-Dev/MPAS-Model/pull/740).

xylar:
> They probably need to be added to individual cases instead here.

Yes, this is part of the problem with trying to use symlinks for efficiency.  It's too easy to modify the contents without realizing the side effects.  The COMPASS redesign I'm working on will not have this problem because it will be much more transparent which functions are used across multiple testcases and which are specific to one testcase.

mark-petersen:
Yes, I really like to the links so we don't have many identical files to change. But the other side, unintended consequences, is a problem. Thanks for thinking about this with the rewrite.

xylar:
@mark-petersen, #25 should address the issue you pointed out.  There was already an issue about it (#19).

Are you wanting to perform additional testing before you aprove?

mark-petersen (approved):
Looks great. I just changed EN to caps to be consistent with other PR. I ran all the way through `ssh adjustment` and 7 days of spin-up. See:
```
/lustre/scratch3/turquoise/mpeterse/runs/201112_EC60to30wISC/ocean/global_ocean/EC60to30wISC/spin_up_en4_1900/prep_spin_up1
```
A quick inspection of coastlines looks good.
image
image

mark-petersen:
I wonder if we should add a forced channel in the future in the Ligurian Sea:
image
The main problem, of course, is that 60km is pretty course. It does look funny to have one long peninsula there.

vanroekel:
@mark-petersen I'd very much prefer we not open that channel in this PR.  I'd rather not remake all the E3SM mapping files for such a small change.  Especially one that will have near zero impact on global simulations.

mark-petersen:
@vanroekel sorry, didn't mean to suggest changing it here. I was just pointing it out since it is visible here, if we want to add a channel that will show up in future versions.

xylar:
@mark-petersen, thank you for the testing and the case correction.  Good to be consistent with #13.

Regarding adding another critical passage (for a future mesh), I don't think dynamics within the Mediterranean are important enough to a global simulation to bother about these things.  As you say say, 60 km is simply too coarse to expect a good representation.

More generally, if we want to represent flow more accurately at coarse resolution, I have always thought we should pursue the porous walls ([Adcroft et al. 2013](https://www.sciencedirect.com/science/article/abs/pii/S1463500313000425)) approach instead of the brute force critical passages approach.  This could be a good project for someone wanting to get started on MPAS-Ocean.

PR #3, opened 2020-11-06
Merge COMPASS contents from MPAS-Dev/MPAS-Model/landice/develop

xylar:
This merge brings in the full history of https://github.com/MPAS-Dev/MPAS-Model/tree/landice/develop/testing_and_setup/compass

PR #18, opened 2020-11-06
Fix git describe for repo without any tags

xylar:
Since this repo doesn't have any tags yet, `git describe` will fail unless it has an `--always` flag.  This fix is needed before this repo will be useful for testing.

### DIFF
--- a/setup_testcase.py
+++ b/setup_testcase.py
@@ -1716,7 +1716,7 @@ if __name__ == "__main__":
     old_dir = os.getcwd()
     os.chdir(config.get('script_paths', 'script_path'))
     git_version = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen and @matthewhoffman, this should go in right after #2 and #3 so the repo is actually usable.

matthewhoffman (approved):
I'm not familiar with the `--always` argument, but it sounds like it provides a fallback to return something useful without causing undesired behavior in situations where `git describe` works.  Unless I'm missing something, that sounds desirable, and this change makes sense to me.

mark-petersen (approved):
Looks good. I'm definitely surpassing "you learn something new every day"

xylar:
Thanks again!

PR #27, opened 2020-11-16
Add a lot more to .gitignore

xylar:
The most important ones are the python cache files, which we definitely don't want to be adding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,109 @@ MPAS-Model/*
 !MPAS-Model/ocean/develop
 !MPAS-Model/landice/develop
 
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen and @matthewhoffman, hopefully, this is one you can just review quickly by inspection.  I was working on some enhancements and it turned out that python cache files were getting added to the repo, which we definitely don't want.  Much of the rest is probably overkill but also probably does no harm.

matthewhoffman (approved):

mark-petersen (approved):
Yes, this is good. Thanks.

PR #30, opened 2020-11-18
Restructure the docs for clarity

xylar:
Make user's guide, ocean, and developer's guide sections.

Add instructions for building the documentation locally.


xylar:
@mark-petersen and @matthewhoffman, I'm not asking for a review on documentation for now, mostly because it's crappy now and only slightly less crappy after this work so it's not really worthy of a review.  But input is welcome if you happen to want to give it.

PR #31, opened 2020-11-18
Add docs from Confluence

xylar:
This merge also includes a bit of reorganization to make the table of contents more readable.
* documentation of the 4 main scripts has been consolidated into a single file

The ocean README has been updated to be consistent with the other documentation.


xylar:
@mark-petersen, this is not quite an exact copy of https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1247215971/MPAS+Testing+and+Machine-specific+Instructions#MPAS-within-E3SM but it's pretty close.  Could you follow the instructions [here](https://mpas-dev.github.io/compass/stable/building_docs.html) to build the documentation locally and tell me if you're happy?

@matthewhoffman, I'm not sure how much you and other MALI users have been making use of the Confluence instructions.  Some of it is a bit ocean-centric so feel free to recommend changes to make it more general.  Or feel free to just remove yourself as a reviewer if you'd prefer.  We can always work on making the documentation better for MALI users when you have the time and interest.

matthewhoffman:
@xylar , I removed myself as a reviewer, but I like the idea of revisiting this once landice is caught up with transition.  These docs look very thorough and useful, so I am motivated to get on board with this eventually.

docs/overview.rst
```diff
@@ -157,8 +157,8 @@ testing. For the ocean core, they are here:
 
 .. code-block:: bash
 
-    $ ls testing_and_setup/compass/ocean/regression_suites/
-    land_ice_fluxes.xml  light.xml  nightly.xml  rpe_tests.xml
+    ls testing_and_setup/compass/ocean/regression_suites/
+       land_ice_fluxes.xml  light.xml  nightly.xml  rpe_tests.xml
```
xylar: @mark-petersen, I don't agree with this change.  The `$` is a standard notation for differentiating something typed at the command line from the output it produces.

mark-petersen (approved):
Thanks @xylar. I spent an hour on it, which helped me learn the file structure. I tested some of the cut and paste instructions, and corrected some typos. I also renamed directories and files to match the table of contents exactly, so I could find which file to edit. I think this is a big improvement and looks very professional, despite the minor inconvenience of PRs rather than editing confluence.

As we add new files, let's match *.rst file name to TOC title, but all lower case.

docs/index.rst
```diff
@@ -14,6 +14,13 @@ altered in those files.
...
    scripts
 
+.. toctree::
+   :caption: Ocean
+   :maxdepth: 2
+
+   ocean
+   ocean_testcases/index
+
 .. toctree::
```
xylar: @mark-petersen, I'm not sure I agree that core info should go before machine info, but we can try this out.

docs/machines/lanl.rst
```diff
@@ -88,6 +88,7 @@ grizzly, gnu
 
 .. code-block:: bash
 
+    module use /usr/projects/climate/SHARED_CLIMATE/modulefiles/all
```
xylar: Isn't this what just stopped working on Grizzly after the latest DST?

docs/overview.rst
mark-petersen: But if we have $ everywhere, we can't copy a block of text. That is quite inconvenient. I know what you mean about differentiating between terminal input and output.

docs/scripts.rst
```diff
@@ -19,7 +19,7 @@ it will only print the flags needed to setup that specific test case.
 
 Command-line options::
 
-    $ ./list_testcases.py -h
+    ./list_testcases.py -h
```
xylar: Please change back, see my comment above

docs/scripts.rst
```diff
@@ -66,7 +66,7 @@ configuration file.
 
 Command-line options::
 
-    $ ./setup_testcase.py -h
+    ./setup_testcase.py -h
```
xylar: Please change back, see my comment above

docs/scripts.rst
```diff
@@ -117,7 +117,7 @@ setting up a test case.
 
 Command-line options::
 
-    $ ./clean_testcase.py -h
+    ./clean_testcase.py -h
```
xylar: Please change back, see my comment above

docs/scripts.rst
```diff
@@ -161,7 +161,7 @@ for each individual test case, and the run script that runs all test cases.
 
 Command-line options::
 
-    $ ./manage_regression_suite.py -h
+    ./manage_regression_suite.py -h
```
xylar: Please change back, see my comment above

docs/index.rst
mark-petersen: The core info seems more central, while the machine-specific feels more like an appendix. I was afraid people wouldn't see 'Ocean' if it is so low in the TOC.

xylar:
@mark-petersen, I like the reogranization you did to match the TOC.  A lot of the corrections are very much appreciated!  A few of the changes I disagree with and would prefer to have reverted, as I indicated above.

docs/machines/lanl.rst
mark-petersen: It doesn't work without it for gnu. It was Intel 19 that changed
```
gr-fe3:~$ module load gcc/5.3.0

gr-fe3:~$ module load openmpi/1.10.5 netcdf/4.4.1 parallel-netcdf/1.5.0 pio/1.7.2
Lmod has detected the following error:  The following module(s) are unknown: "pio/1.7.2" "parallel-netcdf/1.5.0" "netcdf/4.4.1"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore-cache load "pio/1.7.2" "parallel-netcdf/1.5.0" "netcdf/4.4.1"

Also make sure that all modulefiles written in TCL start with the string #%Module
```

docs/overview.rst
xylar: I wouldn't put it everywhere, just places where we have both a command and its output.  This is only done in places where one would be copying a single command anyway (because you don't want to copy the output).  It's important to have a clear distinction between command and output and I don't think your indentation does this.

mark-petersen:
OK. I guess there is a difference between a code block, like the `module load` blocks on the machine pages, versus pasting something with input and output, where the $ help indicate input. I'll put back the $.

docs/machines/lanl.rst
xylar: Okay, I see.  So se still get some modules that we built ourselves for gnu.  That makes sense.

xylar:
Yes, exactly.

docs/index.rst
xylar: Okay, that makes sense.  I need to remind myself what the TOC looks like.  We wouldn't want all the ocean configurations listed, I guess, because that would make the TOC too long.  As long as they're not in the top-level TOC, I agree that this organization will work.

mark-petersen:
done.

xylar:
Great, I'll merge as soon as CI has finished successfully!

PR #33, opened 2020-12-01
Add deployment script for compass envs

xylar:
To run it, you need to start with an environment with the `requests` and `progressbar2` packages (which could be a previous
`compass` environment, `e3sm-unified` or a temporary environment the user creates just for this installation.

This script is based on the one used in e3sm-unified: 
https://github.com/E3SM-Project/e3sm-unified/blob/master/e3sm_supported_machines/create_new_e3sm_unified_env.py

### DIFF
--- a/deploy/create_compass_env.py
+++ b/deploy/create_compass_env.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python
+import subprocess
+import os
+import socket
+import glob
+import stat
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I used the new script to create `compass_0.1.12` conda environments on my laptop, grizzly, compy, cori, anvil and cooley.

PR #38, opened 2021-01-07
update metapackage to 0.2.0; add git

xylar:
Git is needed because the system git on some machines (at least Anvil) does not support the required `git describe` call.


xylar:
This is not urgent and I'll keep the PR in draft mode until there is another need for a metapackage release.  I have added `git` to the install of `compass_0.1.12` on Anvil to tide us over for now.

xylar:
I don't think we will be building another compass environment for `legacy`, so I will close this.  `git` is already in the `compass` package.

PR #45, opened 2021-01-22
Add landIceMask to sea ice IC

### LINKED ISSUE #44: Add landIceMask variable to sea ice grid file
When creating a new grid, the landIceMask variable currently does not get carried into the sea ice grid file from the ocean grid file, but it's needed by MPAS-Seaice. Could this be added?

xylar:
...if there are ice-shelf cavities

Fixes #44

### DIFF
--- a/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
+++ b/ocean/global_ocean/scripts/create_E3SM_coupling_files.py
@@ -213,6 +213,7 @@ def graph_partition_ocean(config):  # {{{
 def initial_condition_seaice(config):  # {{{
 
     mesh_name = config.get('mesh', 'short_name')
+    ice_shelf_cavities = config.getboolean('main', 'ice_shelf_cavities')
 
     init_filename = config.get('main', 'initial_condition')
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

On my Linux laptop, I ran `global_ocean/QU240/init` and `global_ocean/QU240wISC/init`, including the `e3sm_coupling` steps.  The `initial_condition_seaice` function runs successfully for both configurations.  `landIceMask` is not added in the first one but is added in the second one:
```
$ ncdump -h assembled_files_for_upload/inputdata/ice/mpas-cice/QUwISC240E2r01/seaice.QUwISC240E2r01.nc | grep landIceMask
	int landIceMask(Time, nCells) ;
		landIceMask:units = "unitless" ;
		landIceMask:long_name = "Mask indicating where land-ice is present (1) or absent (0)" ;

```

xylar:
@darincomeau, I think a sufficient review would be for you to take a look at the code changes and see if they seem right to you.  I think my testing is sufficient.

darincomeau (approved):
Thanks @xylar for the quick fix!

PR #52, opened 2021-02-12
Remove config_iterative_init_variable from all tests

xylar:
This merge removes references to the `config_iterative_init_variable` namelist option, which was removed in https://github.com/MPAS-Dev/MPAS-Model/pull/812.

It also renames the field `modifySSHMask` to `modifyLandIcePressureMask`, to be consistent with changes in that PR.

### DIFF
--- a/ocean/global_ocean/EC60to30wISC/spin_up_EN4_1900/config_initial_state.xml
+++ b/ocean/global_ocean/EC60to30wISC/spin_up_EN4_1900/config_initial_state.xml
@@ -40,7 +40,6 @@
 		<template file="template_initial_state.xml" path_base="script_configuration_dir"/>
 		<template file="template_critical_passages.xml" path_base="script_core_dir" path="global_ocean"/>
 		<template file="template_init_with_land_ice.xml" path_base="script_configuration_dir"/>
-		<option name="config_iterative_init_variable">'landIcePressure_from_top_density'</option>
 		<option name="config_global_ocean_depth_file">'layer_depth.nc'</option>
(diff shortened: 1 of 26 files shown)


xylar:
## Testing

I tested this with the `z_level_dev` test suite (https://github.com/xylar/compass/blob/z_level_dev_baseline/ocean/regression_suites/z_level_dev.xml) and everything ran as expected.

I did regression testing as part of https://github.com/MPAS-Dev/MPAS-Model/pull/812, and the only changes were expected ones related to `landIcePressure` --> `landIcePressure_from_top_density` as the initialization strategy in two `sub_ice_shelf_2D` tests.

xylar:
@cbegeman, could you run a quick test (either the `nightly` or `z_level_dev` test suite) on this branch using the model code in the submodule `MPAS-Model/ocean/develop`?  You don't have to give it a baseline, we just want to make sure it runs okay.

I have removed the `z_level_dev` test suite from this branch because we want to merge this branch into `legacy` but don't want to keep that test suite in the long term.  So I created another branch just for adding the test suite: https://github.com/xylar/compass/tree/z_level_dev_baseline

If we can get this reviewed and merged, it will make the rest of our testing of `minLevel*` cleaner because the baseline will correspond to the current `MPAS-Model/ocean/develop` and `compass/legacy`.

cbegeman (approved):
@xylar I ran the nightly regression suite and it passes. Thanks for getting this in!

xylar:
Thanks for testing @cbegeman!

PR #28, opened 2020-11-16
compass python package (compass v1.0)

xylar:
This merge creates a `compass` python package that can be used to list, set up, run, validate, and clean up test cases and set up, run, and clean up test suites.

This prototype contains  `landice` and `ocean` MPAS cores.  The `landice` core has all test cases in the `sia_integration` test suite and a few other related test cases in the test groups that are in that test suite.  The `ocean` core has all of the test cases for the `baroclinic_channel` test group, all test cases from the `nightly` regression suite, and the `QU240`, `QUwISC240`, `EC30to60`, `ECwIsC30to60` and `SOwISC12to60` meshes from "legacy" COMPASS (COMPASS as it currently exists).

A significant number of python modules (python files) and packages (subdirectories) have been added within `compass` that break up the tasks related to listing, setting up, and cleaning up test cases that were previously in `list_testcases.py`, `setup_testcase.py` and `clean_testcase.py` as well as functionality related to test suites that was in `manage_regression_suite.py` before.  New functionality has also been added to make setting up and running test cases easier, and to promote more code reuse than the current COMPASS framework permits.  See the design document in #29 for more details.

Each test case has an associated configuration file (`.cfg`) that is constructed by combining config options from various sources: a core-specific config file, one associated with the machine (if running on a "supported" machine), one for the configuration, one for the test case, and a user-defined config file.  Typically, these contain different types of options with different purposes.  For example, the core's config file points to namelist and streams templates that are core specific.  The machine config file contains information about the batch queuing system, the number of cores per node, etc.  The test group and test case config files contain options that can be treated as the python equivalents of the namelist options used in the Fortran MPAS-Model code.  The user-defined config file can override any config options from these other sources, and can provide paths to initial conditions, meshes and other data that the core or test case may require.  An effort has been made to automate most of this so a user config file should only be needed on a non-supported machine.  Users can also edit the local config file within a test case (and symlinked within each step) that collects all of the config options for that test case before running.

closes #11 
closes #9 
closes #20

PR #62, opened 2021-03-27
Switch global ocean mesh culling and files for E3SM to use python masks

xylar:
This merge switches the `global_ocean` test group to using a set of threaded python tools for creating masks.  These same tools are being used in MPAS-Analyisis and have been shown to perform better on larger meshes than the C++ equivalents.

The python version of the transect mask creator also has the advantage that it follows the transect through polygons on the sphere rather than zig-zagging (which has been a problem with `MpasMaskCreator.x`). 

The results with the new tools are not expected to be bit-for-bit as compared with the previous mask creator.  So the test suites should pass but not bit-for-bit for global ocean test cases.

See https://github.com/MPAS-Dev/MPAS-Tools/pull/399

### DIFF
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/diagnostics_files.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/diagnostics_files.py
@@ -6,9 +6,9 @@ from pyremap import get_lat_lon_descriptor, get_polar_descriptor, \
     MpasMeshDescriptor, Remapper
 from geometric_features import GeometricFeatures
 from geometric_features.aggregation import get_aggregator_by_name
-from mpas_tools.mesh.conversion import mask
+from mpas_tools.logging import check_call
+from mpas_tools.ocean.moc import add_moc_southern_boundary_transects
 from mpas_tools.io import write_netcdf
-from mpas_tools.ocean.moc import make_moc_basins_and_transects
 
 from compass.io import symlink
 from compass.step import Step
@@ -63,7 +63,6 @@ class DiagnosticsFiles(Step):
         logger = self.logger
 
         restart_filename = 'restart.nc'
-
         with xarray.open_dataset(restart_filename) as ds:
             mesh_short_name = ds.attrs['MPAS_Mesh_Short_Name']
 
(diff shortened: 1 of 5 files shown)


xylar:
@mark-petersen and @vanroekel, I know I've been bombarding you with `comapss` pull requests.  There has been a backlog as #28 was being reviewed.  This one is perhaps more important than #76 but also perhaps more involved to review.  It would be important to take a look at the meshes that result from the new masks and make sure there are no unexpected surprises, particularly with the critical passages and land blockages.

I will post results from my testing, too, as soon as I have finished.

xylar:
Womp-womp-womp... It turns out this branch won't work on Anvil until I get #64 taken care of.  That's a big project, too, so I'm closing this for now.

xylar:
## Testing

I will test `QU240` through the `file_for_e3sm` test case on Grizzly, Badger, Anvil and Chrysalis with intel.  I will test `EC30to60`, `QUwISC240`, `ECwISC30to60` and `SOwISC12to60` on Chrysalis with intel only.

Tests (checkmark means completed successfully):
- [x] Grizzly intel `QU240`
- [x] Badger intel `QU240`
- [ ] Anvil intel `QU240`
- [ ] Chrysalis intel `QU240`
- [ ] Chrysalis intel `EC30to60`
- [ ] Chrysalis intel `QUwISC240`
- [ ] Chrysalis intel `ECwISC30to60`
- [ ] Chrysalis intel `SOwISC12to60`

## Timing

Here is a note of the times for different test cases on Chrysalis with intel and impi, for comparison with https://github.com/MPAS-Dev/compass/pull/85#issuecomment-829135859:

`QU240`:
```
```

`QUwISC240`:
```
```

`EC30to60`:
```
```

`ECwISC30to60`:
```
```

`SOwISC12to60` (timed out):
```
```

xylar:
@mark-petersen and @vanroekel, I was going to have you both review this PR but you are both busy with other stuff this week.

I would invite you to browse the previous results here (from #85):
/home/ac.xylar/e3sm/ac.xylar/compass_1.0/chrysalis/test_20210429/sys_e3sm_intel
and the new results here:
/home/ac.xylar/e3sm/ac.xylar/compass_1.0/chrysalis/test_20210429/python_masks
if you have time and let me know if you notice any concerning changes in any of the 5 meshes.

If you don't have time or interest in looking at these, I think it is reasonable to revisit this as part of creating the next generation of meshes from `compass` whenever that may happen.

vanroekel:
@xylar not sure if it is most appropriate to comment here, but just wanted to say that I'm happy to take a look at the meshes (may not happen until later next week though), but think it will also be good to also revisit this when new meshes come around.  It would also be great to get Andrew R's scripts available as a part of the standard workflow analyzing these meshes.

xylar:
@vanroekel, fine to comment here.

Andrew's scripts are in Matlab as I understand it. The port would be a nontrivial amount of work but I agree with you.

PR #63, opened 2021-03-27
Change how variable comparison is handled

### LINKED ISSUE #59: request to add separate compare pass/fail status
It would be really helpful for suites with `-b` to report the test result and the comparison result separately:
```
  PASS  compare: PASS
  PASS  compare: FAIL
ocean/baroclinic_channel/10km/decomp_test
  FAIL  compare: FAIL
````
Right now, FAIL could indicate the comparison or the run itself. The two are obviously both checked lower down, but I think only one is passed up to the suite function for reporting.

### PR DESCRIPTION
**Pull request by xylar**
Perform all variable comparisons before either raising an exception (in an individual test case) or printing a failure message that separates the "internal" test status from the comparison with the baseline.

closes #59

### DIFF
--- a/compass/landice/tests/dome/decomposition_test/__init__.py
+++ b/compass/landice/tests/dome/decomposition_test/__init__.py
@@ -62,6 +62,6 @@ class DecompositionTest(TestCase):
         variables = ['thickness', 'normalVelocity']
         steps = self.steps_to_run
         if '1proc_run' in steps and '4proc_run' in steps:
-            compare_variables(variables, self.config, work_dir=self.work_dir,
+            compare_variables(test_case=self, variables=variables,
                               filename1='1proc_run/output.nc',
                               filename2='4proc_run/output.nc')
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

**Comment by xylar** [comment]
This will be next after #76

**Comment by xylar** [comment]
This was a request form @mark-petersen in #59.  I agree that it's a welcome change.  As I mentioned there, typical output with this branch will look like:
```
ocean/global_ocean/QU240/PHC/RK4/performance_test
  PASS  compare: PASS
ocean/global_ocean/QU240/PHC/RK4/restart_test
  FAIL  compare: PASS
  see: case_outputs/ocean_global_ocean_QU240_PHC_RK4_restart_test.log
```
So it's hopefully clear that the RK4 restart test failed "internally" (the restart run didn't match the fully run) but was bit-for-bit when compared to the baseline.

@mark-petersen, please let me know when you get a chance if this does what you wanted.

@matthewhoffman, let us know if this seems okay to you or what changes you would suggest.

**Review by matthewhoffman — APPROVED** [review]
This looks quite useful to me.

A minor suggestion is to adjust the output formatting.  The string "PASS  compare: PASS" is kind of confusing.  Perhaps something like:

```
ocean/global_ocean/QU240/PHC/RK4/performance_test
  test: PASS  baseline comparison: PASS
```

xylar:
I'm fine with @matthewhoffman suggestion,  @mark-petersen, are you okay with it, too?

xylar:
I went head and made the change, but we can further modify if you like, @mark-petersen.

xylar:
@mark-petersen, I'm going to go ahead with this one.

PR #77, opened 2021-04-22
Mixed layer eddy test cases

qingli411:
This PR adds the configuration files for two mixed layer eddy test cases as described in https://github.com/MPAS-Dev/MPAS-Model/pull/855.

### DIFF
--- a/ocean/mixed_layer_eddy/0.6km/double_front/config_driver.xml
+++ b/ocean/mixed_layer_eddy/0.6km/double_front/config_driver.xml
@@ -1,0 +1,8 @@
+<driver_script name="run_test.py">
+	<case name="init">
+		<step executable="./run.py" quiet="true" pre_message=" * Running init" post_message="     Complete"/>
+	</case>
+	<case name="forward">
+		<step executable="./run.py" quiet="true" pre_message=" * Running forward" post_message="     Complete"/>
+	</case>
+</driver_script>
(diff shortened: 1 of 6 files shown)


vanroekel (commented):
approving by visual inspection and a test setting up and starting a run.  Worked great and the code looks good to me.  Thanks @qingli411

vanroekel:
@anirban89 -- this is the first MLE test case PR, this is to set them up.  There is a corresponding PR for the MPAS code itself.

mark-petersen:
@qingli411, please post some results of your test case. Whatever the standard output is - either it runs to completion, or it makes a plot, etc. How do you know it is successful?

Please state the compiler(s), machine, and exact test cases that you ran. I'll give that a look, and then approve. Is this contingent on any further work in MPAS-Ocean code now?

qingli411:
@mark-petersen, this test case does not include an analysis step because the visualizations of this test case as described in https://github.com/MPAS-Dev/MPAS-Model/pull/855 depend on a set of other tools I have. I don't have a better way to determine if it is successful other than to see if it runs to completion. All the modifications of the MPAS-Ocean code it requires are in https://github.com/MPAS-Dev/MPAS-Model/pull/855. The recent force-push in this PR was just to rebase it to the latest compass legacy branch.

I tested this case before the rebase on a Linux workstation (which I no longer have access to) with GNU compiler. The rebased version also appears to work on my MacBook with GNU compiler, though I didn't finish the run on my MacBook since it will take a while.

qingli411:
@mark-petersen, it just occurred to me that I wasn't testing this case on the Linux workstation (I was thinking about another test case). It was on Badger with GNU compiler.

mark-petersen (approved):
@qingli411 Based on your comment here: https://github.com/E3SM-Project/E3SM/pull/4357#issuecomment-869803860, this is working, other than a PIO file error that is probably unrelated to this particular PR.

xylar:
@mark-petersen, it's not a big deal but this should really only have been merged after https://github.com/E3SM-Project/E3SM/pull/4357 and after updating the submodule to point to the new E3SM master.

mark-petersen:
Good point, I forgot about that step with the new E3SM-Project submodule. I'll remember that in the future.

PR #79, opened 2021-04-24
Fix metadata for `global_ocean` test group

xylar:
Put `"not found"` in metadata for missing packages, rather than raising an exception.  We should not require that the `compass` package be present.  I took it out of the test environment and discovered these errors.


xylar:
# Testing

The `nightly` test suite completes with this change, whereas it does not with `master` using a new test environment I created on Anvil.

xylar:
@mark-petersen and @vanroekel, I'm just making you aware of this PR.  I am not going to request a review because I think it's a straightforward bug fix.

PR #81, opened 2021-04-26
Added the new test case: Neverworld2 circumpolar acc

alicebarthel:
Created a "neverworld2" test case (with a circumpolar ACC): single 2-hemispheric basin with circumpolar ACC, adiabatic with analytical topography and wind stress.


test_windprofile.ipynb
```diff
@@ -0,0 +1,270 @@
+{
+ "cells": [
+  {
```
xylar, line 3: @alicebarthel, I don't think this file should have been committed at the top level.

xylar:
@alicebarthel, I can't remember exactly what we discussed.  But this is a work in progress, right, not ready for review yet?

xylar:
@alicebarthel and @mark-petersen, I think it would make sense to keep this branch around for reference but to close this PR. 
 I think at this point is doesn't make sense to add a major test case to `legacy`, and a new branch would be needed to port this work to the new approach in `master`.
 
 If you agree, would you close this PR, with the idea that you could open a new one once you're ready?

PR #78, opened 2021-04-22
Update intel 19 compiler instructions

mark-petersen:
This compiler set for intel 19 + scorpio works.

PR #82, opened 2021-04-27
Update legacy docs

xylar:
This merge updates Intel 19 compiler instructions on grizzly.

It also switches from `load_latest_compass.sh` to `load_legacy_compass.sh`.  A copy of the activation script with the new name has been added on Anvil/Chrysalis, Grizzly/Badger and Cori.

### DIFF
--- a/docs/machine_specific_instructions/anvil.rst
+++ b/docs/machine_specific_instructions/anvil.rst
@@ -9,7 +9,7 @@ First, you might want to build SCORPIO (see below), or use the one from
 
 .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


xylar:
@mark-petersen, I'm taking your commit from #78 here and adding another change.  Please let me know if you're okay with the name change of the activation script.

xylar:
@mark-petersen, I'm going to go ahead with this since you're out.  We can revisit this later if need be.

mark-petersen:
Thanks for keeping up the documentation.

PR #83, opened 2021-04-27
Rename docs from master from "stable" to "latest"

xylar:
The `master` branch here is not necessarily a stable branch, since we don't have a development branch, so the docs should be called `latest` like we do for development branches in most other repos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ to setup and configure a test case.
 
 The latest compass documentation can be found here:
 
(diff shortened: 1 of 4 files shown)


xylar:
@mark-petersen and @matthewhoffman, I'm not going to ask for your review of this change.  Just making you aware that I'm moving the docs from `stable` to `latest` because I think `master` isn't going to be stable in general.  It's our development branch.

PR #87, opened 2021-04-28
Added a case to validate the implementation of suspended sediment as a tracer

caozd999:
This PR added a `sed_tracer` case in the compass to validate the implementation of suspended sediment as a new tracer  (currently without source/sink terms) in MPAS-O. The corresponding code was submitted separately (https://github.com/MPAS-Dev/MPAS-Model/pull/858). 

The test case has the same configurations of the implemented sediment tracer (**SSC**) and an existing debug tracer (**tracer**) in a closed domain. The initial condition is that horizontally half of the domain has tracers and high water temperature, whereas the other half has no tracers and low water temperature. 

There are two scenarios in the test case:
Scenario 1) **tracer_diffusion** without advection by setting `config_disable_tr_adv = .true. `
Scenario 2) **tracer_advection_diffusion** with advection by setting `config_disable_tr_adv = .false. `

A _default_ case was also added as **tracer_regression**.

The validation idea is to check whether the results of `SSC` and `tracer` are identical in each scenario.

Here are the results:
**Scenario 1**
SSC_1
tracer_1

**Scenario 2**
SSC_2
tracer_2

Note: visualization scripts were adopted from https://github.com/qingli411/mpasview. Thank you @qingli411 for sharing.

### DIFF
--- a/ocean/sediment/100m/tracer_regression/config_driver.xml
+++ b/ocean/sediment/100m/tracer_regression/config_driver.xml
@@ -1,0 +1,17 @@
+<driver_script name="run_test.py">
+	<case name="init">
+		<step executable="./run.py" quiet="true" pre_message=" * Running init" post_message="     Complete"/>
+	</case>
+	<case name="forward">
+		<step executable="./run.py" quiet="true" pre_message=" * Running forward" post_message="     Complete"/>
+	</case>
+	<validation>
+		<compare_fields file1="forward/output.nc">
+			<field name="sedConcen" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="temperature" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="salinity" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="layerThickness" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+			<field name="normalVelocity" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
+		</compare_fields>
+	</validation>
+</driver_script>
--- a/ocean/regression_suites/nightly.xml
+++ b/ocean/regression_suites/nightly.xml
@@ -53,4 +53,7 @@
 	<test name="Periodic Planar 20km - LIGHT particle time reset test" core="ocean" configuration="periodic_planar" resolution="20km" test="time_reset_light_test">
 		<script name="run_test.py"/>
 	</test>
+	<test name="Sediment 100m - Tracer regression test" core="ocean" configuration="sediment" resolution="100m" test="tracer_regression">
+		<script name="run_test.py"/>
+	</test>
 </regression_suite>
--- a/ocean/sediment/100m/tracer_advection_diffusion/config_driver.xml
+++ b/ocean/sediment/100m/tracer_advection_diffusion/config_driver.xml
@@ -1,0 +1,8 @@
+<driver_script name="run_test.py">
+	<case name="init">
+		<step executable="./run.py" quiet="true" pre_message=" * Running init" post_message="     Complete"/>
+	</case>
+	<case name="forward">
+		<step executable="./run.py" quiet="true" pre_message=" * Running forward" post_message="     Complete"/>
+	</case>
+</driver_script>
--- a/ocean/sediment/100m/tracer_advection_diffusion/config_forward.xml
+++ b/ocean/sediment/100m/tracer_advection_diffusion/config_forward.xml
@@ -1,0 +1,61 @@
+<?xml version="1.0"?>
+<config case="forward">
+	<add_link source="../init/ocean.nc" dest="init.nc"/>
+	<add_link source="../init/mesh.nc" dest="mesh.nc"/>
+	<add_link source="../init/graph.info" dest="graph.info"/>
+	<copy_file source_path="script_test_dir" source="gotmturb.nml" dest="gotmturb.nml"/>
+
+	<add_executable source="model" dest="ocean_model"/>
+
+	<namelist name="namelist.ocean" mode="forward">
+		<option name="config_ocean_run_mode">'forward'</option>
+		<option name="config_dt">'00:00:10'</option>
+		<option name="config_time_integrator">'RK4'</option>
+		<option name="config_run_duration">'0006_00:00:00'</option>
+		<option name="config_disable_tr_adv">.false.</option>
+		<option name="config_use_debugTracers">.true.</option>
+		<option name="config_use_sedimentTracers">.true.</option>
+		<option name="config_use_mom_del2">.true.</option>
+		<option name="config_mom_del2">1.0e-2</option>
+		<option name="config_use_tracer_del2">.true.</option>
+		<option name="config_tracer_del2">10.0</option>
+		<option name="config_use_cvmix">.false.</option>
+		<option name="config_use_gotm">.true.</option>
+		<option name="config_gotm_namelist_file">'gotmturb.nml'</option>
+		<option name="config_gotm_constant_bottom_roughness_length">5.0e-3</option>
+		<option name="config_gotm_constant_bottom_drag_coeff">1.73e-2</option>
+		<option name="config_use_implicit_bottom_drag">.true.</option>
+		<option name="config_implicit_bottom_drag_coeff">1.73e-2</option>
+	</namelist>
+
+	<streams name="streams.ocean" keep="immutable" mode="forward">
+		<stream name="mesh">
+			<attribute name="filename_template">mesh.nc</attribute>
+		</stream>
+		<stream name="input">
+			<attribute name="filename_template">init.nc</attribute>
+		</stream>
+		<stream name="output">
+			<attribute name="type">output</attribute>
+			<attribute name="filename_template">output.nc</attribute>
+			<attribute name="output_interval">0000-00-00_01:00:00</attribute>
+			<attribute name="clobber_mode">truncate</attribute>
+			<add_contents>
+				<member name="tracers" type="var_struct"/>
+				<member name="velocityZonal" type="var"/>
+				<member name="velocityMeridional" type="var"/>
+				<member name="normalVelocity" type="var"/>
+				<member name="mesh" type="stream"/>
+				<member name="xtime" type="var"/>
+				<member name="layerThickness" type="var"/>
+			</add_contents>
+		</stream>
+	</streams>
+
+	<run_script name="run.py">
+		<step executable="gpmetis">
+			<argument flag="graph.info">2</argument>
+		</step>
+		<model_run procs="2" threads="1" namelist="namelist.ocean" streams="streams.ocean"/>
+	</run_script>
+</config>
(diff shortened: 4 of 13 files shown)


xylar:
@caozd999, the movies linked above aren't working for me.  Do they play for you?

xylar:
@caozd999, I don't think you need everyone to review the COMPASS side so I'm taking @maltrud and @lconlon off.  (Anyone's feedback if appreciated, I just don't think it's required.)

caozd999:
> @caozd999, the movies linked above aren't working for me. Do they play for you?

No, the movies don't play on my side either. I am trying to fix them.

maltrud:
the animations played fine for me.

caozd999:
> @caozd999, the movies linked above aren't working for me. Do they play for you?

The movies can play in Safari.

cbegeman:
@caozd999 The videos look good! I'm wondering if we want this test case in the new `sediment` category or the existing `lock_exchange` category. My inclination would be the latter, as a `sediment_test`. 

Apologies if you know this already: You can test different terms with some existing config options, e.g. `config_disable_tr_adv`. Hopefully the tracer options extend to your sediment tracer. You might find this helpful in making sure you're getting the expected behavior.

caozd999:
> @caozd999 The videos look good! I'm wondering if we want this test case in the new `sediment` category or the existing `lock_exchange` category. My inclination would be the latter, as a `sediment_test`.
> 
> Apologies if you know this already: You can test different terms with some existing config options, e.g. `config_disable_tr_adv`. Hopefully the tracer options extend to your sediment tracer. You might find this helpful in making sure you're getting the expected behavior.

Thanks @cbegeman. I created the new category `sediment` because I would want to put all the sediment related test cases under this category, currently there is only one tho.

And  thanks for  pointing out the `config_disable_tr_adv` option, I have added it to new test case files and it worked perfectly.

caozd999:
> @caozd999, the movies linked above aren't working for me. Do they play for you?

I had to change the movie format from mp4 to gif.

ocean/sediment/sed_tracer/100m/config_forward1.xml
```diff
@@ -0,0 +1,65 @@
...
+                <option name="config_use_tracer_del2">.true.</option>
+                <option name="config_tracer_del2">10.0</option>
+		<option name="config_use_cvmix">.false.</option>
+		<option name="config_use_gotm">.true.</option>
+		<option name="config_gotm_namelist_file">'gotmturb.nml'</option>
+		<option name="config_gotm_constant_bottom_roughness_length">5.0e-3</option>
+		<option name="config_gotm_constant_bottom_drag_coeff">1.73e-2</option>
+		<option name="config_use_implicit_bottom_drag">.true.</option>
+		<option name="config_implicit_bottom_drag_coeff">1.73e-2</option>
+	</namelist>
```
xylar, line 29: You are indenting with a mix of tabs and spaces.  These should all be tabs.

ocean/sediment/sed_tracer/100m/config_forward2.xml
```diff
@@ -0,0 +1,65 @@
...
+                <option name="config_use_tracer_del2">.true.</option>
+                <option name="config_tracer_del2">10.0</option>
+		<option name="config_use_cvmix">.false.</option>
+		<option name="config_use_gotm">.true.</option>
+		<option name="config_gotm_namelist_file">'gotmturb.nml'</option>
+		<option name="config_gotm_constant_bottom_roughness_length">5.0e-3</option>
+		<option name="config_gotm_constant_bottom_drag_coeff">1.73e-2</option>
+		<option name="config_use_implicit_bottom_drag">.true.</option>
+		<option name="config_implicit_bottom_drag_coeff">1.73e-2</option>
+	</namelist>
```
xylar, line 29: Same spaces-instead-of-tabs issue.

ocean/sediment/sed_tracer/100m/config_forward2.xml
```diff
@@ -0,0 +1,65 @@
...
+		</stream>
+		<stream name="input">
+			<attribute name="filename_template">init.nc</attribute>
+		</stream>
+		<stream name="output">
+                        <attribute name="type">output</attribute>
+<!---      
+      <attribute name="filename_interval">0001-00-00_00:00:00</attribute>
+      <attribute name="filename_template">output/output.$Y-$M-$D.nc</attribute>
+-->
```
xylar, line 43: Please remove this or add a comment about under which circumstances it should be commented in.

ocean/sediment/sed_tracer/100m/config_init.xml
```diff
@@ -0,0 +1,70 @@
...
+		<option name="config_ocean_run_mode">'init'</option>
+		<option name="config_init_configuration">'lock_exchange'</option>
+		<option name="config_vert_levels">-1</option>
+		<option name="config_write_cull_cell_mask">.true.</option>
+		<option name="config_use_debugTracers">.true.</option>
+                <option name="config_use_sedimentTracers">.true.</option>
+                <option name="config_lock_exchange_direction">'y'</option>
+                <option name="config_lock_exchange_vert_levels">50</option>
+		<option name="config_lock_exchange_bottom_depth">10.0</option>
+		<option name="config_lock_exchange_warm_temperature">30.0</option>
```
xylar, line 16: Again, these should all be tabs.

caozd999: @xylar I fixed all the spaces-instead-of-tabs issues here manually.

xylar:
@caozd999, could you rebase this onto `MPAS-Deve/compass/legacy` when you get a chance to get rid of commits 828e1c2 to 
d7e4037?  It seems like you maybe cherry-picked those or something like that.  The review would be easier without those commits and we want to make sure they don't get merged, since they have different hashes than the same commits from the `legacy` branch.

ocean/sediment/sed_tracer/100m/config_forward2.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0"?>
+<config case="forward2">
+	<add_link source="../init/ocean.nc" dest="init.nc"/>
+	<add_link source="../init/mesh.nc" dest="mesh.nc"/>
+	<add_link source="../init/graph.info" dest="graph.info"/>
+	<add_link source="../gotmturb.nml" dest="gotmturb.nml"/>
```
xylar, line 6:
```suggestion
	<copy_file source_path="script_test_dir" source="gotmturb.nml" dest="gotmturb.nml"/>
```
This link doesn't point to anything currently, and the test seems to fail as a result.  My suggestion is to make a local copy of the namelist.  This is what was done in `gotm` tests.  That way, if a user modifies the file, it doesn't change the version in the repo and accidentally get committed.

ocean/sediment/sed_tracer/100m/config_forward1.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0"?>
+<config case="forward1">
+	<add_link source="../init/ocean.nc" dest="init.nc"/>
+	<add_link source="../init/mesh.nc" dest="mesh.nc"/>
+	<add_link source="../init/graph.info" dest="graph.info"/>
+	<add_link source="../gotmturb.nml" dest="gotmturb.nml"/>
```
xylar, line 6:
```suggestion
	<add_link source_path="script_test_dir" source="gotmturb.nml" dest="gotmturb.nml"/>
```
Same as for my comment on `forward2`.

xylar:
@caozd999, do you envision always running both `forward1` and `forward2`?  To me, they seem like separate test cases.  If that's the case, I would change that accordingly.

Currenly, your directory structure is:
```
sediment/sed_tracer/100m
```
In `legacy`, we would always organize this as:
```
sediment/100m/sed_tracer
```
My suggestion is that you break `sed_tracer` into:
```
sediment/100m/sed_tracer_scenario1
sediment/100m/sed_tracer_scenario2
```
The `config_init.xml` and `config_driver.xml` would be identical for both but `config_forward.xml` would be different.  Does that make sense for your problem?  This way, a user would more clearly know that they could run either scenario 1 or scenario 2 as their own test case.

xylar:
One more thing, both of these test cases are quite long.  I would suggest making a `default` test case with a very short (say a few time steps) forward run.  This should include some relevant validation of the output so that the test could be included in our nightly regression suite to make sure that no harm has been done to sediment transport by other changes in the code.  Here's an example: 

https://github.com/MPAS-Dev/compass/blob/690b3dcb14dc1a3b0daa72e8e449e98229aa7f3b/ocean/sub_ice_shelf_2D/5km/default/config_driver.xml

xylar:
@caozd999, it could also be that it is too early to have validation and that you prefer to keep the 2 forward runs in the same test case.  If that is true, just let me know.

xylar:
Great, I was able to run both test cases (after rebasing and fixing the GOTM namelist issue mentioned above).   I viewed the horizontal slices in paraview and they seem consistent with the plots above.  This seems great!  Once I've heard back about the questions/suggestions above, I'll be happy to approve.

ocean/sediment/sed_tracer/100m/config_forward2.xml
caozd999: I used `add_link` as you suggested in the following comment.

caozd999:
> @caozd999, it could also be that it is too early to have validation and that you prefer to keep the 2 forward runs in the same test case. If that is true, just let me know.

I modified the test case folder structure as we discussed on slack. Thanks for your help!

ocean/sediment/sed_tracer/100m/config_forward1.xml
caozd999: resolved.

ocean/sediment/100m/tracer_regression/config_driver.xml
```diff
@@ -0,0 +1,8 @@
+<driver_script name="run_test.py">
+	<case name="init">
+		<step executable="./run.py" quiet="true" pre_message=" * Running init" post_message="     Complete"/>
+	</case>
+	<case name="forward">
+		<step executable="./run.py" quiet="true" pre_message=" * Running forward" post_message="     Complete"/>
+	</case>
+</driver_script>
```
xylar, line 8:
```suggestion
	<validation>
		<compare_fields file1="forward/output.nc">
			<field name="sedConcen" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
			<field name="temperature" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
			<field name="salinity" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
			<field name="layerThickness" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
			<field name="normalVelocity" l1_norm="0.0" l2_norm="0.0" linf_norm="0.0"/>
		</compare_fields>
	</validation>
</driver_script>
```
This is needed in order for this test to function as a regression test.  With these lines added, it will compare with the same file in a baseline run (if one is provided to `./manage_regression_suite.py`) and make sure all these fields are exactly (bit-for-bit) identical between the two runs.

xylar:
@caozd999, could you add the `tracer_regression` test to the nightly regression suite at the end?
https://github.com/MPAS-Dev/compass/blob/4cee6c61ab7cb33eddab419ffbcdc99f593d0650/ocean/regression_suites/nightly.xml#L56-L58

xylar:
Closing this for now, as Zhendong isn't likely to have time or interest to finish this work.

PR #84, opened 2021-04-27
Make activation scripts (test envs and releases) on supported machines

xylar:
This merge adds many features related 
* making activation scripts that 
  * activate the ``compass`` conda environment
  * loading modules
  * setting environment variables (`NETCDF`, `PIO`, etc.)
* Use E3SM compilers and MPI libraries on each machine to build:
  * ESMF (needed to make mapping files efficiently)
  * SCORPIO
* Documenting these chnages
* Updating the user's guide to make clearer that the instructions won't work yet (not until we have a 1.0.0 release).

As I added to the documentation, with these changes, a developer should now be able to run:
```
source ./load/load_compass_env.sh -m <machine> -c <compiler>
```
in the local compass branch to active the appropriate compass test environment as well as load the system modules and set the environment variables needed to build MPAS in standalone.  This always includes setting the environment variable to point to a SCORPIO build.

I have build SCORPIO on all machines with all compiler/MPI combinations I intend to support.  I have also build ESMF on all machines and compilers except Cori (no luck there so Cori will use the serial conda-forge version).

You no longer need the `-m <machine>` flag in `python -m compass setup` or `python -m compass suite`.  It knows what machine you are on because of the source command above.

When you set up a test case or suite, there will be a local link to the the environment activation script.  All you have to do there is:
```
source load_compass_env.sh
```
It already knows what machine and what compilers you asked for.

PR #90, opened 2021-04-29
Symlink load_compass_env.sh in test case work dirs

xylar:
This was supposed to happen already but is linking in the base work directory, not the test case's work dir.


xylar:
## Testing

I verified that, when I set up a test case on Chrysalis using the environment and variables produced with:
```
source load/load_compass_env.sh
```
I get the expected links to the same environment in `load_compass_env.sh` in each test case of the ocean nightly regression suite.

xylar:
This was a bug in #84.

PR #92, opened 2021-04-29
Don't try to symlink compass to itself

### LINKED ISSUE #67: Error when using the repo base as a work dir
If you use the compass repo as a work directory (the default), you get an error because `compass` tries to make a symlink to the existing `compass` directory.

xylar:
If the repo base is also the work directory (the default), we can't symlink compass to itself.

This merge also adds directories and files related to testing to `.gitignore` in case anyone does use the repo as the work directory.

closes #67

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,3 +114,14 @@ ENV/
 
 # PyCharm
 .idea/
+
+# local config files
+/*.cfg
+
+# work directories in repo
+/ocean/
+/landice/
+/case_output/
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I confirmed that I can successfully set up the `ocean/baroclinic_channel/10km/default` test case and the `nighly` test suite in the repo, and that the resulting files are ignored when I run `git status`.

PR #93, opened 2021-04-29
Fix mixedLayerDepth (non)output for ssh_adjustment

### LINKED ISSUE #68: SSH adjustment produces error logs
These are errors about a nonexistent stream for mixed-layer depth.  The desired behavior was to have this stream disabled, but `type = 'none'` doesn't seem to have done the trick.  `output_interval = 'none'` does seem to have the desired behavior.  I have a fix in: https://github.com/xylar/compass/tree/fix_ssh_adjustment_errors

xylar:
Also disable SSH consistency check.

These errors aren't fatal but they mean that the `ssh_adjustment` directory has a number of MPAS-Ocean error files, which looks troubling.  Better to prevent them.

closes #68

### DIFF
--- a/compass/ocean/namelists/namelist.ssh_adjust
+++ b/compass/ocean/namelists/namelist.ssh_adjust
@@ -1,2 +1,3 @@
 config_run_duration = '0000_01:00:00'
+config_check_ssh_consistency = .false.
 config_land_ice_flux_mode = 'pressure_only'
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I ran the `quwisc240` test suite on Ubuntu with gnu.  

With these fixes, `ocean/global_ocean/QUwISC240/PHC/init/ssh_adjustment` no longer has any `log*.err` files.

PR #94, opened 2021-04-29
Don't use "git describe" if not in a git repo

### LINKED ISSUE #69: compasss package in conda env producing git warnings
When setting up a test suite in from the conda package, you see errors like:
```
fatal: not a git repository (or any parent up to mount point /turquoise)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```
We shouldn't call `git describe` at all if not in a git repo.  Fix is in: https://github.com/xylar/compass/commits/no_git_commands_from_package

xylar:
This is causing distracting warnings when working with the compass package in the conda environment.

closes #69

### DIFF
--- a/compass/provenance.py
+++ b/compass/provenance.py
@@ -16,12 +16,14 @@ def write(work_dir, test_cases):
     test_cases : dict
         A dictionary describing all of the test cases and their steps
     """
-    try:
-        args = ['git', 'describe', '--tags', '--dirty', '--always']
-        git_version = subprocess.check_output(args).decode('utf-8')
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I built a test package and ran the following in a directory that is not a local git repo:
```
compass suite -c ocean -t nightly -s -f ~/code/mpas-work/compass/compass/ocean.cfg -p ~/code/mpas-work/compass/compass/MPAS-Model/ocean/develop/
```
The test suite was set up correctly from the conda package without any errors or warnings relate to git version.

PR #95, opened 2021-04-29
Do not modify the radius of the Earth

### LINKED ISSUE #61: Make sure we set `config_expand_sphere = .false.` in compass 1.0
We don't want to be renormalizing the Earth sphere, we want to be leaving it at the CIME value.

xylar:
We want to stay with the Earth radius from CIME used throughout compass, not the one in MPAS-Ocean.

closes #61


xylar:
This is a change that @proteanplanet requested quite a long time ago.  We want to make sure that the CIME Earth radius is used consistently throughout the E3SM mesh generation process.  This has already happened for steps that are in MPAS-Tools and compass but MPAS-Model has been renormalizing the sphere radius to its own internal value.

It would still be good if MPAS-Ocean's internal Earth radius were made the same as CIME's because some parameterizations use the Earth radius form MPAS-Model.  But with this change, the mesh should retain the CIME Earth radius.

vanroekel (approved):
looks good to me

mark-petersen (approved):
Looks good, thanks.

xylar:
Thanks @mark-petersen and @vanroekel!

PR #97, opened 2021-04-30
Add z-level versions of the sub-ice-shelf 2D test case

xylar:
These tests were used to test and debug the new `minLevel*` additions in https://github.com/MPAS-Dev/MPAS-Model/pull/825

* `z_level_full_cells_const_S` -- a test case with constant T and S, full top and bottom cells, and the JM equation of state.  Should still remain stationary to nearly machine precision
* `z_level` -- the same test case but with stratified S (and rho) and partial top and bottom cells.  The partial cells will induce further pressure gradient errors and a small flow.

`z_level` has been added to the `nightly` regression suite.


xylar:
@cbegeman, would you be willing to review this?

ocean/sub_ice_shelf_2D/5km/z_level/config_forward.xml
```diff
@@ -0,0 +1,49 @@
...
+		<option name="config_disable_vel_vmix">.true.</option>
+		<option name="config_disable_vel_vadv">.true.</option>
+		<option name="config_disable_tr_all_tend">.true.</option>
+		<option name="config_disable_tr_adv">.true.</option>
+		<option name="config_disable_tr_hmix">.true.</option>
+		<option name="config_disable_tr_vmix">.true.</option>
+		<option name="config_disable_tr_sflux">.true.</option>
+		<option name="config_disable_tr_nonlocalflux">.true.</option>
+		<option name="config_disable_vel_vadv">.true.</option>
+		<option name="config_disable_redi_k33">.true.</option>
```
xylar, line 31: @cbegeman, in all these tests, do we want to leave these as disabled by default?

xylar:
@cbegeman, do you think all 4 test case are useful, or should we take out some?

xylar:
## Testing

I ran all 4 test cases successfully under Ubuntu with gnu compilers.

ocean/sub_ice_shelf_2D/5km/z_level/config_forward.xml
cbegeman: No, I think we should enable all (set to `.false.`).

cbegeman:
> @cbegeman, do you think all 4 test case are useful, or should we take out some?

No, I don't think we need all 4. What do you think about keeping `z_level` and `z_level_full_cells_constS`?

ocean/sub_ice_shelf_2D/5km/z_level/config_forward.xml
xylar: Okay, which means just taking all of these out.  No need to set them to the default.

xylar:
Sure, I would agree that those 2 are probably good enough.

ocean/sub_ice_shelf_2D/check_bit_for_bit.py
```diff
@@ -1,15 +1,21 @@
 #!/usr/bin/env python
```
cbegeman, line 1: @xylar I think we can remove this file. If I remember correctly, it pre-dated `check_z_level_bit_for_bit` and we shouldn't need both.

cbegeman (approved):
I ran both test cases on Grizzly with intel compilers as well as `check_z_level_bit_for_bit` and `plot_sub_ice_shelf_2d`. This all looks good to me. Thanks @xylar!

ocean/sub_ice_shelf_2D/check_bit_for_bit.py
xylar: Okay, I removed it and cleaned up the commit history.

xylar:
I retested and both test cases (as well as the nightly regression suite) work.  Will merge as soon as CI has finished.

PR #99, opened 2021-04-30
Turn off surface restoring in global ocean

xylar:
We have discussed that we do not want to have surface restoring as part of dynamic adjustment anymore.


xylar:
## Testing

I ran the `qu240_for_e3sm` suite on my laptop (Ubuntu, gnu) and it ran successfully.

I verified that `config_use_activeTracers_surface_restoring = .false.` in both steps of the `dynamic_adjustment` test case.

xylar:
@vanroekel and @mark-petersen, since this is a one-line change, all I'm looking for in a review is your approval that we do, indeed, want to make this change.

vanroekel (approved):
confirming this is a change we want for MPAS spin ups

mark-petersen (approved):
Yes, this is fine, thanks.

xylar:
Thanks @mark-petersen and @vanroekel!

PR #100, opened 2021-04-30
Automatically read test-case config files

### LINKED ISSUE #72: Automate config files in test cases
It would be nice if config files from test cases were loaded automatically loaded without needing a `configure()`.  This should mean a lot of test cases don't need `configure()`.

config files for MPAS cores and test groups are already loaded automatically as part of setting test cases in the work directory.  There's no reason the same couldn't be the default behavior for test cases.  And it would be an easy update.

xylar:
This means test cases don't need to override the `configure()` method if all they want to do is load the config file `<test_case>.cfg` in the test-case package (directory).

Land-ice and ocean test case and test groups have been updated to no longer load the test case's config file (since this will have already happened automatically).  For now, this doesn't reduce code very much because in all of these cases, the `configure()` method (or shared configure function at the test group level) did more than just load the test case's config options.  But it is likely to make things simpler in the future.

closes #72

### DIFF
--- a/compass/landice/tests/enthalpy_benchmark/A/__init__.py
+++ b/compass/landice/tests/enthalpy_benchmark/A/__init__.py
@@ -52,9 +52,6 @@ class A(TestCase):
         """
         Modify the configuration options for this test case
         """
-        add_config(self.config, 'compass.landice.tests.enthalpy_benchmark.A',
-                   'A.cfg', exception=True)
-
         with path('compass.landice.tests.enthalpy_benchmark', 'README') as \
                 target:
             symlink(str(target), '{}/README'.format(self.work_dir))
--- a/compass/landice/tests/enthalpy_benchmark/B/__init__.py
+++ b/compass/landice/tests/enthalpy_benchmark/B/__init__.py
@@ -38,9 +38,6 @@ class B(TestCase):
         """
         Modify the configuration options for this test case
         """
-        add_config(self.config, 'compass.landice.tests.enthalpy_benchmark.B',
-                   'B.cfg', exception=True)
-
         with path('compass.landice.tests.enthalpy_benchmark', 'README') as \
                 target:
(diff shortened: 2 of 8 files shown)


xylar:
## Testing

I ran the `sia_integration` and `nightly` test suites against a baseline from `master` on my laptop (Ubuntu, gun).  All tests pass (except the usual failure in the RK4 restart test unrelated to this PR).

I verified that all affected config files were identical between the baseline and the work directory produced from this branch.

xylar:
@matthewhoffman, this was a feature you suggested in our discussion during your review of #28.  Would you be willing to review?  It would be sufficient for you to just review by inspection, relying on my tests that the change performs as expected.  It would be great if you wanted to run tests of your own.

matthewhoffman (approved):
@xylar , this looks nice in making the test case .cfg behavior match that of the other objects.  I see how in practice it only eliminates a bit of code, but I still like the symmetry of it.  I'm approving by inspection.

I have a minor suggestion to somewhere document when/how this bit of 'magic' happens.  One possibility is in the description of the testcase class' `configure` method, so that if someone goes to override it, they'll be reminded of how the .cfg loading occurs:
```
diff --git a/compass/testcase/__init__.py b/compass/testcase/__init__.py
index 69dcc354..edf3e0b7 100644
--- a/compass/testcase/__init__.py
+++ b/compass/testcase/__init__.py
@@ -113,7 +113,9 @@ class TestCase:
         Modify the configuration options for this test case. Test cases should
         override this method if they want to add config options specific to
         the test case, e.g. from a config file stored in the test case's python
-        package
+        package.  If a test case overrides this method, it should assume that
+        the <test_case.name>.cfg file in its package has already been added
+        to the config parser prior to the `configure` method call.
         """
         pass
```

xylar:
@matthewhoffman, thanks for the review.  I have added this to the docstring and also something similar to the documentation of `configure()`.

PR #101, opened 2021-04-30
Allow `add_model_as_input()` to be called in `__init__()`

### LINKED ISSUE #73: add_model_as_input() should be callable from init()
It would be nice if `Step.add_model_as_input()` could be called from init like other methods (e.g. `Step.add_input_file()`) instead of in `Step.setup()`.  That would save a lot of steps from needing a `setup()` method at all.

xylar:
This is more convenient, but means that the model shouldn't actually be added until the necessary config option is available after setup.

This merge moves `add_model_as_input()` to each test cases' `__init__()` 

The docs are updated based on this change.

closes #73


xylar:
I don't think anyone needs to review this one.  It's a pretty straightforward bit of clean up.

PR #103, opened 2021-04-30
Add a `package` argument to `add_input_file()`

### LINKED ISSUE #75: Add `package` argument to `add_input_file()`
This would save some trouble of having to use `importlib.resources` in individual steps, and fits the style of URLs and databases already handled by this method.

xylar:
This argument is used to make a symlink to a file in the ``compass`` package and set that file as an input to the step.

With this change, it's a lot easier to link files from the package in a similar manner to how we are used to doing this in legacy COMPASS.  You can link geojson files, READMEs or even python scripts that a user should optionally run (e.g. for quick and dirty viz).

The ocean and landice cores have been updated with this change, as has the documentation.

closes #75

### DIFF
--- a/compass/landice/tests/enthalpy_benchmark/A/__init__.py
+++ b/compass/landice/tests/enthalpy_benchmark/A/__init__.py
@@ -1,7 +1,6 @@
 from importlib.resources import path
 
 from compass.io import symlink
-from compass.config import add_config
 from compass.validate import compare_variables
 from compass.landice.tests.enthalpy_benchmark.setup_mesh import SetupMesh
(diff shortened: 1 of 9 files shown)


xylar:
## Testing

I ran the `sia_integration` test suite on Ubuntu with gnu, which passed against a baseline of master from earlier today.

I verified that symlinks for the geojson files for the SOwISC12to60 mesh are, indeed, valid.  I didn't run that test case because it's quite time consuming.

xylar:
I don't think this is one where anyone needs to review.

PR #104, opened 2021-04-30
Alter `steps_to_run` in config file or command line

### LINKED ISSUE #80: Ability to view and alter steps within run command
In the https://github.com/MPAS-Dev/compass/tree/legacy format, it is obvious to the user that the run steps are contained in run.py, and it is easy to copy from that script to the command line, or to only run parts of it.

In this new setup, it is not obvious how we see and alter the steps that occur in the command 
```
python -m compass run
```

See conversation at #76

xylar:
This merge adds the ability for users to choose which steps in a test case they want to run.  They have 2 options.

First, they can alter the `steps_to_run` config option in the `test_case` section at the very top of the test case's config file:
```
[test_case]
steps_to_run = initial_state full_run restart_run
```
The list of `steps_to_run` can be separated by spaces, commas or both.

Or they can specify which steps to run (or which steps not to run) via the ``--steps`` (or ``--no-steps``) flag to ``compass run``:

```
python -m compass run --steps initial_state full_run
python -m compass run --no-steps restart_run
```
Both of these accomplish the same thing -- skipping the ``restart_run`` step of the test case.

The new capability has been added to the documentation.

closes #80

### DIFF
--- a/compass/run.py
+++ b/compass/run.py
@@ -168,11 +168,15 @@ def run_test_case():
     config.read(test_case.config_filename)
     test_case.config = config
 
+    steps = config.get('test_case', 'steps_to_run').replace(',', ' ').split()
+    test_case.steps_to_run = steps
+
     # start logging to stdout/stderr
     test_name = test_case.path.replace('/', '_')
     test_case.new_step_log_file = True
     with LoggingContext(name=test_name) as logger:
         test_case.logger = logger
+        logger.info('Running steps: {}'.format(', '.join(steps)))
         test_case.run()
         test_case.validate()
 
--- a/compass/run.py
+++ b/compass/run.py
@@ -155,10 +155,20 @@ def run_suite(suite_name):
             sys.exit(1)
 
 
-def run_test_case():
+def run_test_case(steps_to_run=None, steps_not_to_run=None):
     """
     Used by the framework to run a test case when ``compass run`` gets called
     in the test case's work directory
+
+    Parameters
+    ----------
+    steps_to_run : list of str, optional
+        A list of the steps to run.  The default behavior is to run the default
+        steps unless they are in ``steps_not_to_run``
+
+    steps_not_to_run : list of str, optional
+        A list of steps not to run.  Typically, these are steps to remove from
+        the defaults
     """
     with open('test_case.pickle', 'rb') as handle:
         test_case = pickle.load(handle)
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I have run the `nightly` test suite on Ubuntu with gnu, and everything passed and is bit-for-bit with master from earlier today.

I also verified that I could modify the config option `steps_to_run` and only the desired steps ran.  Similarly, I verified that I could both explicitly include steps with the ``--steps`` flag and exclude them with the ``--no-steps`` flag.  I verified that these flags have no effect when calling `compass run` to run a suite or within a step work directory.

xylar:
@mark-petersen  requested something like this in #80.  This is my suggestion.

@mark-petersen and @matthewhoffman, if you are willing to review, I would appreciate a little testing.  And please let me know if you are happy with this solution or if you have alternative suggestions or changes.

xylar:
@mark-petersen,

> One of my common workflows is to run the full nightly regression suite once, then rerun it with other compilers or debug, openmp settings, but turn off the QU240 mesh and init to save time. Is there a way to do that? I think this PR does not address omitting a test case from a suite altogether.

My suggestion would be to make version of the nightly suite with and without the test you don't want to run.  I can't think of any way to efficiently specify that you don't want to run a particular step but you could open an issue with a suggested method for doing that.

> I tried it with this functionality, ie
...
with a blank, but it ran that step anyway. The suite function must override the config setting.

Hmm, I can look into that.  If `steps_to_run` isn't getting processed during test-suite runs, that isn't intentional.  I'll get back to you..

xylar:
@mark-petersen, indeed, `steps_to_run` wasn't getting parsed for test cases run out of a test suite.  I have fixed this now and run the `nightly` suite once with with `mesh` step and once without it, as you suggested.  It worked for me.  Could you verify that it works for you?  

Please also comment on #114 so I can make sure to address your needs there.

docs/developers_guide/command_line.rst
```diff
@@ -213,11 +213,39 @@ that has been set up in the current directory:
...
+    python -m compass run --steps initial_state full_run
+
+or
+
+.. code-block:: none
+
+    python -m compass run --no-steps restart_run
+
+Would both accomplish the same thing in this example -- skipping the
+``restart_run`` step of the test case.
```
matthewhoffman, line 251: Add note that using command line option will override the cfg version.

compass/run.py
```diff
@@ -144,11 +157,22 @@ def run_test_case():
     config.read(test_case.config_filename)
     test_case.config = config
 
+    if steps_to_run is None:
+        steps_to_run = config.get('test_case',
+                                  'steps_to_run').replace(',', ' ').split()
+
+    if steps_not_to_run is not None:
+        steps_to_run = [step for step in steps_to_run if step not in
```
matthewhoffman, line 165: Check if step name specified is actually a valid step name or not.  (This should be done for both `--steps` and `--no-steps`.)

matthewhoffman (changes requested):
@xylar , this looks great.  I added inline comments for the 2 items we discussed on the phone.

xylar:
@matthewhoffman, thank you for both of your excellent suggestions.  I believe they are now addressed.  Please give this another look when you have a chance.

mark-petersen (approved):
Looks good. Thanks for these improvements.

xylar:
@matthewhoffman, still waiting on a re-review here.  Is that something you can do soon?

matthewhoffman:
I'm looking at this again now.  I got myself confused because I hadnt realized it was rebased on the dev env.

matthewhoffman (approved):
Looks good @xylar.  Thanks for adding the additional error-checking.  I tried all the ways of asking it to do incorrect things, and it gave sensible errors each time.

xylar:
Thanks!

PR #105, opened 2021-05-03
Add WC14 mesh to global_ocean

xylar:
This includes mesh, init, performance test, dynamic adjustment and files-for-E3SM test cases.

Settings are intended to be as close as possible to those from the ``legacy`` branch.


xylar:
## Testing

I'm running the `wc14` test suite on Chrysalis with Intel right now.

Mesh plots (same as before):
cellWidthGlobal
mesh_construction

I'll update as soon as the suite has run (or not).

xylar:
@mark-petersen and @vanroekel, this is ready to review.  If you want to take a look at my output, that would be wonderful.  I don't think we need to spend a ton of time on this for now, though.  If we wanted to recreate the WC14 mesh, this PR would at best do an approximate job of it. The main thing is that the pieces are in place to make a new WC mesh in the future based on WC14 if we need to.

vanroekel (approved):
@xylar I took a close look at the mesh (in particular critical passages) and initial T&S fields with the preview converter and all looks good.  I also looked through the directory structure and namelists.  All looks great, I'm assuming `config_use_activeTracers_surface_restoring` will get set to false once #99 is merged

Thanks @xylar!

xylar:
> I'm assuming `config_use_activeTracers_surface_restoring` will get set to false once #99 is merged

Yes, that's right.  Just waiting for @mark-petersen's approval on that one.

xylar:
I need to add this mesh to the documentation.

xylar:
Documentation has now been added.

mark-petersen (approved):
Looks great thanks! I set up the initial conditions with compass on grizzly and started the mesh generation. I did not do the whole thing, but I'm approving based on your posts.

xylar:
Great, I'll merge this as soon as I get a chance to add documentation.  Thanks @vanroekel and @mark-petersen for your reviews.

PR #108, opened 2021-05-04
Add merry-go-round test case for vertical and horizontal advection

scalandr:
This test case is meant to analyze the accuracy of the tracers' horizontal and vertical advection. The horizontal and vertical velocity profiles are constant in time, therefore RK4 is used as time-stepping scheme. The meshes used are planar with 100, 200 and 400 cells per y-layer. The numbers of vertical layers is 50, 100 and 200 for the 100, 200 and 400 cell meshes, respectively. It is a steady state test case, so the initial condition is the analytical solution which should remain over all time integration. For each mesh, plots of the horizontal velocity, vertical velocity and tracer fields after a 6 hour run time are created. Finally, a convergence plot that shows second-order convergence with increasing grid refinement is created.


scalandr:
Here the plots associated to the test case: 

horizontal velocity, vertical velocity and tracer field for the 100 cell case: 
100Cells
horizontal velocity, vertical velocity and tracer field for the 200 cell case: 
200Cells
horizontal velocity, vertical velocity and tracer field for the 400 cell case: 
400Cells
convergence plot:
order

scalandr:
I did a commit after fixing PEP8 issues.

xylar:
@mark-petersen and @scalandr, could you have a conversation about whether this test case still makes sense of `legacy`, if you want to port it to the new approach in `master`, or if you want to close this for now and revisit it later?

mark-petersen (approved):
Yes, this looks good to me. I tested all three resolutions and the analysis. The python init step is a bit slow for the number of cells - 4 minutes for the 800 cell case - but we don't run it very often, so it's OK. The section.png plots looked exactly like yours. Thanks @scalandr!

PR #110, opened 2021-05-04
Handle test cases that are not packages

xylar:
Prior to this fix, if you have a `TestCase` child class that is in a module but not a package, it causes a `TypeError` to be raised when looking for a config file in the package (there is no package!).  With this change, we simply catch the exception and move on.

### DIFF
--- a/compass/config.py
+++ b/compass/config.py
@@ -52,7 +52,7 @@ def add_config(config, package, config_file, exception=True):
     try:
         with resources.path(package, config_file) as path:
             config.read(path)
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran into this when working on the `cosine_bell` test cases.  It is convenient to have an `AnalysisTest` test case in an `analysis.py` module, but it caused trouble.  With this fix, I can set up the test case without trouble.

xylar:
Another 1-line fix that I don't want to bug folks about reviewing or testing.

PR #112, opened 2021-05-05
`compass run` runs a single suite without arg

xylar:
With this merge, if you only set up a single test suite in a work directory, `compass run` will work without the need to specify the suite name.  Since this is by far the most common work flow, this change should save a little trouble nearly every time a test suite is run.

### DIFF
--- a/compass/run.py
+++ b/compass/run.py
@@ -5,6 +5,7 @@ import pickle
 import configparser
 import time
 import numpy
+import glob
 
 from mpas_tools.logging import LoggingContext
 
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I set up the nightly test suite on Chrysalis and verified that I could run it with `python -m compass run` (without the suite name).  I removed the `nightly.pickle` file and verified that I got the expected error message.  I set up the `qu240_for_e3sm` and `nightly` test suites and verified that I got the expected error message when I didn't specify a suite, but that each suite ran if I specified the name explicitly.

xylar:
@mark-petersen and @matthewhoffman, for a review, all I'd like from you is a quick look at the code to make sure you're okay with the change.  Let me know if you'd like to have the error messages reworded.

mark-petersen (approved):
This is great! I gave it a quick test. Yes, this is handy, as it simplifies the command line argument.

matthewhoffman (approved):
Nice usability improvement, @xylar.  I should have taken a look sooner given how easy this was to review.

xylar:
@matthewhoffman and @mark-petersen, thanks for reviewing this.

PR #111, opened 2021-05-04
Add cosine bell convergence test

xylar:
This is in the `global_convergence` test group (called `convergence_global` in `legacy`).

PR #113, opened 2021-05-05
Two small fixes to the `setup` module

xylar:
This came up in reorganizing #111.

First, it is convenient to add the custom config file to the config options both before and after calling the test case's `configure()` method.  It turns out that you can use config options to determine which steps to add to a test case if you do this.  This is really convenient for defining a flexible parameter study like the convergence test in #111.  You can make the parameter values a config option.  The user could make a custom config file and alter the config options there before setting up the test case.  Depending on how the test case is set up, altering the config options between setup and run might either have no effect (as in #111) or cause you trouble, so that would be something to document as part of each test case that uses this approach.

The second thing is that the symlink in each step to the test case's config file in the work directory was previously just `../<config>.cfg`.  But `compass` allows steps have a more complex directory structure, we just hadn't tried it before.  This merge makes the symlink using an absolute path to the test case's config file, which means we don't have to figure out the relative path between the step and the test case.  Since it isn't really possible to move a work directory without breaking things anyway (i.e. there are already symlinks with absolute paths within the work directory for other purposes), I think this is not a problem.

### DIFF
--- a/compass/setup.py
+++ b/compass/setup.py
@@ -208,7 +208,7 @@ def setup_case(path, test_case, config_file, machine, work_dir, baseline_dir,
         except OSError:
             pass
 
-        symlink(os.path.join('..', test_case_config),
+        symlink(os.path.join(test_case_dir, test_case_config),
                 os.path.join(step_dir, test_case_config))
 
         step.work_dir = step_dir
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the `nightly` suite on Chrysalis with Intel, using `master` as a baseline.  All tests (except the usual RK4 restart) passed.

I verified in a few steps that the links to the config file were good and that they have absolute paths.

xylar:
@mark-petersen or @matthewhoffman, let me know if you would like to review this or if you're fine with me merging based on my testing.

mark-petersen:
@xylar this is fine, I'll trust your judgement on this.

PR #115, opened 2021-05-12
Copy a list of tests in a suite to work dir 

### LINKED ISSUE #114: Need a method to specify which tests of a test suite to run (or not run)
It would be good to have a way for a user to alter the tests in a test suite after setup without having to edit the code and setup again.

@mark-petersen [wrote](https://github.com/MPAS-Dev/compass/pull/104#issuecomment-836809844):
> One of my common workflows is to run the full nightly regression suite once, then rerun it with other compilers or debug, openmp settings, but turn off the QU240 mesh and init to save time. Is there a way to do that? I think this PR does not address omitting a test case from a suite altogether.

xylar:
This list can be edited to remove or comment out (with `#`) the tests not to run.

The user's guide has been updated to mention this option.

closes #114

### DIFF
--- a/compass/run.py
+++ b/compass/run.py
@@ -6,6 +6,7 @@ import configparser
 import time
 import numpy
 import glob
+import warnings
 
 from mpas_tools.logging import LoggingContext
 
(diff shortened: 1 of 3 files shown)


xylar:
Based off of #112, so needs to be rebased after that gets merged.

xylar:
## Testing

I set up the `nightly` suite, then removed many tests and commented out others.  The remaining tests ran as expected.

I added a fake test (`blah`) and got the expected error message.

I removed `nightly.txt` and verified that I got the expected warning.

xylar:
@mark-petersen, please let me know if this addresses your concern in #114.

Issue #57, opened 2021-03-12
Compy srun model steps are hanging


xylar:
I tried to run compass 1.0 on compy and running the model itself just hangs.  This is an issue to track the problem as I try to debug it.

xylar:
@mattdturner, as a compy user, this might be one you want to investigate if you want to do compass runs there.  In my testing, I get to the `srun -n 4 ./ocean_model ...` step and it just hangs indefinitely.

xylar:
I figured this out as part of work on E3SM-Unified.  On Compy, `srun` must be called with `--mpi=pmi2`.  I will make a PR to fix this as soon as I have time.

PR #117, opened 2021-05-14
Significantly change how to create development conda environments

### LINKED ISSUE #91: Investigate using `pip install -e`
In this conversation about CIME:
https://github.com/ESMCI/cime/issues/3886#issuecomment-829456896
@jhkennedy made clear that there are better ways for developers to run `compass` as a package while still developing.

First, we have a yaml file that defines the development environment that users can easily install for themselves, e.g.:
```
conda env create -f ./environment.yml
```
This way, we don't have do maintain a list of dependencies in the documentation itself.  We would still need to have a yaml file for the development environment and another for the conda recipe, as far as I can tell.

Second, we should install the package to our own test environment in edit mode with:
```
python -m pip install -e .
```
This way, we can just do `conda list`, etc. instead of `python -m conda list` but you can still edit the code (in the repo) and have it update in the conda environment (which just points to the repo as I understand it).

xylar:
This merge makes significant changes to how development conda environments are created and activated, following lessons learned from #91.  

These changes include:
* most critically, adding a new `conda/create_compass_env.py` script for developers to use to create and update their *own* development environments (no longer using a shared one for all developers)
* removing `load` and `deploy` directories
* adding a `conda` directory with a `compass_env` subdirectory
* adding `conda/compass_env/spec-file-*.txt` package specs that defines the development environments
* adding other templates and support files to `conda/compass_env/`.
* updating the documentation (fairly extenstively) for these changes.

The gist for reviewers is that:
* you would now have your own Miniconda3 installed on each machine.  `create_compass_env.py` takes a `--conda` flag for the Miniconda3 base directory (typically `${HOME}/miniconda3`) and if you specify a new location, it will install `miniconda3` for you there.
* we now only share the compiled SCORPIO and ESMF libraries, not the development conda environment
* to create a new development environment and activation script, you will run:
  ```
  ./conda/create_compass_env.py --conda ~/miconda3 --compiler {intel18,intel,gnu,pgi}
  ```
  A conda environment and an activation script for the compiler (and default MPI version) will be created for you.  The activation script is `test_compass_1.0.0_<machine>_<compiler>_<mpi>.sh` in the compass branch.  The modules for the activation script are pulled right out of E3SM config files.  I have made local copies so we update them at compass version changes, not necessarily when E3SM changes them.
* your local compass branch has been installed to the conda environment (using `python -m pip install -e .`) so any changes you make will be visible via the `compass` command-line tool from the conda package without you needing to do anything special.  In other words, a development environment looks exactly like a user's environment from a user's perspective.  But there is no need to build a conda package.
* run again with another compiler to get a new activation script (sharing the same conda environment)
* run again in a new `worktree` folder with a new branch to update the conda environment to point to your new branch (and to generate an activation script there.
* there are no longer symlinks to `compass` in the work directories because they aren't needed
* there is, however, still a symlink to `load_compass_env.sh`, which points to the `test_compass_1.0.0_*.sh` that you used to set up the test case or suite
* on your local Linux or OSX machine, things should be very similar.  The main difference is that you don't have a choice of compiler, you have a choice of MPI:
  ```
  ./conda/create_compass_env.py --conda ~/miconda3 --mpi {mpich,openmpi}
  ```
  NetCDF, etc. comes from `conda-forge`.  I built PNetCDF and it is up on the `e3sm` channel, and that gets pulled in (I hope to have it on conda-forge eventually).
* on both supported and personal machines, if you need to, you can specify some extras like the environment name, python version, specifying that you want some safety checks, and forcing the script to recreate the environment (in case it's failing to update).

This is a lot, so I'll chat with you in person, too.

closes #91


xylar:
@mark-petersen, @vanroekel and @matthewhoffman, this is a big enough change that I need to chat with each of you about it (presumably both before and after you've had a look at it).  Would you like one-on-one meetings or should I try to arrange a time for all 4 of us? Ping me on Slack if you like.

This is a reasonably high priority for me because it will change the basic workflow for all of us and I would like that to happen before any new users are added.

xylar:
If you want to build the docs and read them, you can do the following on your Linux or OSX machine:
```
./conda/create_compass_env.py --conda ~/miconda3 --mpi mpich
source test_compass_1.0.0_mpich.sh 
cd docs
make html
google-chrome _build/html/index.hml
```
The development environment already has everything in it you need to build the documentation.

xylar:
@vanroekel, I had a chat with @matthewhoffman and @mark-petersen about this one yesterday and would like to do the same with you in the next week or 2.  This change is a big one but would make things a ton easier for both me and other developers if we can switch over to this workflow soon.

mark-petersen:
I gave this a try on grizzly. I must be missing something. I did this in the top level:
```
./conda/create_compass_env.py --conda ~/miconda3 --compiler gnu
sh test_compass_1.0.0_grizzly_gnu_mvapich.sh
```
but the (test_compass_1.0.0) doesn't show up on my prefix. That's probably the problem, because I can't get it to run then:
```
ls
azure-pipelines.yml  compass           conda       config.ocean  E3SM             LICENSE   MPAS-Model  setup.py
ci                   compass.egg-info  Config_dir  docs          example_configs  MALI-Dev  README.md   test_compass_1.0.0_grizzly_gnu_mvapich.sh

compass list --suites
-bash: compass: command not found

compass suite -s -c ocean -t nightly -m grizzly -w $n/210516_test -p $b/ocean_develop
-bash: compass: command not found
```

xylar:
> `sh test_compass_1.0.0_grizzly_gnu_mvapich.sh`

This needs to be:
```
source test_compass_1.0.0_grizzly_gnu_mvapich.sh
```
Just running the script will load all the modules in a separate bash process and then return to your original environment once the script ends.  This is the equivalent of ```source load_latest_e3sm_unifled.sh```, etc.  You would have the same issue if you tried to run those scripts insead.

xylar:
I just looked through the documentation and I think it's correct but let me know if something incorrectly used `sh` instead of `source`.

mark-petersen:
OK thanks. For some reason `sh` and `source` seemed the same to me before. I was able to do this on grizzly:
```
./conda/create_compass_env.py --conda ~/miconda3 --compiler gnu
source test_compass_1.0.0_grizzly_gnu_mvapich.sh
```
then compiled the ocean model, and then these worked
```
compass list
compass suite -s -c ocean -t nightly -m grizzly -w $work_dir -p $compile_dir
```
then on an interactive node, I again
```
source test_compass_1.0.0_grizzly_gnu_mvapich.sh
compass run nightly
```
and it all worked, except the global ocean cases failed during the run for some reason. May be unrelated to this PR.

So I tried some other compiler/mpi combinations, but these commands failed:
```
./conda/create_compass_env.py --conda ~/miconda3 --compiler gnu -i openmpi
```
<details>
<summary> which ended with the error: </summary> <p>

```
...
Scanning dependencies of target spio_finfo.exe
[ 86%] Building CXX object tools/spio_finfo/CMakeFiles/spio_finfo.exe.dir/__/util/argparser.cxx.o
[ 89%] Building CXX object tools/spio_finfo/CMakeFiles/spio_finfo.exe.dir/__/util/spio_misc_tool_utils.cxx.o
[ 91%] Building CXX object tools/spio_finfo/CMakeFiles/spio_finfo.exe.dir/__/util/spio_lib_info.cxx.o
[ 93%] Building CXX object tools/spio_finfo/CMakeFiles/spio_finfo.exe.dir/spio_file_test_utils.cxx.o
[ 95%] Building CXX object tools/spio_finfo/CMakeFiles/spio_finfo.exe.dir/spio_finfo.cxx.o
[ 97%] Building CXX object tools/spio_finfo/CMakeFiles/spio_finfo.exe.dir/spio_finfo_tool.cxx.o
[100%] Linking CXX executable spio_finfo.exe
[100%] Built target spio_finfo.exe
[ 50%] Built target pioc
[ 50%] Built target genf90
[ 84%] Built target piof
[100%] Built target spio_finfo.exe
Install the project...
-- Install configuration: ""
CMake Error at src/clib/cmake_install.cmake:41 (file):
  file cannot create directory:
  /usr/projects/climate/SHARED_CLIMATE/compass/grizzly/system/compass_1.0.0/gnu/openmpi/scorpio_1.1.6/lib.
  Maybe need administrative privileges.
Call Stack (most recent call first):
  src/cmake_install.cmake:42 (include)
  cmake_install.cmake:42 (include)


make: *** [install] Error 1
  File "./conda/create_compass_env.py", line 882, in <module>
    main()
  File "./conda/create_compass_env.py", line 860, in main
    sys_info, system_libs = build_system_libraries(
  File "./conda/create_compass_env.py", line 537, in build_system_libraries
    check_call(command)
  File "./conda/create_compass_env.py", line 633, in check_call
    raise subprocess.CalledProcessError(proc.returncode, commands)
subprocess.CalledProcessError: Command '/bin/bash build.bash' returned non-zero exit status 2.
  File "./conda/create_compass_env.py", line 882, in <module>
    main()
  File "./conda/create_compass_env.py", line 840, in main
    do_bootstrap(args.no_bootstrap, conda_base, activate_base)
  File "./conda/create_compass_env.py", line 133, in do_bootstrap
    check_call(command)
  File "./conda/create_compass_env.py", line 633, in check_call
    raise subprocess.CalledProcessError(proc.returncode, commands)
subprocess.CalledProcessError: Command 'source /users/mpeterse/miconda3/etc/profile.d/conda.sh; conda activate; ./conda/create_compass_env.py --conda /users/mpeterse/miconda3 --compiler gnu -i openmpi --no-bootstrap' returned non-zero exit status 1
```
</details>

Looks like I need write privileges to
```
/usr/projects/climate/SHARED_CLIMATE/compass/grizzly/system/compass_1.0.0/gnu/openmpi/scorpio_1.1.6/lib
```

I really like this new interface that combines modules with the conda environment, and can use the local compass code directly. But like modules, we probably need to work on the kinks of which combinations of compiler and libraries are already tested on each machine.

I just tried:
```
./conda/create_compass_env.py --conda ~/miconda3 --compiler intel -i openmpi
```
and got a similar error.

xylar:
@mark-petersen, thanks for testing.  You bring up some really good points.

Up to now, I have only built with one (default) MPI flavor per compiler on each machine.  

I have also set up the script so it chmods the directory where ESMF and SCORPIO get built so only the person running the script has write permission.

There are a few options for handling this, and maybe others I haven't thought of:
1. Only support certain compiler/MPI combinations on each supported machine.  There would need to be an easy way for users to know which (probably the docs) and errors (or at least serious warnings) if you try for an unsupported combination one.  I would need to build all supported combinations.  Permissions would remain as they are (only I can write).
2. I could build all possible compiler/MPI combinations on all supported machines every time we update the supported compilers, MPI versions or other system modules.  That sounds like a lot of work but it could be pretty automated so it's not as big a deal as it sounds (as long as thing work).  Permissions would remain as they are (only I can write).
3. We open up permissions on the shared system libraries.  Anyone from the project can use the script to build missing ESMF and/or SCORPIO.  There is a small risk that the build succeeds but the install fails somehow, or that the script is misconfigured somehow and the shared space gets messed up that way.  Based on past experience, there is also a non-negligible risk that our team members manage to bungle up the shared space in unexpected ways.  But, unlike E3SM-Unified which was the basis for the deployment script, there's not too much damage anyone can do.

I feel like 3 is the way to go.  I would build the default compiler/MPI combinations for each release but if folks want to try out others, that's possible.  

If we discover that certain combinations don't work, there should be a way to indicate this.  Maybe just a file `unsupported.txt` that has entries like:
```
grizzly, intel, openmpi
anvil, gnu, mvapich
```
Then, the script could issue an error if a developer tries one of those combinations.

xylar:
@mark-petersen (cc @matthewhoffman), based on your comment, I did implement option 3 above.  You should now have permission to install system libraries on grizzly, badger, anvil and chrysalis if they are missing (however, I have built all supported combinations on these machines).  I haven't updated thing on cori yet and I don't plan to work on compy yet because standalone MPAS doesn't work there (#57).

I added a file `conda/unsupported.txt` for keeping track of combinations that don't work.  This includes `gnu` with `impi` on grizzly and badger (To my surprise, this combination *does* work on chrysalis!), as well as `intel` (as opposed to `intel18`) and `mvapich` on anvil (I'm guessing intel 17 is just too old for the latest ESMF).  As we discover other combinations that fail, we should add them to the list to save time.

Let me know how it goes.

mark-petersen:
I'm still struggling with this. On a fresh login node I use
```
./conda/create_compass_env.py --conda ~/miconda3 --compiler gnu -i openmpi
source test_compass_1.0.0_grizzly_gnu_openmpi.sh
```
Then, without loading any other modules, I compile the code with `make gfortran OPENMP=true` then 
```
compass suite -s -c ocean -t nightly -m grizzly -w $work_dir -p $compile_dir
```
Then `salloc` into compute nodes, again `source test_compass_1.0.0_grizzly_gnu_openmpi.sh`, cd to work directory, and `compass run`. Executable runs but ends with a seg fault.

Maybe I have my executable or namelists mixed up. It appears successful until the very end here.

xylar:
@mark-petersen, it's also possible that some combinations compile but don't work right.  Particularly on IC, we've had that problem.  It could be that we need to mark `gnu` and `openmpi` as an unsupported combination, too.  I have set `gnu` and `mvapich` as the default.

xylar:
I'll try both `intel` and `gnu` on both grizzly and badger with openmpi and see if any of those work.

xylar:
@mark-petersen, both `gnu/openmpi`and `intel/openmpi` are failing for me on Grizzly. I'm not optimistic that things will be different on badger but I'll test just in case.  Then, I'll just mark that combination as unsupported.  Obviously, these machines don't get full support from the E3SM team so it's not really surprising that not all available combinations actually work.

mark-petersen (approved):
@xylar this is great! I tested successfully with default mpi with both gnu and intel on both grizzly and cori. I like your recent changes of listing supported combinations of configurations.

I'm ready to make the jump here. I think the functioning is no more hidden than the standard 'module load' commands, and then we can take advantage of the machine support by E3SM.

mark-petersen:
A few questions and comments:
1. Once you run e.g. `source test_compass_1.0.0_grizzly_gnu_mvapich.sh` then the `compass` command is available everywhere. If I go into a different compass directory where I am, say, developing a new test case, and run compass, am I running the code in the first or second directory?
2. I typically develop in a bunch of terminal tabs all at once. It would be easy to get confused which compass directory I ran `source` in. Is there a way for the compass command to tell me?
3. Similarly, I might also have different tabs sourced with intel and gnu versions. Can I tell which compass was loaded afterwards? I know I could use `module list` or unix `history`, but I wonder if there is a `compass status` or similar command that would return `test_compass_1.0.0_grizzly_gnu_mvapich.sh` and the path it was loaded.
4. After this PR is merged, I assume the command will be e.g. `source compass_1.0.0_grizzly_gnu_mvapich.sh`. I think adding `load`: `source load_compass_1.0.0_grizzly_gnu_mvapich.sh` would be more intuitive.

mark-petersen:
@matthewhoffman and @vanroekel, this worked great for me, and it really simplifies the MPAS compile, test, run instructions, as there are no different sets of modules to load on each machine. The instruction set is identical for all machines!

My students begin June 1, 2 and 7. It would be great to merge this next week so we have just one simplified workflow to teach them.

xylar:
> 1. Once you run e.g. `source test_compass_1.0.0_grizzly_gnu_mvapich.sh` then the `compass` command is available everywhere. If I go into a different compass directory where I am, say, developing a new test case, and run compass, am I running the code in the first or second directory?

If you cd to a new directory, you will still be running `compass` from the old directory.  You can run either (safer):
```
./conda/create_compass_env.py ...
```
which will make sure dependencies are still the same or
```
python -m pip install -e .
```
which will only update the installation of `compass` to point to the current directory without checking anything else.  I'd like to document this option but I'm concerned about the right balance of documenting the "right" way and the "easy" way.

> 2. I typically develop in a bunch of terminal tabs all at once. It would be easy to get confused which compass directory I ran `source` in. Is there a way for the compass command to tell me?

The best thing you can do is to source the `load_compass_env.sh` in the work directory, which is a symlink to the original `test_compass*.sh` that was used to set up the test case or suite.  The alternative is to echo the `LOAD_COMPASS_ENV` environment variable, which is set in the activation script and points to the activation script itself.

> 3. Similarly, I might also have different tabs sourced with intel and gnu versions. Can I tell which compass was loaded afterwards? I know I could use `module list` or unix `history`, but I wonder if there is a `compass status` or similar command that would return `test_compass_1.0.0_grizzly_gnu_mvapich.sh` and the path it was loaded.

Again sourcing `load_compass_env.sh` in the work directory is the preferred way to handle this case, but you can use `echo $LOAD_COMPASS_ENV` to find out what script you sourced.

> 4. After this PR is merged, I assume the command will be e.g. `source compass_1.0.0_grizzly_gnu_mvapich.sh`. I think adding `load`: `source load_compass_1.0.0_grizzly_gnu_mvapich.sh` would be more intuitive.

No, that wasn't my plan.  `test` refers to the fact that it's a development environment.  `load_*` would be reserved for users and or released versions of compass.  The workflow in that situation would be more similar to E3SM-Unified and wouldn't involve anything from this PR.  I'm open to alternative names but I do really personally feel like we're all testing compass 1.0.0 until it get released so the name suits me.

xylar:
> My students begin June 1, 2 and 7. It would be great to merge this next week so we have just one simplified workflow to teach them.

I also have a student, Althea, working with me over the summer (and hopefully beyond) who would benefit from having this new approach.  I have a meeting with @vanroekel on Tuesday, so my hope is that we can move toward merging this sometime after that.  At the same time, I don't want to rush it.  I got the sense from @matthewhoffman that he wasn't ready to endorse this approach yet, which is fair enough.

xylar:
>  I like your recent changes of listing supported combinations of configurations.

@mark-petersen, what I actually implemented is a list of **unsupported** configurations.  It does seem like a list of supported configurations would be more useful in many ways.  My hope is the documentation servers that purpose to some degree (but I haven't documented all available MPI implementations because I'm assuming -- hoping at least -- that most users will be fine with the default).

docs/developers_guide/machines/index.rst
```diff
@@ -16,53 +16,57 @@ cases are configured in a way that is appropriate for your machine.
...
+|              | gnu        | mpt       | gnu-nersc         |
++--------------+------------+-----------+-------------------+
+| cori-knl     | intel      | impi      | intel-mpi         |
+|              +------------+-----------+-------------------+
+|              | gnu        | mpt       | gnu-nersc         |
++--------------+------------+-----------+-------------------+
+| grizzly      | intel      | impi      | intel-mpi         |
+|              +------------+-----------+-------------------+
+|              | gnu        | mvapich   | gfortran          |
++--------------+------------+-----------+-------------------+
```
xylar, line 69: @vanroekel, @matthewhoffman and @mark-petersen, you all had similar, excellent suggestions so here's a table of the supported compilers, MPI libraries and make targets

vanroekel: @xylar this table is a great addition, thanks for adding that.

matthewhoffman:
@xylar , this workflow change looks like it has a lot of promise, and I can see how it could allow things to run very smoothly for general developers.  I'm trying to work through the steps to understand what's happening where and why and work out wrinkles and discrepancies with how we traditionally have run MALI.  I'll describe some things I've encountered so far.

1. Creating the conda env took about 20 minutes and seemed to install about a gazillion packages.  If that's just a 1-time setup thing, that's not a big deal, but I wanted to make sure that was normal.
2. Compass commands felt a bit "heavier" than I remember.  e.g. compass list would take 5-20 seconds.  This could have been a temporary disk issue or something, but thought I'd mention it to see if that's expected.
3. The MALI build worked smoothly following this workflow when building without Albany.  Of course building with Albany requires very different modules/libs, so it will presumably take a lot of work to figure that out.  I don't think this PR should in any way be held up by that, but that will delay our full adoption of developing in compass.  We've talked about this offline, so I'll leave it at that here.
3. Related to the previous item, in working through the steps, I realized I had some confusion about what the purpose of the dev environment.  I had initially assumed it was to allow compass development, but as I understand it now, it is to provide a unified compass+mpas environment.  Is that right?  I have mixed feelings about that.  I think it would be more useful for us (at least in the near term) to have a minimal compass dev env that only has the modules/libs to run compass itself from a conda env and is unrelated to modules/libs needed by MPAS.  That would (in theory) allow landice folks to continue using our existing MALI build env for the foreseeable future until we tackle bringing Albany into the fold.  Do you think this is possible?  Could even be an undocumented thing for now.
4. Setting aside Albany, I was able to set up and launch the SIA integration test suite using the instructions above.  However, all the runs died with the pio stride error (which is presumably unrelated to this branch):
```
ERROR: PIO io_task_stride has a value of 4.
ERROR:         It must be between 1 and 1.
ERROR: PIO io_task_stride * io_task_count has a value of 4.
ERROR:         It must be between 1 and 1.
CRITICAL ERROR: Invalid PIO configuration.
```
I don't remember running into that before when I tested earlier branches, but maybe I've forgotten (or maybe I was on a different machine).  In any case, can you advise me on how to correct this?

xylar:
@matthewhoffman, I appreciate the thorough review as always.  You're a tough customer but the results are always better for it.

> Creating the conda env took about 20 minutes and seemed to install about a gazillion packages. If that's just a 1-time setup thing, that's not a big deal, but I wanted to make sure that was normal.

So, yes, this is the typical process. I just usually do it for you.  20 minutes does sound longer than normal to me, though, if you aren't building ESMF or SCORPIO.  Building ESMF takes some time for sure.  In any case, yes, this should be a one-time thing per system unless you run into trouble and need to use the `--recreate` flag to start over.  Even then, the shared system libraries won't get rebuilt until we go to a new compass version number.

> Compass commands felt a bit "heavier" than I remember. e.g. compass list would take 5-20 seconds. This could have been a temporary disk issue or something, but thought I'd mention it to see if that's expected.

The could be true. I noticed that as well.  Perhaps that's a necessary cost of using `python -m pip install -e .` instead of `python -m compass`.  I think it's worth the lag.  In "user" mode, we shouldn't have as much lag.

However, another source of this lag could be the addition of more and more test cases.  If we're concerned, we need to do some timing to find out if the creation time spent in the constructor of each test case is getting too costly.  If so, we would need to figure out how to optimize better.

> The MALI build worked smoothly following this workflow when building without Albany. Of course building with Albany requires very different modules/libs, so it will presumably take a lot of work to figure that out. I don't think this PR should in any way be held up by that, but that will delay our full adoption of developing in compass. We've talked about this offline, so I'll leave it at that here.

Yes, I think that's definitely a follow-on task and not something that can be part of this PR.

> Related to the previous item, in working through the steps, I realized I had some confusion about what the purpose of the dev environment. I had initially assumed it was to allow compass development, but as I understand it now, it is to provide a unified compass+mpas environment. Is that right? I have mixed feelings about that. I think it would be more useful for us (at least in the near term) to have a minimal compass dev env that only has the modules/libs to run compass itself from a conda env and is unrelated to modules/libs needed by MPAS. That would (in theory) allow landice folks to continue using our existing MALI build env for the foreseeable future until we tackle bringing Albany into the fold. Do you think this is possible? Could even be an undocumented thing for now.

Oh, wow, yes, I'm sorry that wasn't clear from the get-go here.  I feel strongly that this environment needs to be for compass and building/running MPAS components.  This is what enables us to use a version of ESMF built using system compilers, which is critical for creating mapping files as part of our workflows (something we need for several purposes such as interpolating topography and observed melt rates).  To me, your proposed idea of keeping compass separate from loading of system libraries sounds really regressive and prone to lots of user mistakes.  But it could be possible to create a simplified process that only builds the conda environment without loading any modules or setting environment variables.  I'm disappointed that that seems the preferable way forward for MALI but I understand that it would be an option.

> Setting aside Albany, I was able to set up and launch the SIA integration test suite using the instructions above. However, all the runs died with the pio stride error (which is presumably unrelated to this branch):
ERROR: PIO io_task_stride has a value of 4.
ERROR:         It must be between 1 and 1.
ERROR: PIO io_task_stride * io_task_count has a value of 4.
ERROR:         It must be between 1 and 1.
CRITICAL ERROR: Invalid PIO configuration.
I don't remember running into that before when I tested earlier branches, but maybe I've forgotten (or maybe I was on a different machine). In any case, can you advise me on how to correct this?

At least immediately, I don't see any reason this PR would have caused this issue and I have never run into this previously.  It seems like maybe you started a job with only one core?  But even then, the stride should have automatically been set to 1.  I would be happy to help debug this separately but would need more details so I can reproduce the issue.  Let's move that discussion somewhere else.

xylar:
@matthewhoffman, I added and documented a new ``--env_only`` flag for the script.  Please let me know if that meets the needs you described above.

matthewhoffman:
@xylar , thanks for the detailed response.  Regarding the long set up time and the slower command line response times, I agree those are both minor issues and seem acceptable costs for being able to work out of a conda env (if in fact that is even the cause), so we can dismiss those items.  Regarding the PIO thing, I'll try to reproduce that in master and get that sorted independently of this PR.  

I think that really only leaves the combined compass/mpas env issue I brought up.  Things make more sense now that the goal is a combined env.  Coming from our MALI environment workflow that has always been kind of jury-rigged and hardcoded, I wasn't even imagining being able to have such a thing.  Regarding your comment " I'm disappointed that that seems the preferable way forward for MALI", I should clarify that I do not see separate environments as the way *forward*, but it is definitely the way *now*.   Being able to eventually use a combined env that includes Albany is very attractive, but I think that will realistically take months to have worked out (because Trilinos is such a pain), and I don't want that to hold up our ability to use the python version of compass in the meantime.  (I'd like Mike K. to be able to start running it within days and the rest of the MALI team start porting the higher order Albany test cases in 'development mode'.)

So I was simply imagining an alternate dev env that has the minimal packages to support compass itself that we can use now.  We currently don't use ESMF regridding, so our near term solution could skip that complicating factor.  I wouldn't even expect this "dev-lite" env to necessarily be in the repo or docs or supported long-term, but could be a one-liner command we stick in our [MALI wiki](https://github.com/MALI-Dev/E3SM/wiki/MALI-Development-Overview).  Does that make sense for a short-term solution and is that easy to define somehow?

In either case, I imagine completing my evaluation of this PR in the context of the SIA-only MALI model configuration and not letting these larger issues slow the adoption of the new workflow.

matthewhoffman:
@xylar , I messages crossed.  That was really fast of you to provide a solution for our current workflow.  I'll give it a try now.

xylar:
@matthewhoffman, thanks for the further clarification.  I appreciate more what the needs for MALI with Albany are and how this is a necessary way forward.  The compass environment certainly doesn't need to be entwined with the system libraries so I'm happy with this solution, particularly as a workaround until we can build Albany and Trilinos  in a similar way to SCORPIO and ESMF.  I agree with you that's probably at least a couple of months off.

matthewhoffman (approved):
@xylar , I tested end-to-end in two modes:
1. Using the combined compass/mpas env, building MALI with that, and running the SIA test suite without any adjustments.
2. Using the --env_only option, building MALI independently using our env that supports Albany, and then running the SIA test suite with a manual adjustment to LD_LIBRARY_PATH  to account for some weirdness with that build.  (Note that Albany is not actually run, but the tests are using an exe that is built with Albany.)

Both methods worked!  So this clearly gives us a path to stay in sync with the desired long-term workflow the ocean is using while having the option of using the 'slim' env for working on porting over the remaining MALI test cases that require Albany.

A few notes: 
* The PIO problem went away this time around.  I think it is because I switched my request for an interactive node from `salloc -N 1 -t 00:30:00 --qos=interactive -n 36` to `salloc -N 1 -t 00:30:00 --qos=interactive` (dropped the -n argument).
* Using your unified env led to much better runtime performance.  There has been something messed up with the IO stack we are using on badger that leads to output taking forever to write on each time step.  It's annoying but we've been living with it, because once the Albany velocity solver is running, it's not too terrible (but it makes the short tests in COMPASS take forever).  COMPASS just zipped by in about 1/10 the time.  More motivation to migrate our Albany build to your unified library stack. :) 

I'm approving this PR and would be happy to see it merged ASAP.  Any other issues that come up can be adjusted in follow-on PRs, but I think it's good to complete this fundamental shift in how COMPASS development happens before introducing other team members to COMPASS.

xylar:
> The PIO problem went away this time around. I think it is because I switched my request for an interactive node from `salloc -N 1 -t 00:30:00 --qos=interactive -n 36` to `salloc -N 1 -t 00:30:00 --qos=interactive` (dropped the `-n` argument).

Okay, I'll look into that.  It does sound like a bug, probably in how I am determining the number of available processors.

matthewhoffman:
@xylar , can you comment a bit more on how the dev env is used when working with multiple branches and multiple developers?  I know that has been covered a bit, but I could use some confirmation.  When someone runs `./conda/create_compass_env.py` to create an activation script, that activation script is tied to the clone of COMPASS in which it was generated, right?  So each developer would need to create their own env/activation script.  Similarly, if I am working with multiple clones (or multiple worktrees of one clone), I would need to run `./conda/create_compass_env.py` for each one.  However, if I merely switch branches within a clone where I had already run `./conda/create_compass_env.py`, do I need to do anything at all?

I had tried alternating between two clones with different activation scripts, but the conda env that gets activated had the same name, so it was confusing.  What do you think about making the conda env name include the hash of compass from git describe?  

On another note, @vanroekel , how much do you want to look this over yet?  It sounds like getting this merged soon would be helpful for Mark and Xylar (and the MALI team).  I will be so bold as to suggest if you haven't already started working with compass development, I don't see a reason to even learn the old way of doing things and instead just see this approach as the starting point.  I imagine refinements or adjustments are possible but it's hard for me to see not following this general approach (working from a 'live' conda env that sees real-time updates to a compass repo).

xylar:
> @xylar , can you comment a bit more on how the dev env is used when working with multiple branches and multiple developers? I know that has been covered a bit, but I could use some confirmation. When someone runs ./conda/create_compass_env.py to create an activation script, that activation script is tied to the clone of COMPASS in which it was generated, right?  So each developer would need to create their own env/activation script. 

Yes, that's right.

> Similarly, if I am working with multiple clones (or multiple worktrees of one clone), I would need to run ./conda/create_compass_env.py for each one. 

Yes, you can to that or you can run the lighter weight `python -m pip install -e .` (one of the steps) as long as the dependencies haven't changed.

> However, if I merely switch branches within a clone where I had already run ./conda/create_compass_env.py, do I need to do anything at all?

The answer is that I don't think so.  (* fingers crossed *)  But that isn't my workflow so I've never tried it.

> I had tried alternating between two clones with different activation scripts, but the conda env that gets activated had the same name, so it was confusing. What do you think about making the conda env name include the hash of compass from git describe?

The conda environment is purposefully the same one, so it should have the same name.  Each activation script is reusing the same environment on purpose.  Different activation scripts are for loading different modules for different compilers.  The name of the conda environment should not have any extra details beyond what it is for (testing compass 1.0.0) because you don't want to be creating a unique conda environment per branch you work on, you just want to switch between which `compass` package is "installed" in that environment by calling the `conda/create_compass_env.sh` script to update the environment to point to the current directory.

If you want to have multiple conda environments for your own purposes, use the `--env_name` flag to name it whatever you want.  But again, I wouldn't create a new conda environment for each branch, that's just too much.

matthewhoffman:
So I should not see it as an activation script is tied to an environment.  The compass dev env is generic, and it is created once the first time an activation script is run by a given user.  Then the activation script from any given clone basically hijacks this one compass dev env.  Is that right?

I'm still confused by the fact that the activation script does not already exist, so in any new clone it would be necessary to run `create_compass_env.py` to create the activation script.  But that implies it would also be re-creating the env as well.

xylar:
> So I should not see it as an activation script is tied to an environment. The compass dev env is generic, and it is created once the first time an activation script is run by a given user. 

The dev env is created the first time `conda/create_compass_env.sh` is run by a given user, and is updated by subsequent calls unless `--recreate` is specified.  I can definitely understand why the name would not make that clear.  This was not the original behavior but it became clear in testing that recreating the environment should be the exception, rather than the default.  I think calling the script `update_compass_env.sh` would cause confusion in the other direction because it will create it if it doesn't exist.  I'm open to another name if a clearer one comes to mind.

BTW, an "activation script" in my mind is one of the scripts you source (like `test_compass_1.0.0*.sh`), which doesn't create anything.  So I was confused by your comment at first.

> Then the activation script from any given clone basically hijacks this one compass dev env. Is that right?

I guess you could see it that way.  Any activation script will activate the same environment, and there is no guarantee that the `compass` installed in that environment is the one from the directory where the activation script was called.

I hadn't thought about a way around that before but maybe we could include a call to `python -m pip install -e .` as part of the activation script if `setup.py` is found in the current directory.  That way, activation would also ensure that the environment points to the expected `compass`.

> I'm still confused by the fact that the activation script does not already exist, so in any new clone it would be necessary to run create_compass_env.py to create the activation script. But that implies it would also be re-creating the env as well.

Creating or updating.  Again a problem with the name of the script?

matthewhoffman:
Ok that helps.  It helps understanding that `conda/create_compass_env.sh` will not create a compass env if the correct one already exists.  As you suggest, the name implied to me it would always be creating a new env.  It could be renamed 'configure_compass_env.sh' or something.  But I'm also fine with it staying how it is and using the docs to explain the details.

xylar:
I'm good with changing the name to `configure_compass_env.py`.

mark-petersen:
It would be helpful to have this merged so I can use this method when I train students with MPAS next Tuesday. 

@vanroekel would you mind if we proceed with the merge if you are busy, or would you like to look it over? Thanks.

xylar:
> I hadn't thought about a way around that before but maybe we could include a call to python -m pip install -e . as part of the activation script if setup.py is found in the current directory. That way, activation would also ensure that the environment points to the expected compass.

I'm trying this out now.  Fingers crossed.

conda/configure_compass_env.py
```diff
@@ -0,0 +1,946 @@
...
+    with open(filename, 'r') as f:
+        template = Template(f.read())
+
+    if is_test:
+        update_compass = \
+            'if [[ -f "./setup.py" && -d "compass" ]]; then\n' \
+            '   # safe to assume we\'re in the compass repo\n' \
+            '   # update the compass installation to point here\n' \
+            '   python -m pip install -e .\n' \
+            'fi'
```
xylar, line 576: Okay, this worked well in a test on my laptop.  I think it's pretty harmless and will hopefully help to prevent confusion when you source the activation script in a new repo.  It's also fast enough to not slow down anyone's workflow.

xylar:
@mark-petersen and @matthewhoffman, apologies about the slightly moving target.  I renamed the script as @matthewhoffman suggested.  I also added a bit of code to the activation script (see my comment above) that will install `compass` into the conda environment as part of sourcing the activation script if you are within the compass repo.  (You could source the activation script from a different worktree, but it would still install the compass in the CWD -- I've been using this trick to take advantage of this branch before it gets merged to help with my other compass work.)

Hopefully, this helps, rather than creating new confusion.

matthewhoffman:
@xylar , thanks for making those changes.  I think they reduce confusion, but then I'm the one who initiated them, so hopefully the others agree.

vanroekel:
I’m okay with proceeding to merge without a detailed re-review from me.  I won’t have time anyway.  I will approve now.

vanroekel (approved):

xylar:
Thanks @vanroekel, @matthewhoffman and @mark-petersen!

PR #118, opened 2021-05-15
Reorganize ocean vertical module

xylar:
This merge adds general support for z-star and z-level vertical coordinates, similar to MPAS-Ocean init mode.  Starting with only `bottomDepth` and `ssh`, first a 1D vertical grid is computed based on config options:
* `refTopDepth`
* `refZMid`
* `refBottomDepth`
* `refInterfaces`

Then, the rest of the vertical grid is computed, including:
* `minLevelCell`, `maxLevelCell`
* `layerThickness`
* `restingThickness`
* `zMid`

In most cases, `bottomDepth` and `ssh` will be adjusted because of full or partial cells.

The functionality that was previously in the `_init_.py` file for the `vertical` module has been moved to `grid_1d.py` because it is not the primary functionality.

The 4 ocean test groups have been updated based on these changes.

Updates to the `iceshelf` module are also included because they relate to supporting the `minLevelCell` index.


xylar:
This work will make it easier to design general vertical coordinates and test them on all idealized test cases.

xylar:
## Testing

The `nightly` test suite passes with these changes.  the `ice_shelf_2d` restart test case is not bit-for-bit because there are changes in the order of operations for computing `zMid`, resulting in changes in `salinity`.  The changes are on the order of machine round-off so I think they're fine.  All other tests pass and are bit-for-bit (except for the usual failure in the RK4 restart test).

xylar:
@mark-petersen, I'm just pinging you on this work.  I'm not going to ask for a review but wanted you to be aware that I'm doing this reorganization in anticipation of work that @cbegeman will be doing on vertical coordinates.

mark-petersen:
OK, thanks for letting me know.

xylar:
Rebased and retested.  I did not do a bit-for-bit comparison for `ice_shelf_2d` because it was already found to have (expected) non-bit-for-bit changes (see https://github.com/MPAS-Dev/compass/pull/118#issuecomment-841794455) but all other tests are bit-for-bit on Linux with Gnu.

PR #119, opened 2021-05-16
Add color to validation

xylar:
This has helped a lot with suites, and would make it much easier to tell success from failure when validating individual test cases.


xylar:
## Testing

I ran the `ice_shelf_2d` restart test case with this branch, compared against a recent baseline, and got the expected green "PASS" text for all variables.

xylar:
@mark-petersen and @matthewhoffman, just a heads up about this.  I'm just going to add it, since it's a minor (aesthetic) change.

mark-petersen:
Great, thanks!

PR #120, opened 2021-05-16
Add z-level ice-shelf 2d test cases

xylar:
The z-star versions of the test cases are now in a z-star subdir.

The z-level restart test has been added to the nightly suite.

The cavity thickness has been increased from 10 to 50 m to accommodate the z-level coordinate.

### DIFF
--- a/compass/ocean/suites/nightly.txt
+++ b/compass/ocean/suites/nightly.txt
@@ -22,7 +22,8 @@ ocean/global_ocean/QU240/EN4_1900/performance_test
 ocean/global_ocean/QU240/PHC_BGC/init
 ocean/global_ocean/QU240/PHC_BGC/performance_test
 
-ocean/ice_shelf_2d/5km/restart_test
+ocean/ice_shelf_2d/5km/z-star/restart_test
+ocean/ice_shelf_2d/5km/z-level/restart_test
 
 ocean/ziso/20km/default
 ocean/ziso/20km/with_frazil
--- a/compass/ocean/tests/ice_shelf_2d/__init__.py
+++ b/compass/ocean/tests/ice_shelf_2d/__init__.py
@@ -15,23 +15,26 @@ class IceShelf2d(TestGroup):
         super().__init__(mpas_core=mpas_core, name='ice_shelf_2d')
 
         for resolution in ['5km']:
-            self.add_test_case(
-                Default(test_group=self, resolution=resolution))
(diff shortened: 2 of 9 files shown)


xylar:
## Testing

I ran the `nightly` test suite without comparing with a baseline.  The changes to the ice-shelf 2d test group (both the ice draft and the directory structure) are such that they are expected to fail the comparison anyway.  Since no other tests were modified by the changes specific to this PR, I didn't feel like bit-for-bit checks were needed.

xylar:
This will need to be rebased once #118 is merged.

xylar:
@cbegeman, I think it would be sufficient for your review if you 1) make sure you can rerun all 4 `ice_shelf_2d` test cases (a chance to get to know the new compass a bit) and 2) take a look at the namelist files for the `z-level` setups to make sure they are consistent with testing you've done with this test case.

xylar:
I ran with the `E3SM-Project` submodule but I also brought over the bug fixes from https://github.com/cbegeman/MPAS-Model/tree/ocean/minlevelcell_feature_bugfix in my branch https://github.com/xylar/E3SM/tree/ocn/bug-fixes-min-level-cell following the procedure described in https://github.com/E3SM-Project/E3SM/pull/4214#issuecomment-832007454

The bug fixes didn't seem t make a difference for these particular test cases (I didn't do a bit-for-bit check, thought), so I don't think it is necessary to wait for them to go into E3SM/master before merging this PR.

cbegeman:
@xylar I successfully ran these test cases without my bugfixes. The namelist files look good. I didn't do any viz though. Did you bring over anything to serve the same function as https://github.com/MPAS-Dev/compass/blob/legacy/ocean/sub_ice_shelf_2D/plot_sub_ice_shelf_2D.py? If not, no problem. I'm fine with you merging this as-is. Thanks!

cbegeman (approved):

xylar:
@cbegeman, bringing over that plotting script is a very good idea.  I'll add that.

xylar:
I added a viz step to make those plots.  It's not like the old way where there's just a script you can hack up in the local directory.  You have to hack it up in the compass package.  But otherwise, it works similarly.

vert_grid
vert_grid (1)

cbegeman:
Thanks, @xylar!

xylar:
@cbegeman, thanks for the review!

PR #122, opened 2021-05-26
Note that user's guide instructions are only valid after release

xylar:
...and we haven't released the package yet.

Add missing WC14 image.


xylar:
@vanroekel, no need to review this but your comments are welcome if you do want to take the time.

PR #123, opened 2021-05-26
Add E3SM-Project and MALI-Dev submodules

xylar:
Remove `MPAS-Model` submodules.

Change references to `MPAS-Model` repo --> `E3SM-Project` and `MALI-Dev` both the code and the documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,6 @@
 command_history
 local.config
 
-# exclude any new worktrees added
-MPAS-Model/*
-
-# except the ones we have on purpose
-!MPAS-Model/ocean/develop
-!MPAS-Model/landice/develop
-
 # Ignore all netcdf files and graph files
 *.nc
 *.info
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,0 @@
-[submodule "MPAS-Model/ocean/develop"]
-	path = MPAS-Model/ocean/develop
-	url = git@github.com:MPAS-Dev/MPAS-Model.git
-[submodule "MPAS-Model/landice/develop"]
-	path = MPAS-Model/landice/develop
-	url = git@github.com:MPAS-Dev/MPAS-Model.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "E3SM-Project"]
+	path = E3SM-Project
+	url = git@github.com:E3SM-Project/E3SM.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "E3SM-Project"]
 	path = E3SM-Project
 	url = git@github.com:E3SM-Project/E3SM.git
+[submodule "MALI-Dev"]
+	path = MALI-Dev
+	url = git@github.com:MALI-Dev/E3SM.git
--- a/compass/landice/landice.cfg
+++ b/compass/landice/landice.cfg
@@ -5,7 +5,7 @@
 
 # the relative or absolute path to the root of a branch where MALI has been
 # built
-mpas_model = MPAS-Model/landice/develop
+mpas_model = MALI-Dev/components/mpas-albany-landice
 
 # The namelists section defines paths to example_compact namelists that will be used
 # to generate specific namelists. By default, these point to the forward and
--- a/compass/ocean/ocean.cfg
+++ b/compass/ocean/ocean.cfg
@@ -5,7 +5,7 @@
 
 # the relative or absolute path to the root of a branch where MPAS-Ocean
 # has been built
-mpas_model = MPAS-Model/ocean/develop
+mpas_model = E3SM-Project/components/mpas-ocean
 
 # The namelists section defines paths to example_compact namelists that will be used
 # to generate specific namelists. By default, these point to the forward and
--- a/docs/developers_guide/command_line.rst
+++ b/docs/developers_guide/command_line.rst
@@ -87,8 +87,14 @@ the configuration and the names of its steps:
 compass setup
 -------------
 
-The ``compass setup`` command is used to set up one or more test cases. The
-command-line options are:
+The ``compass setup`` command is used to set up one or more test cases.
+
+.. note::
+
+    You must have built the executable for the standalone MPAS component you
+    want to run before setting up a compass test case.
+
+The command-line options are:
 
 .. code-block:: none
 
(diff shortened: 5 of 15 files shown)


xylar:
@mark-petersen and @matthewhoffman, could you test and review this a your convenience?

matthewhoffman:
@xylar , this looks good and I haven't encountered any problems in testing.  However, one issue I quickly realized is that our day-to-day testing will not be using the `E3SM-Project/E3SM` repo but instead in the `MALI-Dev/E3SM` repo.  I'm on the fence about suggesting we add a 2nd submodule for MALI or if the MALI team just manually adds the appropriate remote and branch before running any tests.  That touches a larger question of how frequently the ocean team anticipates updating the submodule.  Would that always be master?  Would a shared submodule hash ever be what we want in this situation, even if MALI was developing out of `E3SM-Project/E3SM`?

xylar:
I think it's quite important to have some way to indicate which E3SM git hash(es) a given compass is compatible with.  So it does seem like we need a `MALI-Dev/E3SM` submodule, suggesting that we need to use `E3SM-Project/E3SM` for the other, rather than just `E3SM`.

matthewhoffman:
That makes sense.  What about calling them `MPAS-Ocean` and `MALI` instead of `E3SM-Project/E3SM` and `MALI-Dev/E3SM`?  Or `E3SM-MPASO` and `E3SM-MALI`.

xylar:
If the submodules don't have a clear connection to the org/repo, I would go for `MPAS-Ocean` and `MALI` without the `E3SM`, which doesn't add clarity in my view.  @mark-petersen, what do you think?

xylar:
@matthewhoffman, I'm trying out `MPAS-Ocean` and `MALI` submodules and a couple of points for further discussion are emerging.
* It isn't clear to me that it's a good idea to have a submodule `MPAS-Ocean` that points to `E3SM-Project/E3SM`.  The long-term goal is for this submodule to point to the version of E3SM that is compatible with all compass cores (ocean, landice and eventually omega).
* But if this submodule is named `E3SM`, it isn't very clear that `landice` developers shouldn't be using that submodule and should use `MALI` instead, particularly since there's a `components/mpas-albany-landice` in there.
* It's weird documenting the path `MPAS-Ocean/components/mpas-ocean` and `MALI/components/mpas-albany-landice` in a bunch of places, but maybe that's unavoidable.

Anyway, I wanted to raise these questions/concerns now before I get too far with modifying the documentation (again).

matthewhoffman:
@xylar , I agree the situation is awkward.  The confusing meaning of the submodules is what motivated me to suggest `E3SM-MPASO` and `E3SM-MALI`, as in 'it is E3SM and for MPASO'.  (Could also do `E3SM-for-MPASO` I suppose.)  While the long term goal is a single submodule, I don't MALI being included for a year or two.  One other option would be have the ocean use `E3SM-Project/E3SM` (or just plain `E3SM` and MALI use `MALI-Dev/E3SM`.

xylar:
I think at this point I would, indeed, lean toward `E3SM-Project/E3SM` and `MALI-Dev/E3SM`.  That way, nothing is getting hidden under the rug and there's no implication about which component "owns" `E3SM-Project/E3SM`.  It would also be straightforward to add `E3SM-Ocean-Discussion/E3SM` as another submodule if that seems useful (I started to do that but realized it's potentially counterproductive and confusing because we *don't* plan to keep it in sync with compass and we don't want users treating it as the default place to run from or anything).

@mark-petersen, any thoughts?

matthewhoffman:
@xylar , based on this discussion, I agree that is the best solution, despite the slightly longer name.  If we want to avoid the extra empty directory level (so tab completion works), we could always replace the / with a . or _ though that may be confusing.  In any case, I'm good with this solution if @mark-petersen  is.  Let me know if you want me to contribute by making the associated revision since you did everything else.

mark-petersen:
I agree that we should have the exact organization name to avoid confusion. So the current proposal is that we could eventually have:
```
├── E3SM-Ocean-Discussion
│   └── E3SM
├── E3SM-Omega
│   └── E3SM
├── E3SM-Project
│   └── E3SM
├── E3SM-Seaice-Discussion
│   └── E3SM
```
I think that is a lot of extra directories at the highest level. My preference is to switch the order to `repo/organization`:
```
├── E3SM
│   ├── E3SM-Ocean-Discussion
│   ├── E3SM-Omega
│   ├── E3SM-Project
│   ├── E3SM-Seaice-Discussion
│   └── MALI-Dev
```
That is more logical to me, and neater. At the top level it says 'here is the E3SM code' and one directory down it says 'choose your fork'.
```

**Comment by mark-petersen** [comment]
Would we ever have repositories other than E3SM as submodules in compass anyway? If not, just 
```
E3SM-Project
MALI-Dev
```
at the top level would be sufficient. There is not a reason to include the Discussion forks, because it would just be tracking E3SM-Project/E3SM:master. Everyone will be developing on branches on their personal forks, so those could be anywhere on disk, not necessarily here as a submodule. 

Also, the purpose of a submodule is to track which version of E3SM is compatible with compass. So no need to do that with discussion forks.

xylar:
@mark-petersen, thanks.  I agree, let's eliminate the `E3SM` directory and just have the organization names.  I think that's clear enough.  I also agree that we don't need the discussion organizations for the reasons you said.

@matthewhoffman, are you good with that plan?  If so, I'll start making those changes.

matthewhoffman:
Agreed on all accounts.

xylar:
@matthewhoffman and @mark-petersen, this has been modified based on our discussion above and is ready for you to take another look.

xylar:
There is, of course, very little chance that I didn't miss anything in the documentation.  @matthewhoffman, you said you would go through at some point and make a PR with a lot of suggested changes.  I don't think we're quite at the point where it's time for that.  Let's get #117 in first.  But I'm hoping this change is one you can have a eye out for in that process.

xylar:
## Testing

I built MPAS-Ocean and MALI in the new submodules on my Linux laptop with gnu.  I successfully set up the `nightly` and `sia_integration` suites from the default locations (from each component's config file).  The suites ran as normal (known errors in the `enthalpy_benchmark/A` and RK4 restart tests).

I didn't check against a benchmark because I'm not 100% sure the code is the same as before the move to these modules and any diffs in the code itself aren't within the scope of this PR.  But I can check if anyone is concerned.

matthewhoffman (approved):
Looks good, @xylar .  The documentation are impressively thorough.

mark-petersen (approved):
Looks good. I did a quick retrieval on the submodules, compiled MPAS-O and ran. It all worked fine.

xylar:
Thank @mark-petersen and @matthewhoffman!

PR #124, opened 2021-05-26
Add E3SM submodule, remove MPAS-Model submodules

xylar:
Update legacy docs with E3SM move

PR #127, opened 2021-06-05
Fix pattern for xml search

xylar:
After the introduction of the E3SM submodules, `list_testcases.py` now finds some XML files in E3SM that it tries (and fails) to
parse.  One solution would be to have a try/except but the approach taken in this merge is to assume all COMPASS xml files start with `config`.  The XML files from E3SM that are causing trouble do not.

### DIFF
--- a/list_testcases.py
+++ b/list_testcases.py
@@ -72,43 +72,42 @@ if __name__ == "__main__":
     os.chdir(script_path)
 
     # Iterate over all cores
-    for core_dir in sorted(os.listdir('.')):
-        if os.path.isdir(core_dir) and core_dir != '.git':
-            # Iterate over all configurations within a core
-            for config_dir in sorted(os.listdir(core_dir)):
-                config_path = '{}/{}'.format(core_dir, config_dir)
-                if os.path.isdir(config_path):
-                    # Iterate over all resolutions within a configuration
-                    for res_dir in sorted(os.listdir(config_path)):
-                        res_path = '{}/{}'.format(config_path, res_dir)
-                        if os.path.isdir(res_path):
-                            # Iterate over all tests within a resolution
-                            for test_dir in sorted(os.listdir(res_path)):
-                                test_path = '{}/{}'.format(res_path, test_dir)
-                                if os.path.isdir(test_path):
-                                    do_print = False
-                                    # Iterate over all files within a test
-                                    for case_file in sorted(
-                                            os.listdir(test_path)):
-                                        if fnmatch.fnmatch(case_file, '*.xml'):
-                                            tree = ET.parse('{}/{}'.format(
-                                                test_path, case_file))
-                                            root = tree.getroot()
+    for core_dir in ['landice', 'ocean', 'test']:
+        # Iterate over all configurations within a core
+        for config_dir in sorted(os.listdir(core_dir)):
+            config_path = '{}/{}'.format(core_dir, config_dir)
+            if os.path.isdir(config_path):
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen and @matthewhoffman, apparently I didn't do sufficient testing of #124.  I was using it yesterday and getting the following error:
```
Available test cases are:
  File "./list_testcases.py", line 94, in <module>
    tree = ET.parse('{}/{}'.format(
  File "/home/xylar/miniconda3/lib/python3.8/xml/etree/ElementTree.py", line 1202, in parse
    tree.parse(source, parser)
  File "/home/xylar/miniconda3/lib/python3.8/xml/etree/ElementTree.py", line 595, in parse
    self._root = parser._parse_whole(source)
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 17, column 7
```
I added a print statement right above the place where the error happens and the offending file is `cice_decomp.xml` (`E3SM-Project/components/cice/bld/cice_decomp.xml` or the same path in `MALI-Dev`). This file seems to have some comment lines of all dashes as separators that are invalid XML.  Anyway, the point is that they can't be parsed.

An alternative to the solution here would be:
``` python
try:
    tree = ET.parse('{}/{}'.format(test_path, case_file))
except ET.ParseError:
    continue
```
I have tried this as well and it works.  But the result would be that bad XML files in COMPASS itself would simply be skipped silently.  I think that's risky.

I believe Doug's original design was that COMPASS XML files didn't necessarily have to be named `config*.xml`, which is why the list script looks at all XML files.  However, I think in practice we have always named our XML files starting with `config`.  This is corroborated by the fact that both the proposed fix here (searching with the `config` prefix) and the alternative with a `try/except` list 208 test cases.

matthewhoffman:
@xylar , rather than restrict the allowable xml file names, what about simply ignoring the files in the submodules?

xylar:
@matthewhoffman, in that case it would be safer to list the cores (`landice`, `ocean` and `test`) that it's allowed to look in, rather than the submodules it's not allowed to use because I've run into the same issue when I create new worktrees for other E3SM branches.  But that could be done.

mark-petersen:
I'm prepping for all the students today. I'm happy to trust your judgement here.

xylar:
@matthewhoffman, could you please take a look at this alternative and see if it fits your needs?

list_testcases.py
```diff
@@ -72,43 +72,42 @@ def print_case(quiet, args, core_dir, config_dir, res_dir, test_dir, case_num):
...
-                                                do_print = True
-
-                                            del root
-                                            del tree
-
-                                    if do_print:
-                                        case_num = print_case(
-                                            quiet, args, core_dir, config_dir,
-                                            res_dir, test_dir, case_num)
+    for core_dir in ['landice', 'ocean', 'test']:
```
xylar, line 75: This is the only line that actually changed.  The rest is just indentation.

list_testcases.py
```diff
@@ -72,43 +72,42 @@ def print_case(quiet, args, core_dir, config_dir, res_dir, test_dir, case_num):
     os.chdir(script_path)
 
     # Iterate over all cores
-    for core_dir in sorted(os.listdir('.')):
-        if os.path.isdir(core_dir) and core_dir != '.git':
```
xylar, line 76: These 2 lines were removed.  Again, the rest is just indentation.

matthewhoffman (approved):
@xylar, I like that solution better.   That seems nice and clean.

PR #128, opened 2021-06-05
Add a tutorial on porting test groups from legacy

xylar:
This tutorial uses the `gotm` test group and the default test case as an example.

PR #129, opened 2021-06-05
Add the `gotm` test group and `default` test case

xylar:
These are ported from the `ocean/gotm/2.5km/default` test case in legacy COMPASS.

This merge also adds documentation of the new test group.


xylar:
This test group is used as the example in the tutorial in #128.

xylar:
## Testing

Using my Ubuntu laptop and Gnu, I ran the legacy test case and this test case, comparing the results of ``init`` and ``analysis``.  (I didn't have an efficient way of comparing the results of ``forward`` but could do so if desired.)  The analysis plots looked identical.

xylar:
@qingli411, it would be helpful to have this merged soon-ish so it is available to new developers as an example.  But I also recognize that you have not used the new `compass` yet and you may need to read the documentation and get up to speed before you're ready to review this.  Please let me know if you're able to review and, if so, on what time scale.  Thanks very much in advance!

qingli411:
@xylar, thanks for this and the tutorial in https://github.com/MPAS-Dev/compass/pull/128! I'm certainly interested in learning the new compass and testing this. But I'm currently traveling out of town and won't be able to test it until 6/11 when I come back. Does that work?

xylar:
@qingli411, for this PR, I'm more than happy to wait.  For #128, we may go ahead just so it's in for new developers to work from.  But your feedback after-the-fact would be very much appreciated.  It's easy to put in another PR later with updates and fixes.

qingli411:
@xylar, this sounds good. Please go ahead with https://github.com/MPAS-Dev/compass/pull/128. I will test this PR when I get back and let you know.

xylar:
This needs to be rebased after #138 goes in.

qingli411 (approved):
@xylar This looks good to me. I also tested the gotm case in this PR on Cori with GNU and it works great.

xylar:
@qingli411, thanks so much for reviewing.

PR #131, opened 2021-06-06
Simplify and generalize the `download()` function

xylar:
We don't use the ability to use config options for downloads directly with this function, so this merge eliminates the unused feature.

Cases have arisen where it is useful to have a URL that has a different file name than the target filename that we want to download to.  Sometimes, the URL doesn't have a particularly unique or descriptive name.

This merge also updates the `Step` class to handle the changes in `download()`, making sure URLs include the file name.

### DIFF
--- a/compass/io.py
+++ b/compass/io.py
@@ -2,80 +2,64 @@ import os
 import tempfile
 import requests
 import progressbar
+from urllib.parse import urlparse
 
-from compass.config import get_source_file
 
-
-def download(file_name, url, config, dest_path=None, dest_option=None,
-             exceptions=True):
+def download(url, dest_path, config, exceptions=True):
     """
     Download a file from a URL to the given path or path name
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I used a config file that pointed to a new location for downloads on my Linux laptop.  Then, I ran the ocean nightly suite.  The files got downloaded again as expected. The suite ran as usual (all tests succeeded except RK4 restart).

xylar:
@mark-petersen and @matthewhoffman, just FYI, I need this functionality because the Adusumilli melt rates are at this not-very-useful URL: http://library.ucsd.edu/dc/object/bb0448974g/_3_1.h5  I'd like to save that file to something more descriptive (Sushil himself uses `ANT_iceshelf_melt_rates_CS2_2010-2018_v0.h5` in his example scripts.  So I need the ability to save a URL to a new file name, which is something we hadn't needed so far.

PR #132, opened 2021-06-08
Make 'compass list --suites -v' print the contents of a suite

matthewhoffman:
This merge also updates the documentation for this change and adds a
short section to the Developer's Guide about developing suites.

### DIFF
--- a/compass/list.py
+++ b/compass/list.py
@@ -100,12 +100,15 @@ def list_suites(cores=None, verbose=False):
             if suite.endswith('.txt'):
                 print('  -c {} -t {}'.format(core, os.path.splitext(suite)[0]))
                 if verbose:
-                   text = resources.read_text('compass.{}.suites'.format(core), suite)
-                   tests = list()
-                   for test in text.split('\n'):
-                      test = test.strip()
-                      if len(test) > 0 and test not in tests and not test.startswith('#'):
-                         print("\t* {}".format(test))
+                    text = resources.read_text(
+                        'compass.{}.suites'.format(core), suite)
+                    tests = list()
+                    for test in text.split('\n'):
+                        test = test.strip()
+                        if (len(test) > 0 and test not in tests
+                                and not test.startswith('#')):
+                            print("\t* {}".format(test))
+
 
 def main():
     parser = argparse.ArgumentParser(
--- a/compass/list.py
+++ b/compass/list.py
@@ -21,8 +21,8 @@ def list_cases(test_expr=None, number=None, verbose=False):
         The number of the test to list
 
     verbose : bool, optional
-        Whether to print details of each test or just the subdirectories
-        When applied to suites, verbose will list the tests in the suite
+        Whether to print details of each test or just the subdirectories.
+        When applied to suites, verbose will list the tests in the suite.
     """
     mpas_cores = get_mpas_cores()
 
@@ -126,7 +126,8 @@ def main():
                         help="List test suites (instead of test cases)")
     parser.add_argument("-v", "--verbose", dest="verbose", action="store_true",
                         help="List details of each test case, not just the "
-                             "path")
+                             "path.  When applied to suites, verbose lists "
+                             "the tests contained in each suite.")
     args = parser.parse_args(sys.argv[2:])
     if args.machines:
         list_machines()
--- a/compass/list.py
+++ b/compass/list.py
@@ -2,6 +2,7 @@ import argparse
 import re
 import sys
 import os
+from importlib import resources
 from importlib.resources import contents
 
 from compass.mpas_cores import get_mpas_cores
@@ -21,6 +22,7 @@ def list_cases(test_expr=None, number=None, verbose=False):
 
     verbose : bool, optional
         Whether to print details of each test or just the subdirectories
+        When applied to suites, verbose will list the tests in the suite
     """
     mpas_cores = get_mpas_cores()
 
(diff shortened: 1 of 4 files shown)


compass/list.py
```diff
@@ -21,6 +22,7 @@ def list_cases(test_expr=None, number=None, verbose=False):
 
     verbose : bool, optional
         Whether to print details of each test or just the subdirectories
+        When applied to suites, verbose will list the tests in the suite
```
xylar, line 25:
```suggestion
        when applied to suites, verbose will list the tests in the suite
```

xylar, line 25:
```suggestion
        Whether to print details of each test or just the subdirectories.
        When applied to suites, verbose will list the tests in the suite.
```
If you have multiple sentences, the API documentation may put them on the same line (the line breaks here aren't necessarily retained) so it's important to punctuate.

compass/list.py
```diff
@@ -120,7 +128,7 @@ def main():
     if args.machines:
         list_machines()
     elif args.suites:
-        list_suites()
+        list_suites(verbose=args.verbose)
```
xylar, line 131: Could you update the help string for `--verbose` to also mention suites?

docs/developers_guide/organization.rst
```diff
@@ -1718,3 +1718,16 @@ forgotten to compile the model, this will be obvious by the broken symlink and
...
+
+Test Suites
+-----------
+
+As described in the :ref:`user_testsuites` section of the User's Guide, COMPASS
+test cases can be organized into test suites.  Each core has separate regression
+suites, and a core can have multiple independent regression suites.  A developer
+defines a test suite by creating a `.txt` file within the `compass/CORE/suites`
+directory.  The format of the `.txt` file is a list of the paths to the tests
+desired to be part of the suite.  A line starting with `#` will be treated as a
```
xylar, line 1732:
```suggestion
directory.  The format of the `.txt` file is a list of the work directories of the tests
desired to be part of the suite.  A line starting with `#` will be treated as a
```

compass/list.py
```diff
@@ -97,7 +99,13 @@ def list_suites(cores=None):
         for suite in sorted(suites):
             if suite.endswith('.txt'):
                 print('  -c {} -t {}'.format(core, os.path.splitext(suite)[0]))
-
+                if verbose:
+                   text = resources.read_text('compass.{}.suites'.format(core), suite)
```
xylar, line 103:
```suggestion
                   text = resources.read_text('compass.{}.suites'.format(core), 
                                              suite)
```

compass/list.py
```diff
@@ -97,7 +99,13 @@ def list_suites(cores=None):
...
+                   tests = list()
+                   for test in text.split('\n'):
+                      test = test.strip()
+                      if len(test) > 0 and test not in tests and not test.startswith('#'):
+                         print("\t* {}".format(test))
```
xylar, line 108:
```suggestion
                      if len(test) > 0 and test not in tests and \
                              not test.startswith('#'):
                          print("\t* {}".format(test))
```

xylar (changes requested):
This is a really great addition and works like a charm.  I hadn't originally done this because the text files are pretty easy to read but I can definitely see the benefit because they're not that easy to find.  This will be really handy.

I've made a few small suggested edits.  In addition, it would be great if you could take care of the 8 linting issues found above by pep8speaks.  I would like us to stick with the PEP8 style guide pretty strictly.  The main exception are long lines that really can't be broken up without hampering readability, such as very long type names or URLs in doc strings.  Please make sure indentation is always a multiple of 4.  The two long lines pointed out here could be broken into multiple lines as I have suggested below.

xylar:
@matthewhoffman, I also really appreciate you taking the initiative to implement this, and not just request this feature.  It makes a really big difference to how the project feels to have others diving into development.

compass/list.py
xylar: One more thing. I'm not sure why pep8speaks didn't point it out but there need to be 2 blank lines between functions (at line 109).

matthewhoffman:
@xylar , thanks for reviewing this and providing me feedback on coding style and documentation.  I have flake8 installed now (including a vim plugin), and it really makes linting easy.  (flake8 did identify the missing 2 blank lines after a function, BTW.)  I was aware there were coding standards but I wasn't sure how to easily check my code against them so I kinda ignored it the first time around. :)  I've made all the changes you requested - let me know if you see anything else.

I realized this was a kind of trivial PR to make, functionality-wise, but I had a few reasons to go ahead with it.  The first was I was trying to figure out if I could comment my suite definition .txt files, and this seemed like a nice way to test that.  The second was that I realized that for a user using the released python package, they would not have access to the .txt files (or even for someone working from the repo, they might not be familiar with where to find them).  And then I decided this was a nice simple modification for me to practice making a compass PR!   It's nice to get my feet wet and feel like I can jump into the framework.  With the developer docs and the well-commented source code, it was really easy to figure out how to implement this.

matthewhoffman:
@xylar , hold of on re-reviewing.  I think some changes are required to support comments in the suite .txt files.  I had tested that yesterday, but I most of have missed something or had a problem with my conda env.

xylar:
> I think some changes are required to support comments in the suite .txt files.

Yes, I also just tested and all values are parsed, so you are right.  It should be pretty easy to add comment parsing.

matthewhoffman:
I added a new commit to support test case commenting by the `compass suite` command.  I had previously added it for `compass list` but forgot to also add it to `compass suite`.  

I tested by adding a new line to the landice integaration suite:
```
# foo
```
and I confirmed it was ignored by:
* `compass list --suites -v`
* `compass suite -c landice -t sia_integration -w $scratch4/compass_testing -s -m badger`
* `compass suite -c landice -t sia_integration -w $scratch4/compass_testing --clean -m badger`

The comment line was ignored by all those commands.  This is ready for final review and merge.

xylar (approved):
@matthewhoffman this looks good.  I think your testing is sufficient.  Feel free to merge when you're ready

PR #133, opened 2021-06-08
Add HO versions of dome tests

matthewhoffman:
This merge creates HO versions of the various dome tests.  It does so by creating a velo_solver argument to these tests and sets up tests for both 'sia' and 'FO'.  This merge also adds a full_integration test suite that includes both sia and FO tests.


compass/landice/tests/dome/run_model.py
```diff
@@ -72,6 +77,11 @@ def __init__(self, test_case, mesh_type, name='run_model', subdir=None,
                 'compass.landice.tests.dome', 'namelist.landice',
                 out_name='namelist.{}'.format(suffix))
 
+            replacements = {'config_velocity_solver': '{}'.format(velo_solver)}
+
+#            update(replacements=replacements, step_work_dir=self.path,
```
matthewhoffman, line 82: @xylar , when I uncomment this line, `compass list` fails with this error:
```
FileNotFoundError: [Errno 2] No such file or directory: 'landice/dome/sia/2000m/smoke_test/run_model/namelist.landice'
```
My guess is that the namelist `update` method doesn't work in this context because self.add_namelist_file hasn't actually done anything yet in `compass list`.  I may need to rethink how this is organized.  I don't want to wait to update the namelist until the `run` method, because then the namelist set up in the working directory would be wrong and likely confusing if anyone looked at it.  What do you suggest?  Is it actually necessary for `compass list` to run a step's `__init__` function?

compass/landice/tests/dome/decomposition_test/__init__.py
```diff
@@ -26,12 +26,16 @@ def __init__(self, test_group, mesh_type):
         test_group : compass.landice.tests.dome.Dome
             The test group that this test case belongs to
 
+        velo_solver : str
```
xylar, line 29:
```suggestion
        velo_solver : {'sia', 'FO'}
```
I'm assuming there will only ever be these two, in which case listing them is a bit better than giving the type as `str`.

compass/landice/tests/dome/restart_test/__init__.py
```diff
@@ -26,12 +26,16 @@ def __init__(self, test_group, mesh_type):
         test_group : compass.landice.tests.dome.Dome
             The test group that this test case belongs to
 
+        velo_solver : str
```
xylar, line 29:
```suggestion
        velo_solver : {'sia', 'FO'}
```

compass/landice/tests/dome/run_model.py
```diff
@@ -29,6 +30,9 @@ def __init__(self, test_case, mesh_type, name='run_model', subdir=None,
         test_case : compass.TestCase
             The test case this step belongs to
 
+        velo_solver : str
```
xylar, line 33:
```suggestion
        velo_solver : {'sia', 'FO'}
```

compass/landice/tests/dome/smoke_test/__init__.py
```diff
@@ -24,19 +24,23 @@ def __init__(self, test_group, mesh_type):
         test_group : compass.landice.tests.dome.Dome
             The test group that this test case belongs to
 
+        velo_solver : str
```
xylar, line 27:
```suggestion
        velo_solver : {'sia', 'FO'}
```

compass/landice/tests/dome/run_model.py
xylar: I can understand the confusion.  There is a method `self.add_namelist_options()` that is what you want here.  

https://mpas-dev.github.io/compass/latest/developers_guide/organization.html#adding-namelist-options

`compass.namelist.update()` only works in `run()` as you have discovered.  It would be really helpful if we could come up with a way to make these two more consistent.  One thing would be to move `update()` to be a method in `Step` as well, maybe named something like `update_namelist_at_runtime()`.  Verbose but hopefully clearer.

I'd appreciate your thoughts on what would clear this up for you.

compass/landice/tests/dome/run_model.py
```diff
@@ -72,6 +77,11 @@ def __init__(self, test_case, mesh_type, name='run_model', subdir=None,
                 'compass.landice.tests.dome', 'namelist.landice',
                 out_name='namelist.{}'.format(suffix))
 
+            replacements = {'config_velocity_solver': '{}'.format(velo_solver)}
+
+#            update(replacements=replacements, step_work_dir=self.path,
+#                                   out_name='namelist.{}'.format(suffix))
```
xylar, line 83:
```suggestion
            self.add_namelist_options(replacements=replacements, 
                                      out_name='namelist.{}'.format(suffix))
```

compass/landice/tests/dome/run_model.py
```diff
@@ -91,6 +91,9 @@ def __init__(self, test_case, velo_solver, mesh_type, name='run_model',
                             target='../setup_mesh/landice_grid.nc')
         self.add_input_file(filename='graph.info',
                             target='../setup_mesh/graph.info')
+        if velo_solver == 'FO':
+            self.add_input_file(filename='albany_input.yaml',
```
matthewhoffman, line 95: @xylar , when adding Albany's configuration file in this way, it shows up in the workdir run directory as a symlink.  My understanding from looking at the code is that using the `filename` argument should add it as an actual file.  Am I reading that wrong, or is this function not behaving properly?

It's not a big deal if this file is a symlink or file, but I think a standard file is preferable in this case.

matthewhoffman:
@xylar , I have finished revising this branch.  With your suggestions, I know have all 4 of the Albany-based dome tests running successfully.  As we discussed previously, right now they still require building MALI outside of the COMPASS/MPAS conda build env, but doing that and using the -p argument to compass setup/suite gives me successful tests.  I also threw in a couple minor tweaks to get some of the existing non-Albany tests that were failing to now pass.  At this point I get successful results for all of the 13 non-Albany tests and the 4 new Albany-based tests.  

I'm happy with the implementation and the results with this branch, but if/when you have time it would be great to get your eyes on this second round of changes to make sure you don't see anything questionable.  Also, I added an inline comment above about how to avoid a symlink in an input file.  If possible I'd like to fix that before merging this, but otherwise it is ready to go.

Also, we should follow up with how you want to handle reviews and merges for landice tests.  I imagine long-term you don't want to have to be responsible for integrating those PRs, but initially it's probably best to have you involved at some level.  I don't feel 100% comfortable integrating in this repo yet.

matthewhoffman:
cc: @mperego , @trhille , @stephenprice , this PR is an example of porting MALI test cases to the new COMPASS system.  I don't expect you to follow what's involved in this PR, but you might skim it to get a general idea of how this goes.  This PR adapts the SIA dome test case that Xylar already ported to also apply to the FO version of the dome tests.  Porting test cases that are not already in the new COMPASS system (like circular shelf) will require substantially more work.

matthewhoffman:
@xylar , I added a commit with updates to the docs (following your good example).  For some reason, when I re-build the docs, I'm not seeing my changes, so I'm not sure what is going on there.

compass/landice/tests/dome/run_model.py
xylar: It's functioning properly, `add_input_file` always makes a symlink unless you are downloading a file directly to the work directory from a URL.  There is currently no way to make a copy of an input file from the package but that would be an easy functionality to add.  My preference would be to add a new `copy=False` keyword argument to `add_input_file()` so you could pass `copy=True` if you want to copy rather than symlink.  I'll do this now, and you can review.

xylar: See https://github.com/MPAS-Dev/compass/pull/138

docs/users_guide/landice/test_groups/dome.rst
```diff
@@ -8,12 +8,24 @@ from `Halfar (1983) <https://doi.org/10.1029/JC088iC10p06043>`_ and
...
+an analytic solution when using the shallow-ice approximation for dynamics.
+See the `MALI User's Guide <https://zenodo.org/record/1227426#.YMEznB1MGAY>`, 
+section 12.1 for the equations solved.
 
-<<<Exmaple image here>>>
+.. image:: images/halfar.png
+   :width: 700 px
+   :align: center
+Halfar test case results after 200 years of dome evolution. This figure is 
+generated by the visualization script.
```
xylar, line 21:
```suggestion
.. image:: images/halfar.png
   :width: 700 px
   :align: center

Halfar test case results after 200 years of dome evolution. This figure is 
generated by the visualization script.
```
You always need a blank line after any indented text like here in rst.

xylar:
> @xylar , I added a commit with updates to the docs (following your good example). For some reason, when I re-build the docs, I'm not seeing my changes, so I'm not sure what is going on there.

Maybe try running `make clean; make html`?

xylar:
> Also, we should follow up with how you want to handle reviews and merges for landice tests. I imagine long-term you don't want to have to be responsible for integrating those PRs, but initially it's probably best to have you involved at some level. I don't feel 100% comfortable integrating in this repo yet.

I think it makes sense for me to review all PRs for any core for the foreseeable future, just for formatting and as a sanity check.  I probably won't have the time or experience necessary to actually run tests.  So probably the right combination is that you review for functionality and I review for consistency of the implementation with the design goals, code style, making sure docs are included, that sort of thing.

xylar:
@matthewhoffman, if you could fix the 7 remaining PEP8 issues (https://github.com/MPAS-Dev/compass/pull/133#issuecomment-857207803), that would be great!

I was able to build the docs after making the small correction above.

With this branch (including the code built from the update MALI-Dev submodule), the `sia_integration` suite ran perfectly for me on Linux with Gnu compilers.  Everything passed for me in under 2 minutes:
```
Total runtime 01:60
```
(I've been meaning to fix that rounding issue...)

compass/landice/tests/dome/run_model.py
matthewhoffman: I like the idea of renaming it `update_namelist_at_runtime()`.  And if the function can really only be applied to a step then moving into the `Step` class where it would exist alongside `add_namelist_options()` would be clean.

By the way, I did have a look in the docs to try to answer my question but for some reason didn't find the section that holds the answer.  I found this section instead and stopped looking: https://mpas-dev.github.io/compass/latest/developers_guide/framework.html#updating-a-namelist-file  I just noticed now the docs have a search function, so I should remember that in the future.

matthewhoffman:
BTW, @trhille , your PR https://github.com/MALI-Dev/E3SM/pull/1 fixed some fails in some of the existing tests.  One of the commits in this PR takes advantage of that by updating the version of MALI to be used with COMPASS, instantly resolving those fails. 🎉

matthewhoffman:
@xylar, I've made all the new changes that came up from yesterday's review and discussion.  I would like to hold off on merging this until #138 is merged if you think that will happen in the next few days (I'm not trying to rush you).  Then I'll rebase this branch to introduce a commit that copies the Albany yaml files instead of symlinking them.

compass/landice/tests/dome/run_model.py
xylar: I have a fix for this mostly implemented.  I don't think this PR would be affected by it, since you're now using `add_namelist_options()`, right?

xylar (approved):
Feel free to rebase, but I'm happy for you to merge when you're ready.

We're not picky about what you put into your merge commit.  You can use the green merge button.  You can copy the description from the PR but you don't have to.

compass/landice/tests/dome/run_model.py
matthewhoffman: That is correct.  I will proceed to merge without that change.

PR #134, opened 2021-06-09
Add flake8 to development environments

xylar:
This is helpful for folks that want to use it directly for linting or want to use the vim plugin.

### DIFF
--- a/conda/compass_env/spec-file-mpich.txt
+++ b/conda/compass_env/spec-file-mpich.txt
@@ -34,6 +34,7 @@ xarray
 c-compiler
 cmake
 cxx-compiler
+flake8
 fortran-compiler
 libpnetcdf=*=mpi_mpich*
 m4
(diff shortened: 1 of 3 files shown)


xylar:
@matthewhoffman, you can probably just approve by inspection.  I believe it *should* work to just rerun `./conda/configure_compass_env.sh` with the appropriate flags but I found I needed the `--recreate` flag in at least one case for it to notice new dependencies.  Let me know if that's the case for you.  It would be worth tracking down.

xylar:
I'm just going to merge this...

matthewhoffman:
@xylar , thanks for adding this so quickly.  I tried it out and it worked for me without the `--recreate` option.  It did require a bunch of new packages that seemed unrelated to flake8, but that might have been from other updates on master since I last ran the configure script.  Or it could just be the intricacies of conda package management, and I won't attempt to understand them!

xylar:
That's really good to hear.  Yes, I wouldn't worry about other updates.  Most likely, you are right.  These just happen to be packages that have had updates in recent days, nothing to do with `flake8`. And definitely not worth investigating.  Just glad to know that you don't have to do `--recreate` to get updates.

PR #137, opened 2021-06-10
Add make to dev environments

xylar:
I ran into errors on my new Linux laptop because `make` was not available.


xylar:
## Testing

I'm able to create a development environment with `./conda/configure_compass_env.sh` with this branch, whereas it failed with `master`.

PR #138, opened 2021-06-10
Add a copy option for input files

xylar:
If ``copy=True``, the file will be copied to the work directory rather than symlinked.

The documentation has been updated to include the new ``copy`` keyword argument.

### DIFF
--- a/compass/step.py
+++ b/compass/step.py
@@ -2,6 +2,7 @@ import os
 from lxml import etree
 import configparser
 from importlib.resources import path
+import shutil
 
 from compass.io import download, symlink
 import compass.namelist
@@ -201,7 +202,8 @@ class Step:
         pass
 
     def add_input_file(self, filename=None, target=None, database=None,
-                       url=None, work_dir_target=None, package=None):
+                       url=None, work_dir_target=None, package=None,
+                       copy=False):
         """
         Add an input file to the step (but not necessarily to the MPAS model).
         The file can be local, a symlink to a file that will be created in
@@ -242,6 +244,9 @@ class Step:
 
         package : str or package, optional
             A package within ``compass`` from which the file should be linked
+
+        copy : bool, optional
+            Whether to make a copy of the file, rather than a symlink
         """
         if filename is None:
             if target is None:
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I used the new `copy=True` option in the GOTM test group (#129) and it worked as expected.  I got a copy of the `gotmturb.nml` file as expected, and the test ran as normal.

matthewhoffman (approved):
@xylar , this looks great!  Thanks for implementing it so quickly.

When I tried it on my existing test suite, I got this error:
```
shutil.SameFileError: '/turquoise/usr/projects/climate/mhoffman/mpas/compass/compass/landice/tests/dome/albany_input.yaml' and '/lustre/scratch4/turquoise/mhoffman/compass_testing/landice/dome/2000m/fo_decomposition_test/1proc_run/albany_input.yaml' are the same file
```
but that seems like an edge case that we should not design around.  If a developer changes an existing symlink to the copy option, then any users would just have to delete existing instances of that test case before proceeding.

xylar:
@matthewhoffman, can you share here the code that caused your error above?  It seems like the target and destination are somehow the same, and that shouldn't happen.  But maybe there's a problem with my code.

matthewhoffman:
This only happened because I was trying to set up the test case after switching to the copy method after already having the test case setup using the symlink method.  I've since deleted the working dir, but I think that is expected behavior.  The cp command exhibits that behavior if you try to copy a file to a symlink (since it's effectively trying to copy through the symlink to the original file, which is also what we are copying from).  In my experience, the only way around this is to either use `unlink` or `rm` on the symlink before copying the file.  I don't think that is something we should try to catch programmatically.  If a developer decides to switch from the symlink to cp method, any affected users can just delete an existing working dir for that test.

xylar:
Ah, I see.  I agree, this is the kind of thing where you need to run `compass clean` before `compass setup` if you run into trouble.  I agree that it's not something I would try to fix in the code.

xylar:
Now that I understand the issue you ran into, I'm comfortable just merging.

PR #140, opened 2021-06-10
Add more info about using nonzero norms for validation to Dev Guide

matthewhoffman:
This merge provides some additional info to the Developer's Guide about using the validation method with nonzero norms.


xylar:
@matthewhoffman, this is great!  Would you be willing to edit the docstring in this PR, too, to add details about setting each norm to `None` if you don't want to compute it?

matthewhoffman:
Sure - I've pushed those changes.

PR #141, opened 2021-06-10
Allow `compass run suitename` to include or not the .pickle suffix

matthewhoffman:
Allowing the .pickle suffix in the `compass run` command is a convenience
change to support command-line tab completion.

### DIFF
--- a/compass/run.py
+++ b/compass/run.py
@@ -31,7 +31,7 @@ def run_suite(suite_name):
     # Allow a suite name to either include or not the .pickle suffix
     if suite_name.endswith('.pickle'):
         # code below assumes no suffix, so remove it
-        suite_name = suite_name.replace('.pickle', '')
+        suite_name = suite_name[:-len('.pickle')]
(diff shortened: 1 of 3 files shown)


xylar (approved):
Looks good.  I'll trust you that it actually works.  Looks like it will.

** Updaate **: felt like a jerk after not testing things properly.  I tested this out on the ocean nightly test suite and `compass run nightly.pickle` worked as expected.  (`compass run .picklenightly.pickle` also worked ;-)).

matthewhoffman:
@xylar , the edge case you identified bugged me, despite it seeming extremely unlikely to ever be relevant. :) But I changed the str parsing method to only remove .pickle from the end of a string.  I tested all the variants I could think of and it worked as expected.  I'll merge this once the automated tests complete.

xylar:
I agree that this would never come up but as a general practice, I think your latest version is better.  Thanks for looking into how to remove the suffix from the end.

PR #142, opened 2021-06-10
Move `namelist.update` and `model.update_namelist_pio` into `Step`

xylar:
These are more intuitive if they're methods of Step instead. This merge also updates the test cases and core framework that is affected.  It also updates the documentation with these changes.

### DIFF
--- a/compass/step.py
+++ b/compass/step.py
@@ -345,7 +345,8 @@ class Step:
         """
         Update an existing namelist file with additional options.  This would
         typically be used for namelist options that are only known at runtime,
-        not during setup, typically those related to the number of nodes and cores.
+        not during setup, typically those related to the number of nodes and
+        cores.
 
         Parameters
         ----------
--- a/compass/step.py
+++ b/compass/step.py
@@ -3,6 +3,7 @@ from lxml import etree
 import configparser
 from importlib.resources import path
 import shutil
+import numpy
 
 from compass.io import download, symlink
 import compass.namelist
@@ -321,7 +322,7 @@ class Step:
         ----------
         options : dict
             A dictionary of options and value to replace namelist options with
-            new values.
+            new values
 
         out_name : str, optional
             The name of the namelist file to write out, ``namelist.<core>`` by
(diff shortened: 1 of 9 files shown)


xylar:
@matthewhoffman, I'm mostly look for a review of the documentation changes from you.  Let me know if this would have helped you with the confusion you were having.

compass/step.py
```diff
@@ -340,6 +341,70 @@ def add_namelist_options(self, options, out_name=None, mode='forward'):
 
         namelist_list.append(dict(options=options, mode=mode))
 
+    def update_namelist_at_runtime(self, options, out_name=None):
+        """
+        Update an existing namelist file with additional options.  This would
+        typically be used for namelist options that are only known at runtime,
+        not during setup.  For example, the number of PIO tasks and the stride
```
matthewhoffman, line 348: Should this sentence be removed now with the new function you've added specifically for PIO?

docs/developers_guide/organization.rst
```diff
@@ -1604,12 +1604,34 @@ an option passed in when creating this step.
   Namelist values must be of type ``str``, so use ``'{}'.format(value)`` to
   convert a numerical value to a string.
 
+
+.. _dev_step_update_namelist_options:
```
matthewhoffman, line 1608: I like that you've moved this namelist topic to live with the other namelist info.

matthewhoffman (approved):
I left one minor comment about something you may want to fix in a docstring.  I like your change to the documentation itself.

By the way, there is a missing : on line 1518 of docs/developers_guide/organization.rst.  I tried to change it myself with a suggestion, but I think one can only make a suggestion for lines that were touched by the PR.

compass/step.py
```diff
@@ -340,6 +341,70 @@ def add_namelist_options(self, options, out_name=None, mode='forward'):
...
+        between tasks, which are related to the number of nodes and cores.
```
xylar, line 349:
```suggestion
        not during setup, typically those related to the number of nodes and cores.
```

compass/step.py
xylar: Good suggestion.

PR #143, opened 2021-06-12
Update mpas_tools to 0.7.0

xylar:
This brings in https://github.com/MPAS-Dev/MPAS-Tools/pull/426

Remove unused rasterio and affine dependencies.  These are not needed in compass (they aren't currently used).  Rasterio was removed from MPAS-Tools, and now conflicts with the latest version of that package because they have not been built with the same version of `libnetcdf`.

### DIFF
--- a/conda/compass_env/spec-file-mpich.txt
+++ b/conda/compass_env/spec-file-mpich.txt
@@ -18,7 +18,7 @@ jupyter
 lxml
 matplotlib-base
 metis
-mpas_tools=0.6.0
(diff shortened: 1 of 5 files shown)


xylar:
I'll merge this once the package is available.

xylar:
@matthewhoffman, FYI, this will bring in the version of MPAS-Tools you want as long as CI passes now.

matthewhoffman:
Thanks, @xylar !

PR #144, opened 2021-06-13
Fix updating namelist in gotm/default test case

xylar:
This got missed in #129 because the feature was changed in #142 but I didn't remember to rebase.


xylar:
## Testing

I recreated my compass development environment following #143, then I rebuilt MPAS-Ocean (because #143 brings in a new version of `libnetcdf` in Linux), then I ran the `gotm/default` test case with this fix, and it worked as expected:

velocity_profile
viscosity_profile

PR #145, opened 2021-06-13
Add a test merge to Azure

xylar:
The lack of a test merge caused trouble in #129


xylar:
It looks like this is already the default behavior.  The problem was not detected in #129 because CI was simply not rerun after other merges happened.

PR #146, opened 2021-06-14
Add FO versions of Greenland test cases

matthewhoffman:
This merge adds FO versions of the Greenland test cases, following the procedure used for the dome.  The tests are added to the appropriate test suites and the docs are updated.


matthewhoffman:
The 2 new tests pass and add a reasonable amount of time to the suite:
```
00:56 PASS landice_greenland_fo_decomposition_test
00:57 PASS landice_greenland_fo_restart_test
```

compass/landice/tests/greenland/decomposition_test/__init__.py
```diff
@@ -10,22 +10,33 @@ class DecompositionTest(TestCase):
...
         name = 'decomposition_test'
-        super().__init__(test_group=test_group, name=name)
+        self.velo_solver = velo_solver
+        subdir = '{}_{}'.format(velo_solver.lower(), name)
+        super().__init__(test_group=test_group, name=name, subdir=subdir)
+
+        if velo_solver == 'sia':
+            self.cores_set = [1, 8]
+        elif velo_solver == 'FO':
+            self.cores_set = [16, 32]
```
xylar, line 33:
```suggestion
        elif velo_solver == 'FO':
            self.cores_set = [16, 32]
        else:
            raise ValueError('Unexpected velo_solver {}'.format(velo_solver))
```
It's best to fail here rather than in an ambiguous way later on (e.g. a variable that isn't defined).

compass/landice/tests/greenland/decomposition_test/__init__.py
```diff
@@ -39,7 +50,43 @@ def validate(self):
...
+            else:
+                l1_norm = 0.0
+                l2_norm = 0.0
+                linf_norm = 0.0
+                quiet = True
+            compare_variables(test_case=self, variables=variable,
+                              filename1='{}/output.nc'.format(name1),
+                              filename2='{}/output.nc'.format(name2),
+                              l1_norm=l1_norm, l2_norm=l2_norm,
+                              linf_norm=linf_norm, quiet=quiet)
```
xylar, line 90: For 2 variables, this format is okay.  I think if you were validating more variables, you would want an outer ` if self.velo_solver == 'FO':` and then you would have a single call to `compare_variables()` for SIA cases and multiple calls for FO cases.  You also wouldn't need to define `quiet` because it is always either `True` or `False`.

compass/landice/tests/greenland/decomposition_test/__init__.py
```diff
@@ -39,7 +50,43 @@ def validate(self):
...
+        else:
+            assert False, "Error in decomposition test directory structure"
```
xylar, line 92: Several comments about this.  First, the most generic error in python should be a `ValueError`.  This is preferred if you can't come up with another type of error to raise.  `assert False` is, at least in my experience, discouraged. 

Second, this isn't going to work the way you want.  If you run just the `1proc_run` step (in the SIA configuration) on its own, which is a valid thing to do, you will trigger this error.  So the `else` should just be removed here.  We only do validation if both steps got run but it is valid to run only one of the steps.

compass/landice/tests/greenland/decomposition_test/__init__.py
xylar: The same for the other 2 test cases.

xylar:
@matthewhoffman, I suggest rebasing this onto `master` to pick up the changes in the `dev` environment.  After that, you may need to make your environment over again with the `--recreate` flag because it will potentially run into too many conflicts for it to just update on its own.  (maybe you'll be okay but I've had to recreate).

xylar:
I checked SIA tests and they are bit-for-bit with `master` (after I create symlinks to account for the name change).  So I think this looks good other than the small comments I made above.  Very nice to see this progress, and it seems like it's not proving to be too difficult (hopefully).

matthewhoffman:
>  I suggest rebasing this onto master to pick up the changes in the dev environment. 

Do you mean so we have access to the MPAS-Tools update you did for us?  I could also just merge this to master and work off master, right?

compass/landice/tests/greenland/decomposition_test/__init__.py
matthewhoffman: That's a good suggestion.  I'll take that approach if we encounter this situation again.

compass/landice/tests/greenland/decomposition_test/__init__.py
matthewhoffman: Ok, thanks.  Let me explain the situation I was trying to avoid.  Before my final commit, the decomposition test was passing because it was checking `if '1proc_run' in steps and '4proc_run' in steps:`, but neither of those dirs existed (due to my name change) so it was just skipping validation entirely and reporting a pass.  Maybe the solution is to remove the if-statement entirely and let the module err if it can't find the required dir (due to parsing errors or something else).

matthewhoffman: Oh, I see - that if -statement is only there to support running a single step without an error.  I'm wondering about the expected behavior in such a situation.  Running a single step of a decomposition test doesn't really make sense from a testing perspective - it's presumably just for debugging or some one-off sort of thing.  Getting a validation error in such a situation could be considered to technically be correct.

xylar:
> Do you mean so we have access to the MPAS-Tools update you did for us? I could also just merge this to master and work off master, right?

Yes, but to test with `master` as a baseline, you would have to either do a test merge or rebase (I did the latter).  So I thought it would be best to rebase now.  If you're able to do the testing your need without comparing with `master` as a baseline, you're free to merge without rebasing.

compass/landice/tests/greenland/decomposition_test/__init__.py
xylar: I guess this is a philosophical issue.  I don't think it's invalid to run one step of a test case that includes comparison.  As a user, I would not find it useful to get an error message telling me I hadn't run both steps of the test case if I purposefully ran only one.  I would possibly expect that I could manually run one and manually run the other and then get a validation.  But currently there's not a clear way to distinguish between the situation in which a step has run but didn't produce the expected output file (an error) or the case where the step hasn't run yet (presumably not an error).

If you want the validation to fail if both steps are not run, you can do that but I think the error message should be different if a user has run a single step (on purpose) and is getting an error vs. if the steps were not set up correctly.

It seems like you are trying to create validation here that would catch a development issue.  It seems like that would need to happen separately above this statement, in which you would check if the expected steps are, indeed, in the keys of the `steps` dictionary.

It could be that we could come up with a more sophisticated way of handling validation in general.  We could have a way to keep track of whether a step has been run successfully (e.g. it writes out a hidden file or a log or something).  The validation function could take the steps involved in the comparison, not just the file paths.  This would catch a case where the step name you are trying to validate with doesn't exist.  It would also allow you to check that the file is an output of the step, which should be a requirement.

Do you want to start an issue about this and we can discuss further before I write up a solution?

matthewhoffman: > It seems like you are trying to create validation here that would catch a development issue.

I think that's the crux of it here.  My misunderstanding of the purpose of the if-statement led me down that path.  I like your solution that I just check to make sure the module can find the 2 steps it expects before trying to do the validation.

Regarding your suggestion to do validation more sophisticated-ly, I think that is not necessary and a tangent we don't need to take from here.  I'll code up a fix doing what you suggest above and have you take a look at it.

matthewhoffman: > happen separately above this statement, in which you would check if the expected steps are, indeed, in the keys of the steps dictionary

I realized that `steps_to_run` will only have 1 step when running a single step manually, so I think the best solution for me here is to go back to the way you had things setup with the if-statement, and then do some further testing in development (i.e. now) to make sure my list of proc values is getting parsed properly.

compass/landice/tests/greenland/decomposition_test/__init__.py
matthewhoffman: Good suggestion.  I'm not up to speed on what error to raise, so this suggestion was perfect.

matthewhoffman:
My final retesting just completed successfully.  @xylar , please have a look again now when you have a chance.

xylar (approved):
Perfect, I think all of my comments have been addressed.  I'm not retesting on the assumption that we'll catch anything your testing might have missed soon enough.

matthewhoffman:
Thanks, @xylar.  I'll merge this now.

PR #148, opened 2021-06-14
Add Albany yaml files specific to Greenland configuration

matthewhoffman:
When I ported the test cases in the previous branch, I had inadvertently
used the Albany yaml file from the dome test case.


matthewhoffman:
@xylar , as it happens I discovered this omission right after merging the previous branch.  When it is something trivial like this in the landice core, do you still want me to have you review before I merge?

xylar:
Nope, no need. you can just merge.  But I'm sorry I didn't notice this in my review.

matthewhoffman:
No worries, I didn't notice either.

PR #152, opened 2021-06-15
Fix appending of global ocean metadata

xylar:
We were using `ncra` but should have been using `ncks` because we only want to add the metadata, not average the files in time.


xylar:
@mark-petersen, FYI, I found an important bug in the global ocean test cases.  In the process of appending metadata, I was accidentally averaging the data.  I discovered this when I got unexpected changes when I added more metadata in #151

xylar:
## Testing

Many global ocean tests in the `nightly` fail a comparison with `master` because of this branch.  However, I ran a baseline with `add_metadata = False` in the `[global_ocean]` section of the custom config file and used this as a new baseline.  With this baseline, all tests except the RK4 restart test (a known issue) pass with Linux and Gnu compilers and are bit-for-bit.

PR #153, opened 2021-06-15
Add missing output files to 3 ocean test groups

xylar:
The `global_ocean/analysis_test`, both `ice_shelf_2d` test cases, and the `ziso/frazil_test` are all missing output files that should be added with `add_output_file()`.  This was revealed by #151


xylar:
## Testing

This passes the nightly test suite on Linux with Gnu compilers using the same baseline as in #152 (i.e. without the metadata bug).

PR #149, opened 2021-06-14
Add Thwaites 4km test group to landice core

matthewhoffman:
This merge adds the Thwaites variable resolution 4-14km test group to the landice core.  This test group uses a pre-existing mesh and initial condition file and provides decomposition and restart tests.  The test structure follows closely to that used for the Greenland FO tests.  The tesst are added to the fo and full integration suites, and the test group is briefly described in the docs.  

In the future additional Thwaites tests cases could be added that actually build the mesh and initial condition.

PR #155, opened 2021-06-16
Skip steps not run by default in validation

xylar:
In `compass.validate.compare_variables()`, skip validation if one or more of the steps involved in the variable comparison were not run.  Log a message in such cases saying why validation was skipped.

### DIFF
--- a/compass/landice/tests/thwaites/decomposition_test/__init__.py
+++ b/compass/landice/tests/thwaites/decomposition_test/__init__.py
@@ -16,7 +16,7 @@ class DecompositionTest(TestCase):
 
         Parameters
         ----------
-        test_group : compass.landice.tests.thwaites
+        test_group : compass.landice.tests.thwaites.Thwaites
             The test group that this test case belongs to
 
         """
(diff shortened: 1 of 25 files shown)


xylar:
## Testing

I tested the `nightly` and `sia_integration` suites, using Linux, Gnu compilers, and today's `master` as a baseline.  All tests except RK4 restart passed, and all were bit-for-bit with the baseline.

xylar:
@matthewhoffman, I'm asking for your review because I modified `landice` test cases and because I'd like you to take a look at the revised documentation.

compass/validate.py
```diff
@@ -58,38 +58,68 @@ def compare_variables(test_case, variables, filename1, filename2=None,
...
         if not file1_found:
             raise ValueError('{} does not appear to be an output of any step '
                              'in this test case.'.format(filename1))
         if filename2 is not None and not file2_found:
             raise ValueError('{} does not appear to be an output of any step '
                              'in this test case.'.format(filename2))
 
+    if skip_if_step_not_run:
+        step1_not_run = (file1_found and
+                         step_name1 not in test_case.steps_to_run)
```
matthewhoffman, line 100: If a step is run manually (`compass run` from the step dir), does `steps_to_run` get updated  accordingly?  If not, this will not be accurate.

matthewhoffman (commented):
@xylar , this looks like a nice simplification for the developer.  I just have two questions related to robustness.  The changes to the landice tests and the explanation in the docs all look good.

compass/validate.py
```diff
@@ -58,38 +58,68 @@ def compare_variables(test_case, variables, filename1, filename2=None,
...
 
+    file1_found = False
+    file2_found = False
+    step_name1 = None
+    step_name2 = None
+    for step_name, step in test_case.steps.items():
+        for output in step.outputs:
+            # outputs are already absolute paths combined with the step dir
+            if output == path1:
+                file1_found = True
```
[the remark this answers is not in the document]
xylar: That's a good question.  No, the list of outputs are those identified by calling `add_output_file`.  There is no relationship to whether the file exists here, just whether it was identified as an output in a step.  `file1_found` doesn't literally mean that the file was found on disk but rather than the file was found as a listed output of a step in the test case, thus we know what step the file should have been created by.

compass/validate.py
xylar: Yes, when you run a single step, `steps_to_run` automatically gets set to a list with only that step in it, so this will work as expected.  This is also how things work now (before this PR), which is why all these test cases were checking to see if steps were in `steps_to_run`, just like is happening here.

compass/validate.py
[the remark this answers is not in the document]
matthewhoffman: Ok, I thought that might be the case.  Thanks for clarifying.

compass/validate.py
matthewhoffman: Ok, thanks - is that stored in the pickle file and then populated in the testcase object when `compass run` is run?

matthewhoffman:
@xylar , after your answers to my questions, I'm happy for you to merge this when you are ready.  We'll want to coordinate my new Thwaites tests to make this change.  Should I wait to merge that until you merge this, and then rebase my PR and make the corresponding changes?  Or vice versa?

compass/validate.py
xylar: I believe the pickle file stores the full list of `steps_to_run`, and it's here in the `run` module that it `steps_to_run` gets modified:
https://github.com/altheaden/compass/blob/master/compass/run.py#L224

xylar:
@matthewhoffman, go ahead with #149.  I'll rebase this once that goes in.

matthewhoffman:
Ok, thanks - I just merged #149

compass/landice/tests/thwaites/decomposition_test/__init__.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, test_group):
 
         Parameters
         ----------
-        test_group : compass.landice.tests.thwaites
+        test_group : compass.landice.tests.thwaites.Thwaites
```
xylar, line 19: @matthewhoffman, pycharm pointed out this type issue, which I took the liberty of fixing

xylar:
@matthewhoffman, would it be easy for you to run a quick test of the Thwaites test cases from this branch?  I don't have a setup for that.  I'd hate to break your work.

matthewhoffman:
I checked the Thwaites tests using your branch and they pass.  Go ahead and merge when you're ready.

xylar:
Great, thanks!

PR #156, opened 2021-06-16
Switch dev environment creation to use mamba

xylar:
This should speed up creating the development environment a bit


xylar:
## Testing

I tested recreating my development environment on Linux with this approach.  I'm not sure how much faster it really is (it may be that solving packages, which is what `mamba` improves isn't a big bottleneck for this process) but it worked fine.

PR #158, opened 2021-06-21
Add z-star ISOMIP+ Ocean0 test case

xylar:
This ports the ISOMIP+ Ocean0 test case ([Asay-Davis et al. 2016](http://doi.org/10.5194/gmd-9-2471-2016)) to the new `compass` package.

More description to follow soon.


xylar:
@cbegeman, this appears to be working in the sense that I can run for multiple months at 5 km resolution.  (I haven't tested that long a 2 km resolution but haven't seen any trouble).

I'm working out kinks in various places: viz, missing doc strings, etc.  I will clean up the commit history once that's taken care of.

I also need to add documentation for the test case.

For now, I'm making a PR as a starting point for discussion and so you can hopefully use the `initil_state` step of this test case to explore other vertical coordinates.

cbegeman:
@xylar This all looks good to me! I successfully ran all the steps but `misomip`, which I didn't try to run. I perused the scripts you added and nothing jumped out at me but let me know if you want me to take a closer look at anything.

xylar:
Great, thanks @cbegeman!  I did a bit more testing today and should be able to wrap up the testing I want to do tomorrow.  Shall I merge after that or did you want to do some additional testing?

cbegeman (approved):
@xylar You can go ahead and merge when you're ready. Many thanks for prioritizing this!

xylar:
## Testing

I tested this at 2 km and 5 km resolution for 2 years of simulation on Chrysalis.  This has never worked before with z-star, so recent MPAS-Ocean improvements must mean we can run at much higher Haney numbers than before!

However, there seems to be something wrong, since melt rates are continuously dropping.  I think the restoring at the northern boundary isn't working.

I also tested just the default tests with `draft_scaling = 0.5` and it looked as expected:
sectiontemp_0001

I also tested at 1 km resolution (which I manually added to the test group) and it didn't get through SSH adjustment unless I set `draft_scaling = 0.95`.  That's not super surprising, given the very thin layers that z-star produces and the higher Haney numbers at higher horizontal resolution.

xylar:
@cbegeman, I'm going to assume you don't need to re-review the PR based on recent changes.  I feel confident in my own testing but we can also always fix small issues in a later PR.  I will be adding Ocean1 and Ocean2 test cases soon anyway.

xylar:
@cbegeman, also, thank you very much for your testing!

PR #162, opened 2021-06-23
Fix activation script name when `--env_name` flag is used

xylar:
When running `./conda/configure_compass_env.sh` with `--env_name`, the name of the activation script now starts with `load_`.  That prefix was missing before.


xylar:
## Testing

I verified that, with this fix and using `--env_name mpas_dev`, I get the expected load script (`load_mpas_dev_mpich.sh` in my case).

PR #164, opened 2021-06-23
Add inactive top cell global ocean test case

cbegeman:
This test case is designed to determine the proper functioning of inactive top cells (i.e., `minLevelCell > 1`). The test follows the configuration of `QU240/PHC/performance_test` except an additional vertical layer is added and `minLevelCell` and `maxLevelCell` are shifted down the column by 1. A pass of this test constitutes bit-for-bit comparison with `QU240/PHC/performance_test`.


cbegeman:
@xylar At some point, it would be good to get your feedback on this but no rush. Both `init` and `performance_test` run, the latter only with tendency terms disabled. `performance_test` fails validation state and I'm still looking into why this is.

xylar:
Wow, @cbegeman, I'll take a more careful look at this soon but it looks really great so far!

compass/ocean/tests/global_ocean/init/initial_state.py
```diff
@@ -180,11 +180,9 @@ def run(self):
 
                 ds = ds.isel(Time=0)
 
-                if ('maxLevelCell' in ds) and ('minLevelCell' in ds):
+                if ('minLevelCell' in ds):
                     minLevelCell = ds.minLevelCell+1
-                    maxLevelCell = ds.maxLevelCell+1
```
cbegeman: `maxLevelCell` did not need to be increased here, but I don't have an explanation for it.

cbegeman:
Using this test case and [a branch with related MPAS-Ocean bugfixes](https://github.com/cbegeman/E3SM/tree/ocn/bugfix-inactive-top-cells), I have located two terms that lead to a non-bfb solution between `minLevelCell = 1` and `minLevelCell = 2` on chrysalis, intel-mpi:
* `config_disable_vel_pgrad = .false.` for `Jacobian_from_TS` or `Jacobian_from_density`
* `config_disable_vel_surface_stress = .false.`

I haven't yet found any obvious bugs in the relevant code.

cbegeman:
@xylar I think this test case is working as expected. Do you want to give it a look and let me know if other changes or testing are needed? Still no rush on this.

xylar:
Yep, I'll take a look soon.  I was hoping to look at this today but didn't get to it yet.

compass/ocean/tests/global_ocean/init/initial_state.py
xylar: @cbegeman, maybe you figured this out but `maxLevelCell` is determined based on the bathymetry and the 1D vertical coordinate.  Because you've altered the 1D vertical coordinate already to have an index that is one greater than normal for a given depth, the index corresponding to the bathymetry will already be incremented.

I was also expecting you would need to do this at first (and also move the bathymetry somehow) but I now realize that all you had to do was alter the 1D vertical coordinate.

xylar:
@cbegeman, this is really great!  

I tested a merge of this branch with #171, together with your branch https://github.com/cbegeman/E3SM/tree/ocn/bugfix-inactive-top-cells merged with `E3SM/master` with gnu compilers on Ubuntu.  Everything ran fine with QU240.  

For the QUwISC240 test case, I got NaNs in `layerThickness` during the SSH adjustment phase.  I didn't look into it too thoroughly but I suspect that there's some debugging still to be done there.  I ran the version without inactive top cells just to make sure it worked okay, and it did.

I have a suggested addition.  I would be really nice to have another test case under global ocean for each configuration with inactive top cells.  It would only work if you had run the init and performance tests both with and without inactive top cells, so it would be helpful to have test suites defined for this check.  (The QU240 version could also be included in the nightly test suite.) 
This test case would only have one step.  First, it would use xarray to remove the top layer from both `init/initial_state/initial_state.nc` and `performance_test/forward/output.nc` and save the results in new files local to the step.  Then, it would compare to make sure the cropped files are identical to the results form the run without inactive top cells using the usual validation method (`compass.validate.compare_variables()`).

If results are not identical when comparing `initial_state.nc` in this way, we may have to see if the 1D vertical coordinates are identical except for the top layer.  It could be that they are only close.

My other request would be to clean up the PEP8 issues that are mentioned here https://github.com/MPAS-Dev/compass/pull/164#issuecomment-867180499 and which will update automatically as you push changes to this PR branch.

xylar:
When I run in debug mode, I'm seeing errors in `initial_state`.  Maybe these are things we can include in https://github.com/cbegeman/E3SM/tree/ocn/bugfix-inactive-top-cells?

```
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7fd554d2220f in ???
#1  0x563c7dc8277e in __ocn_init_vertical_grids_MOD_ocn_compute_haney_number
	at /home/xylar/code/compass/add_inactive_top_cell/bugfix-inactive-top-cells-debug/components/mpas-ocean/src/mode_init/mpas_ocn_init_vertical_grids.F:795
#2  0x563c7d970552 in __ocn_init_global_ocean_MOD_ocn_init_setup_global_ocean
	at /home/xylar/code/compass/add_inactive_top_cell/bugfix-inactive-top-cells-debug/components/mpas-ocean/src/mode_init/mpas_ocn_init_global_ocean.F:511
#3  0x563c7d94b84a in __ocn_init_mode_MOD_ocn_init_mode_run
	at /home/xylar/code/compass/add_inactive_top_cell/bugfix-inactive-top-cells-debug/components/mpas-ocean/src/mode_init/mpas_ocn_init_mode.F:307
#4  0x563c7d7f9a24 in __ocn_core_MOD_ocn_core_run
	at /home/xylar/code/compass/add_inactive_top_cell/bugfix-inactive-top-cells-debug/components/mpas-ocean/src/driver/mpas_ocn_core.F:116
#5  0x563c7d257990 in __mpas_subdriver_MOD_mpas_run
	at /home/xylar/code/compass/add_inactive_top_cell/bugfix-inactive-top-cells-debug/components/mpas-framework/src/driver/mpas_subdriver.F:358
#6  0x563c7d256689 in mpas
	at /home/xylar/code/compass/add_inactive_top_cell/bugfix-inactive-top-cells-debug/components/mpas-framework/src/driver/mpas.F:20
#7  0x563c7d2566fa in main
	at /home/xylar/code/compass/add_inactive_top_cell/bugfix-inactive-top-cells-debug/components/mpas-framework/src/driver/mpas.F:10
```

xylar:
Actually, regarding my last comment, I think the issue is that the top layer has zero thickness.  So the issue isn't to do with init mode in MPAS-Ocean at all, but rather that the vertical coordinate it's being given isn't valid.  Could we give the vertical coordinate a non-zero, positive value for the first entry so it doesn't cause divide-by-zero issues?  Or does that cause trouble with computations elsewhere?

cbegeman:
@xylar Thanks for all your feedback and suggestions.

Is there a reason that you propose adding an additional test case that does the xarray trim and then validate, rather than adding a new method or function `crop_output` (or `remove_inactive_top_cells_output`) and adding
```
if self.init.with_inactive_top_cells:
     crop_output(test_case=self, variables=variables, filename1=forward/output.nc)
```
to either `performance_test/__init__.py` (called from `validate`, new method in `PerformanceTest` class) or to `validate.py` (called from `compare_variables`, new function in `compass.validate`). I'm thinking that it would be desirable to just run the inactive cells init and performance tests and have the performance test do the comparison with baseline automatically if the baseline is provided.

xylar:
@cbegeman, that would be a good way to go.  I hadn't thought about it.  I would suggest adding a `validate` to the `Init` class as well that does the same thing.  So it would make sense to make `remove_inactive_top_cells_output()` (the longer name is better in my view) live somewhere outside of any individual test case or step in the `global_ocean` infrastructure.

Presumably, you will just need to check whether the files `PHC/init/initial_state/initial_state.nc` and `PHC/performance_test/forward/output.nc` are present, and decide whether to perform each validation only if so?  I don't think there's currently a good way to know whether those steps got run except by whether their output exists.

xylar:
@cbegeman, I thought about this a bit further.  We don't usually put any output files at the test-case level, just within steps.  So it would be better if the output gets cropped as part of the `init/initial_state` and `performance_test/forward` steps, respectively, and the cropped output ends up in the directories for those steps.  This will become more important when we implement task parallelism, as it is an important aspect of determining data dependencies.  It would also be important that the cropped files are included as outputs of the steps.

cbegeman:
@xylar Thanks for giving it some thought. I think I understand how to implement it as you suggested. 

Do you mind if I throw the `remove_inactive_top_cells_output()` function in validate.py and I'll just import it for the two steps you mentioned?

xylar:
> Do you mind if I throw the `remove_inactive_top_cells_output()` function in validate.py and I'll just import it for the two steps you mentioned?

Hmm, that doesn't really seem like the right place or it to me.  `validate.py` is general infrastructure for all of `compass`, whereas what you are proposing belongs within the `ocean` package for sure and probably within the `global_ocean` test group unless it would conceivably get used somewhere else.  I would suggest creating a new module either in `ocean` or `global_ocean` for this function.  I don't think there is an existing module you can use without it seeming oddly out of place.

cbegeman:
@xylar You can ignore this while on vacation.

I didn't know the best way to set up a comparison with the baseline because the baseline has a different test name than the inactive cell run (e.g., `PHC` vs. `PHC_inactive_top`). Right now, the comparison requires that you copy the baseline files into the output directories for `PHC_inactive_top` tests, which isn't ideal.

Other than that I think the test is good to go, and I added the check for zero layer thickness in init mode to that bugfix branch.

xylar:
@cbegeman, thanks for working on this. I think I see the dilemma: validation was only designed either for comparison with a baseline or within a test case, not between test cases. I'll give it some thought and try to come up with a better way when I come back.

xylar:
@cbegeman, I'm getting back to thinking about the validation problem here.  I'm going to try to come up with a general solution outside of this branch/PR so we can rebase this onto that work.

cbegeman:
Thanks, @xylar!

compass/ocean/tests/global_ocean/init/__init__.py
```diff
@@ -46,22 +47,25 @@ def __init__(self, test_group, mesh, initial_condition, with_bgc):
...
         name = 'init'
         mesh_name = mesh.mesh_name
+        ic_dir = initial_condition
         if with_bgc:
-            ic_dir = '{}_BGC'.format(initial_condition)
-        else:
-            ic_dir = initial_condition
+            ic_dir = '{}_BGC'.format(ic_dir)
+        if with_inactive_top_cells:
+            ic_dir = '{}_inactive_top'.format(ic_dir)
```
xylar, line 54: Here, it will also be useful to store the subdirectory for the "original" test case we want to compare with, as I will explain below.
```suggestion
        self.inactive_top_comp_subdir = None
        if with_inactive_top_cells:
            self.inactive_top_comp_subdir = os.path.join(mesh_name, ic_dir, name)
            ic_dir = '{}_inactive_top'.format(ic_dir)
```
Also, make sure to add ``inactive_top_comp_subdir`` as an attribute in the docstring for the class above.

compass/ocean/tests/global_ocean/init/__init__.py
```diff
@@ -117,6 +121,14 @@ def validate(self):
...
                               filename1='initial_state/initial_state.nc')
 
+        if self.with_inactive_top_cells:
+            variables = [
+                'temperature', 'salinity', 'layerThickness', 'normalVelocity']
+            compare_variables(test_case=self, variables=variables,
+                              filename1='initial_state/initial_state_crop.nc',
+                              filename2='initial_state/initial_state_comp.nc',
+                              quiet=False, check_outputs=False)
+
```
xylar, line 131: I'm guessing you manually copied `initial_state/initial_state_comp.nc` from the other test case.  I think this comparison would also work if you constructed the absolute path to the original output file from the other test case and also set `skip_if_step_not_run=False` (which is not ideal, and I'm working on a better approach as a separate PR).  To construct this path, you would need to use `self.inactive_top_comp_subdir` that I suggested above, and then do a bit of path manipulation: 
```suggestion
        if self.with_inactive_top_cells:
           # construct the work directory for the other test
            work_dir2 = os.path.join(self.base_work_dir, self.mpas_core, self.test_group, 
                                     self.inactive_top_comp_subdir)
            filename2 = os.path.join(work_dir2,  'initial_state/initial_state.nc)
            variables = [
                'temperature', 'salinity', 'layerThickness', 'normalVelocity']
            compare_variables(test_case=self, variables=variables,
                              filename1='initial_state/initial_state_crop.nc',
                              filename2=filename2, quiet=False, check_outputs=False, 
                              skip_if_step_not_run=False)

```

xylar (commented):
@cbegeman, sorry this is taking so long to get merged.  I have another round of comments, but we're getting close.

The biggest things are:
1. a way to avoid needing to manually copy the output files for comparison that I think should work.  Validation will fail if the "comparison" test case (i.e. the one without inactive top cells) was not run, but I think that's okay for now.
2. there needs to be a way for the forward test case (`PerformanceTest`) to find out the value of `inactive_top_cells` from the output of `init/initial_state`, rather than using its config file as I explain in my comments.

Let me know if you want to chat about either or both of these (or any of my other comments).  Once those 2 get ironed out, I think we'll be in good shape!

compass/ocean/tests/global_ocean/init/initial_state.py
```diff
@@ -160,6 +170,36 @@ def run(self):
...
+            out_filename = in_filename
+
+            with xarray.open_dataset(in_filename) as ds:
+
+                # keep the data set with Time for output
+                ds_out = ds
+
+                ds = ds.isel(Time=0)
+
+                if ('minLevelCell' in ds):
```
xylar, line 187:
```suggestion
                if 'minLevelCell' in ds:
```
These parentheses aren't needed.  It doesn't seem to trigger a PEP8 warning but I think we want to avoid them in the standard compass coding style nonetheless.

There are some other PEP8 warnings about trailing white space that it would be good to have fixed as well.

compass/ocean/tests/global_ocean/init/initial_state.py
```diff
@@ -160,6 +170,36 @@ def run(self):
...
+                                         'inactive_top_cells'):
+                        offset = config.getint('vertical_grid', 
+                                               'inactive_top_cells')
+                    minLevelCell = ds.minLevelCell+offset
+                    ds_out['minLevelCell'] = minLevelCell
+                else:
+                    logger.info("   - Streams missing for inactive top cells")
+
+            remove_inactive_top_cells_output(in_filename, 
+                                             inactive_top_cells=offset)
```
xylar, line 198:
```suggestion
                if config.has_option('vertical_grid', 'inactive_top_cells'):
                    offset = config.getint('vertical_grid',
                                           'inactive_top_cells')
                else:
                    offset = 0

                if 'minLevelCell' in ds:
                    minLevelCell = ds.minLevelCell+offset
                    ds_out['minLevelCell'] = minLevelCell
                else:
                    logger.info("   - Variable minLevelCell, needed for "
                                "inactive top cells, is missing from the "
                                "initial condition")

            remove_inactive_top_cells_output(in_filename, 
                                             inactive_top_cells=offset)
```
The logic here needs to be rearranged so that `offset` is always defined since you use it int the call to `remove_inactive_top_cells_output()`.  I think a more verbose output is needed as I have suggested.  But I actually think you should have this raise a `ValueError` instead of just printing a log message.  If `minLevelCell` is missing, something has gone wrong.

compass/ocean/inactive_top_cells.py
```diff
@@ -0,0 +1,28 @@
...
+    Remove inactive top cells from the output netCDF file
+
+    Parameters
+    ----------
+    in_filename : str
+        Filename for the netCDF file to be cropped
+
+    inactive_top_cells : str, optional
+        The number of inactive top cell layers. It should be equal to
+        ``config.inactive_top_cells``.
```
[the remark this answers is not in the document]
xylar: I take back what I said about passing the `config`.  That won't work because you want to also call this function from a `ForwardStep`, where the config option is not available (or, really, not useful because it doesn't necessarily have the same value as in the `Init` test case).

My suggestion here would be to not pass the `inactive_top_cells` at all bur rather to read `minLevelCell` from the file, raise an error if it's not a constant value, an use that constant value as the offset.  This should be a good solution in the user has changes `inactiv_top_cells` in the config file in `init`.

compass/ocean/tests/global_ocean/mesh/qu240/qu240.cfg
```diff
@@ -16,6 +16,7 @@ min_layer_thickness = 3.0
 # The maximum layer thickness
 max_layer_thickness = 500.0
 
+inactive_top_cells = 0
```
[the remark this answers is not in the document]
xylar: Thinking about this more, it seems like it only belongs in `init`, so `compass/ocean/tests/global_ocean/init/init.cfg` (which hasn't been needed up to now).

compass/ocean/inactive_top_cells.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    if not os.path.exists(in_filename):
+        raise OSError('File {} does not exist.'.format(in_filename))
+
+    out_filename = in_filename.split('.')[0] + '_crop.nc'
+
+    with xarray.open_dataset(in_filename) as ds_in:
+        ds_out = ds_in.isel(nVertLevels=slice(1, None))
```
xylar, line 26: Here would be my suggested way of handling the offset.  The `inactive_top_cells` would no longer be needed, which is important for implementation elsewhere.

```suggestion
        minLevelCell = ds.minLevelCell
        minval = minLevelCell.min().values
        maxval = minLevelCell.max().values
        if minval != maxval:
            raise ValueError('Expected minLevelCell to have a constant value '
                             'for inactive top cell tests')
        ds_out = ds_in.isel(nVertLevels=slice(minval, None))
```

compass/ocean/tests/global_ocean/performance_test/__init__.py
```diff
@@ -61,6 +62,18 @@ def validate(self):
         compare_variables(test_case=self, variables=variables,
                           filename1='forward/output.nc')
 
+        if self.config.has_option('vertical_grid', 'inactive_top_cells'):
+            offset = self.config.getint('vertical_grid', 'inactive_top_cells')
+            if offset > 0:
+                remove_inactive_top_cells_output('forward/output.nc',
+                                                 inactive_top_cells=offset)
```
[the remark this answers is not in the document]
xylar: Both here and in `ForwardStep`, you can use `self.init.with_inactive_top_cells` to find out if this forward step was started from an initial condition with inactive top cells.

compass/ocean/inactive_top_cells.py
xylar: Hmm, there's a complication with my suggestion.  The mesh isn't included in `output.nc` files you default.  We could add it but I think the better solution is also supplying a mesh file name (`None` by default) so you can get `minLevelCell` from it.

compass/ocean/tests/global_ocean/init/initial_state.py
```diff
@@ -160,6 +170,36 @@ def run(self):
...
 
+        if self.with_inactive_top_cells:
+
+            logger.info("   * Updating minLevelCell for inactive top cells")
+
+            in_filename = 'initial_state.nc'
+            out_filename = in_filename
+
+            with xarray.open_dataset(in_filename) as ds:
+
```
xylar, line 181: In my testing, `xarray` is getting confused in the following.  Because of its "lazy" loading of files, it doesn't deal well with a file being read in and written out to the same file name.  The best way around this is to call `ds.load()` immediately after opening the file to make sure the file is read into memory.  After doing that, things work as expected.
```suggestion
            with xarray.open_dataset(in_filename) as ds:

                ds.load()
```

xylar:
@cbegeman, I realize I've asked for a ton of changes and it might be pretty confusing.  I made my own version of this branch, rebased onto `master`.  It works but gives errors for me during validation (partly because I was using using `E3SM/master` rather than the bug-fix branch in my tests).

https://github.com/xylar/compass/tree/add_inactive_top_cell

You could use my changes just as an easier way to see what my suggestions are and make the changes you like on your own branch, our you can include my commits on your branch if you prefer.

Once these or similar changes are in place, I'll help figure out what the remaining bugs are in MPAS-Ocean itself.

cbegeman:
Thank you, @xylar, for making all these suggestions and I really appreciate that you implemented them as well! I have a few things on my plate that I'm finishing off but I think I'll get to this in a couple days.

xylar:
Just a note to say that I plan to return to this work in the next few days.

cbegeman:
> I have located two terms that lead to a non-bfb solution between `minLevelCell = 1` and `minLevelCell = 2` on chrysalis, intel-mpi:
* `config_disable_vel_pgrad = .false.` for `Jacobian_from_TS` or `Jacobian_from_density`
* `config_disable_vel_surface_stress = .false.`

@xylar Ok, let's merge this with the two terms that cause the non-BFB result disabled so it is BFB. I'll re-test and hopefully push that change shortly.

xylar:
Sounds good!

xylar:
@cbegeman, the rebase may be challenging at this point.  If you want me to do it (or parts of it), let me know.

cbegeman:
@xylar I've finished the rebase and made a few changes to how the case works. I think it makes more sense to have `performance_test_inactive_cells` and `init_inactive_cells` which are compared to `performance_test` and `init` if those are run. Formerly, these tests were set up at `{initial_condition}_inactive_cells`. The consequence of this is that it isn't convenient to have different namelist settings for the inactive cells test because they have to be modified at both `performance_test` and `performance_test_inactive_cells` to make the comparison valid. Should we wait to merge this PR until we can get PASS for the comparison with the default namelist settings?

There is one lingering issue I spotted but didn't get around to fixing. The initial condition ends up with real values at k=1:minLevelCell-1. It doesn't affect the test outcome because those values are not used, but I think we should set them to the fill value.

PR #166, opened 2021-06-24
Add global ocean test case with Lagrangian particles

mark-petersen:
Add a test that creates an initial condition particle file, then runs LIGHT for the forward global ocean run.

### DIFF
--- a/compass/ocean/tests/global_ocean/init/initial_state.py
+++ b/compass/ocean/tests/global_ocean/init/initial_state.py
@@ -1,8 +1,9 @@
 from compass.ocean.tests.global_ocean.metadata import \
     add_mesh_and_init_metadata
-from compass.model import run_model
+from compass.model import partition, run_model
 from compass.ocean.vertical.grid_1d import generate_1d_grid, write_1d_grid
 from compass.ocean.plot import plot_vertical_grid, plot_initial_state
+from compass.ocean import particles
 from compass.step import Step
 
 
@@ -162,6 +163,12 @@ class InitialState(Step):
 
         add_mesh_and_init_metadata(self.outputs, config,
                                    init_filename='initial_state.nc')
+
+        cores = self.cores
+        partition(cores, self.config, self.logger)
+        particles.write(init_filename='initial_state.nc', particle_filename='particles.nc',
+                        graph_filename='graph.info.part.{}'.format(cores),
+                        types='buoyancy')
 
         plot_initial_state(input_file_name='initial_state.nc',
                            output_file_name='initial_state.png')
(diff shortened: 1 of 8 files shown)


compass/ocean/tests/global_ocean/init/initial_state.py
```diff
@@ -163,5 +164,9 @@ def run(self):
         add_mesh_and_init_metadata(self.outputs, config,
                                    init_filename='initial_state.nc')
 
+        particles.write(init_filename='initial_state.nc', particle_filename='particles.nc',
+                        graph_filename='graph.info.part.{}'.format(cores),
+                        types='buoyancy')
```
mark-petersen, line 169: I copied this particle creation line from the ziso test. I wasn't sure where to put it - here (would run every time) or within the `lagrangian_particles` test.

compass/ocean/tests/global_ocean/lagrangian_particles/__init__.py
```diff
@@ -0,0 +1,82 @@
+from compass.validate import compare_variables, compare_timers
+from compass.ocean.tests.global_ocean.forward import ForwardTestCase, \
+    ForwardStep
+
+
+class PerformanceTest(ForwardTestCase):
```
xylar, line 6:
```suggestion
class LagrangianParticles(ForwardTestCase):
```

compass/ocean/tests/global_ocean/lagrangian_particles/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    A test case for performing a short forward run with an MPAS-Ocean global
+    initial condition assess performance and compare with previous results
```
xylar, line 9: Replace this

compass/ocean/tests/global_ocean/init/initial_state.py
xylar: I think this belong as an optional part of the `ForwardStep`.  You should add an argument to the constructor of `ForwardStep` to indicate that particles should be included.  Then, store that argument in the class (e.g. `self.with_particles = with_particles`).  In the `run()` method, you can call this function.  That will be with the number of cores used for running the forward step, not for init (which is what would happen here, and which isn't appropriate).

xylar:
I gather that your test case is mostly still a copy of the `PerformanceTest` at this point, right?  Sorry if I jumped the gun with some of my comments.

compass/ocean/tests/global_ocean/init/initial_state.py
xylar: Let me know if this is something you'd like help with.  I could make the commit to update `ForwardStep` if that would help you make progress on this PR.

xylar:
@mark-petersen, could you close this one for now and re-open it when you're ready to make progress?

mark-petersen:
This is no longer a priority. Closing for now.

PR #167, opened 2021-06-24
Add landice circular shelf test group

matthewhoffman:
This merge ports the landice circular shelf test group from legacy COMPASS.  The circular shelf decomposition test is added to the FO and full integration suites.


matthewhoffman:
@xylar , there are a number of PEP8 errors in the visualization step that I did not want to fix.  It's a legacy script that will honestly probably never be run, and the formatting would be a pain to correct.

I confirmed the new test passes on Badger.

xylar:
@matthewhoffman, I'm taking a look now.  It seems like the test case and its steps need to be added to `landice/api.rst` in the developer's guide.  Other than that, the documentation looks great!

I was able to set up the test case fine but wasn't able to run because I'm not set up to run with Albany yet.

I didn't see anything concerning about the code itself.  I'm fine with ignoring the PEP8 errors on the `visualize` module. They would be easy to fix but I can understand not wanting to bother.  Please do fix the docstring that's too long in the other file.  I don't see any reason that line needs to be longer than 79 characters.  Maybe you have 80 as your limit in vim?  If so maybe reduce to 79 because that's the PEP8 standard.  Presumably that's because of weirdness that happens when lines are exactly 80 characters in an 80-character terminal window.

xylar (approved):
Approved on the assumption that you'll add the missing API.

matthewhoffman:
@xylar , is there a correct procedure for updating the API page?  I don't really understand what I'm supposed to do for that.

xylar:
@matthewhoffman, yes, you need to add all of the "public" functions, classes and methods from the new test group.  Otherwise, they cannot be link to in the developer guide (all of your current references are dead links).  When you build the documentation, any function, class or method referenced there gets its docstring turned into a documentation page.  The page itself gets turned into a series of tables.  

Please see if you can duplicate the example of the other test groups.  If you have trouble, I can do it and push to your branch.

matthewhoffman:
@xylar , I see, thanks for the clarification.  So the lists of classes and methods to be indexed needs to be manually selected, correct?  It is pretty easy to follow a previous test group as an example, but I wasn't sure if there was some automation or criteria I wasn't aware of.

xylar:
>  So the lists of classes and methods to be indexed needs to be manually selected, correct? 

Yes, sphinx doesn't attempt to guess automatically what is part of the public API and what is not, so you have to add things manually.

PR #168, opened 2021-06-25
Add ISOMIP+ Ocean1 and Ocean2 test cases

xylar:
These are the remaining test cases for the `isomip_plus` test group that MPAS-Ocean can currently support (without moving grounding lines).


xylar:
## Testing

I have run the `initial_state`, `ssh_adjustment`, `performance`, `simulation`, `streamfunction` and `viz` steps at 2 and 5 km resolution for Ocean1 and Ocean2 on Chrysalis.  I have run all simulations for at least 1 month and some for 2 years.  Results look as expected, with cold cavities and warm restoring for Ocean1, and warm cavities with cold restoring and a more retreated ice-shelf geometry for Ocean2.

Ocean1:
sectionTemp_0024

Ocean2:
sectionTemp_0024

PR #174, opened 2021-07-07
Add a troubleshooting section to the developer's guide

xylar:
This will be an area for collecting common issues that developers run into and their likely solutions.

### DIFF
--- a/docs/developers_guide/troubleshooting.rst
+++ b/docs/developers_guide/troubleshooting.rst
@@ -28,7 +28,7 @@ you may run into an error like:
 
     subprocess.CalledProcessError: ...
 
-Details of teh error may vary but the message indicates in some way that there
+Details of the error may vary but the message indicates in some way that there
 was a problem solving for the requested combination of packages.  This likely
 indicates that you have an existing compass development environment
 (``dev_compass_*``) that can't be updated to be compatible with the new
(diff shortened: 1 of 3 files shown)


xylar:
@altheaden, could you let me know if this describes the problem you ran into in a way that makes sense to you, and if the solution works for you?

docs/developers_guide/troubleshooting.rst
```diff
@@ -0,0 +1,46 @@
...
+.. code-block:: none
+
+    Encountered problem while solving:
+      - nothing provides geos 3.5.* needed by cartopy-0.14.3-np110py27_4
+
+    ...
+
+    subprocess.CalledProcessError: ...
+
+Details of teh error may vary but the message indicates in some way that there
```
altheaden, line 31: Typo - "teh" instead of "the"

altheaden:
That sounds good to me! You could add a line about what `--recreate` does, compared to just updating, but it's not necessary for solving the problem. Maybe the explanation would be helpful, or maybe the troubleshooting guide should be left as simple as possible - I think that's a design decision.

docs/developers_guide/troubleshooting.rst
xylar: Fixed!

xylar:
@altheaden, thanks a lot for taking a look.  I added a bit more text about what `--recreate` does.  Let me know if that looks good (and if you notice more typos!).

altheaden:
@xylar everything looks good to me!

PR #177, opened 2021-07-12
Update E3SM-Project submodule

xylar:
There is no super strong reason to do this but it would be nice to have the latest as a baseline.


xylar:
@mark-petersen, are you fine with me updating to the current https://github.com/E3SM-Project/E3SM

mark-petersen (approved):
Yes, it is good to update. I tested this PR beginning to end, with
```
./conda/configure_compass_env.py --conda ~/miconda3 --compiler gnu
source load_dev_compass_1.0.0_grizzly_gnu_mvapich.sh
```
on grizzly, then compiled mpas-ocean in the E3SM-Project submodule, and it passes the nightly regression suite, except for the ones caused by other issues (RK4 restart https://github.com/E3SM-Project/E3SM/pull/4369, global stats top level https://github.com/E3SM-Project/E3SM/pull/4387). We can update the E3SM-Project submodule again after those are merged.

xylar:
Great, thanks for testing, @mark-petersen!

PR #178, opened 2021-07-14
Better document and configure conda envs. on Linux and Mac OS

xylar:
First, this merge adds an error that is raised when the `--mpi` flag is missing (and `--env_only` was not specified) on an "unknown" machine (Linux or Mac).  The hope is to avoid unexpectedly creating an environment without MPI support.

Second, it adds what is hopefully better documentation that clarifies the need to specify the ``--mpi`` flag when installing on an Linux and Mac OS.

### DIFF
--- a/conda/configure_compass_env.py
+++ b/conda/configure_compass_env.py
@@ -883,6 +883,11 @@ def main():
         activ_path = get_env_setup(args, config, machine, is_test, source_path,
                                    conda_base)
 
+    if machine is None and not args.env_only and args.mpi is None:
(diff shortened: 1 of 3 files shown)


xylar:
@vanroekel, can you take a look at these changes and let me know if they would have lessened the confusion you were having yesterday?

vanroekel (approved):
This is really great @xylar! I think this will be helpful for others like me just coming to compass on a personal machine.  Thanks for the added documentation and error without mpi!

xylar:
@vanroekel, thanks for the review.

PR #180, opened 2021-07-15
Update `mpas_tools` to 0.8.0 

xylar:
This should take care of an issue with meshes that are run through `MpasMeshConverter.x` that end up with garbage values in `kiteAreasOnVertex`.  This issue was causing trouble in tests in debug mode, fixed in https://github.com/MPAS-Dev/MPAS-Tools/pull/431


xylar:
@mark-petersen and @matthewhoffman, just a heads up that I'm updating the MPAS-Tools in `compass`.

PR #181, opened 2021-07-20
Move "Building MPAS components" earlier in the developer's quick start

xylar:
@altheaden pointed out that it's pretty confusing to have instructions for setting up a test case or suite before the instructions for building the model, since the command for setting up the suite fails when the model hasn't been build yet.

### DIFF
--- a/docs/developers_guide/quick_start.rst
+++ b/docs/developers_guide/quick_start.rst
@@ -190,6 +190,31 @@ to the current directory.
     not up-to-date.  Since dependencies change fairly rarely, this will usually
     be safe.
 
+
+Building MPAS components
+------------------------
+
+The MPAS repository is a submodule of compass repository.  For example, to
+compile MPAS-Ocean:
+
+.. code-block:: bash
+
+    source ./load_dev_compass_<version>_<machine>_<compiler>_<mpi>.sh
+    cd E3SM-Project/components/mpas-ocean/
+    make <mpas_compiler>
+
+For MALI:
+
+.. code-block:: bash
+
+    source ./load_dev_compass_<version>_<machine>_<compiler>_<mpi>.sh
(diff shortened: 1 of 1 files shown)


xylar:
@matthewhoffman, you are probably the person who has paid the most attention to the documentation.  If you want to review this, feel free.  If not, that's fine too.  We're just moving a section of the quick start a bit earlier.  No other changes to the text.

xylar:
@altheaden, have a quick look and make sure this would have helped you out if things had been in this order to begin with.

altheaden (commented):
This change looks good to me - now the line about "if you follow all the above steps you can run compass..." will be a little bit more true!

altheaden:
I did notice one more confusing thing in the dev quick start, but I don't know if you want to create a secondary PR for this since it's not about moving the building section. There was no mention of the `-f` flag to indicate a config file but I found that I needed this to run the setup (it would tell me I needed to specify a config file otherwise).  It seems to me like this should be a step in the dev quick start as opposed to having to go to the `Command-line interface` section, and it's actually mentioned in the `Quick Start for Users` section as well!

altheaden:
Actually @xylar and @matthewhoffman if you're OK with it, I can just make another PR myself to edit `Quick Start for Developers` and add this info about the `-f` flag and additionally fix a couple typos I just noticed going through the page.

xylar:
@altheaden, go ahead and make another PR. That sounds good.

Regarding the `-f` option, in your changes please make clear that it is only required on unknown machines.  The reason it is probably missing in the quick start is because this has been target at supported machines, but that doesn't need to be the case.  The important thing is to make clear that `-f` is needed on your personal machine but not on a supported machine.

xylar:
@matthewhoffman, given that @altheaden wants to make changes in this same file, I just going to merge to make things easier.

PR #183, opened 2021-07-20
Update Quick Start for Developers

altheaden:
Quick Start needed info about config files and some misc typos fixed.

### DIFF
--- a/docs/developers_guide/quick_start.rst
+++ b/docs/developers_guide/quick_start.rst
@@ -48,6 +48,11 @@ workstation, you will need to specify which flavor of MPI you want to use
 We only support one set of compilers for Mac and Linux, so there is no need to
 specify them.  See :ref:`dev_other_machines` for more details.
 
+In addition, unknown machines require a config file to be specified when setting
+up the compass test environment.  A config file can be specified using
+``-f <filename>``.  More information, including example config files, can be found
+in :ref:`config_files`.
+
 If you don't have `Miniconda3 <https://docs.conda.io/en/latest/miniconda.html>`_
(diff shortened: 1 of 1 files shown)


altheaden:
@xylar, I wrote something up about config files, but I have no idea if the `:ref:` was set up correctly and I'm not sure how to check. Also, I'm not sure how everyone will feel about linking to the user's guide, so if you want it to link somewhere else (maybe `Command-line interface`?) I can do that.

altheaden:
OK, thinking over this as I do more things. I was getting an error trying to set up the compass dev environment without a config file, but I'm realizing that's not something that I usually need to specify for it to work. It seems like this was more of an isolated incident than I realized initially, and saying `-f` is always required on an unknown machine is not true. Still think `-f` is worth mentioning in case a person gets the same error, but let me know your thoughts.

xylar:
@altheaden, I think you're original idea was right. Any unknown machine will require a config file. So I think your changes are correct and appropriate as they are.

altheaden:
@xylar, sounds good, just wanted to double check!

xylar (approved):
This looks great!  I built the docs locally and everything looks good.  Thanks for the work!

xylar:
@altheaden, I believe that you don't have permission to merge the PR since you are in the "triage" group on GitHub.  On that assumption, I'm going to assign this to myself and merge.  Let me know if you see a green "Merge pull request" button on compass PRs.  If so, you have more permission than I thought.  If not, that's fine.  I'm happy to play "gatekeeper" to some extent.

xylar:
You can delete the branch now if you want to keep things tidy.

PR #187, opened 2021-07-28
Add support for specifying the NetCDF output engine

### LINKED ISSUE #185: Writing of large netcdf files (> 3.4 GB) with `write_netcdf()` is hanging
I don't yet know what the cause is or when this behavior started.  There have been updates to `libnetcdf` and `xarray`, among other related libraries, since the last time I tried to run the `SOwISC12to60` test cases.  My guess is that it's something specific to the `NC_FORMAT_64BIT_OFFSET` format that we continue to have to use because of E3SM standards.

So far, I have only seen the behavior during the `SOwISC12to60/PHC/init/ssh_adjustment` step on Chrysalis.  I don't believe other test cases we currently have will face this problem either because they don't use xarray to write large files (e.g. WC14) or because they don't have meshes this large.

xylar:
This is useful because the `scipy` engine seems to perform well on Anvil and Chrysalis, whereas the `netcdf4` engine seems to be hanging.

This merge also updates subprocess calls in `global_ocean` to specify the NetCDF engine.

This requires https://github.com/MPAS-Dev/MPAS-Tools/pull/434

closes #185

### DIFF
--- a/compass/default.cfg
+++ b/compass/default.cfg
@@ -31,6 +31,20 @@ verify = True
 partition_executable = gpmetis
 
 
+# The io section describes options related to file i/o
+[io]
+
+# the NetCDF file format: NETCDF4, NETCDF4_CLASSIC, NETCDF3_64BIT, or
+# NETCDF3_CLASSIC
+format = NETCDF3_64BIT
+
+# the NetCDF output engine: netcdf4 or scipy
+# the netcdf4 engine is not performing well on Chrysalis and Anvil, so we will
+# try scipy for now.  If we can switch to NETCDF4 format, netcdf4 will be
+# required
+engine = scipy
+
+
 # Options related to deploying a compass conda environment on supported
 # machines
 [deploy]
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/diagnostics_files.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/diagnostics_files.py
@@ -9,6 +9,7 @@ from geometric_features.aggregation import get_aggregator_by_name
 from mpas_tools.logging import check_call
 from mpas_tools.ocean.moc import add_moc_southern_boundary_transects
 from mpas_tools.io import write_netcdf
+import mpas_tools.io
 
 from compass.io import symlink
 from compass.step import Step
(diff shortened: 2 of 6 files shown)


xylar:
## Testing

I ran the nightly ocean test suite successfully with this branch and https://github.com/MPAS-Dev/MPAS-Tools/pull/434 on Ubuntu with gnu and Chrysalis with intel, comparing against a baseline of the current `master`.

I will test SOwISC12to60 tomorrow, which was the mesh giving me trouble in #185

xylar:
@mark-petersen, would you be willing to test this on Grizzly or Badger?  I just want to make sure there aren't any surprises on another machine.  I did a quick test with the `scipy` NetCDF output engine on Compy (outside of `compass`) and it worked fine so I'm pretty sure there will be no surprises there.

xylar:
@matthewhoffman, first, I want to make sure you're okay with these changes.  My testing has indicated that performance for output of NetCDF3 files seems to be better with the `scipy` than the `netcdf4` engine in all cases I have tried (though this difference isn't appreciable for small files).  Where it really matters is for multi-GB files, and Chrysalis was completely choking on these large files when I used the `necdf4` engine yesterday and the day before.  I'm not completely sure if this is 1) and issue with the LCRC file system and/or 2) and issue with how `xarray` is using `netcdf4`.  I haven't tried using the `NetCDF4` python package on its own to read and write these large files.  But given that `scipy` performs really well, I don't feel the need to investigate too much further.

By the way, `netcdf4` performs very well with NetCDF4-format files (as one might expect) but these don't tend to work as expected in MPAS and are not the agreed-upon format for E3SM.  So that's something for the future.

matthewhoffman:
@xylar , is this netcdf output engine thing what xarray is using under the hood?  Or are you somewhere managing netCDF operations manually depending on this option.

It looked like all the changes only really affect the ocean core in COMPASS.  Is that right?  Or is there a reason to think this would affect landice tests?

matthewhoffman (approved):
Tested landice tests and experience no problems.

xylar:
> @xylar , is this netcdf output engine thing what xarray is using under the hood? Or are you somewhere managing netCDF operations manually depending on this option.

That's correct, it is what xarray uses under the hood.

> It looked like all the changes only really affect the ocean core in COMPASS. Is that right? Or is there a reason to think this would affect landice tests?

No, that's not correct.  `landice` tests are also affected.  `mpas_tools.io.write_netcdf()` is a wrapper around `xarray.Dataset.to_netcdf()` that makes sure the files written out are compatible with MPAS/E3SM.  Many `landice` test cases use this function.  Here are a few examples:
https://github.com/MPAS-Dev/compass/blob/864ebf7425d5f076d4c18673b43d102308da7071/compass/landice/tests/eismint2/setup_mesh.py#L47
https://github.com/MPAS-Dev/compass/blob/864ebf7425d5f076d4c18673b43d102308da7071/compass/landice/tests/enthalpy_benchmark/setup_mesh.py#L50
https://github.com/MPAS-Dev/compass/blob/864ebf7425d5f076d4c18673b43d102308da7071/compass/landice/tests/dome/setup_mesh.py#L67

Thank you for testing. It's good to know that landice is also working as expected with this change.

xylar:
@mark-petersen, I'm taking you off this one as a reviewer, too, to not completely overwhelm you with review requests.

PR #184, opened 2021-07-21
Add the capability to use cached meshes and ICs

xylar:
This merge adds a new capability for steps to have cached outputs in a `compass_cache` database.  Files in this database have a directory structure similar to the work directory (but without the MPAS core subdirectory, which is redundant).  The files include a date stamp so that new revisions can be added without removing older ones (supported by older `compass` versions).

As an example, here are cached files for the `mesh` and `init` test cases for the `QU240` mesh:
```
compass_cache/global_ocean/QU240/mesh/mesh/critical_passages_mask_final.210727.nc
compass_cache/global_ocean/QU240/mesh/mesh/culled_graph.210727.info
compass_cache/global_ocean/QU240/mesh/mesh/culled_mesh.210727.nc
compass_cache/global_ocean/QU240/PHC/init/initial_state/init_mode_forcing_data.210727.nc
compass_cache/global_ocean/QU240/PHC/init/initial_state/initial_state.210727.nc
```

The mapping between outputs in the cached versions of steps and the files in the `compass_cache` database are maintained in a file in `compass.<mpas_core>` called `cached_files.json`.  This file contains a python dictionary that maps from the output files in the cached versions of each step (the symlinks) to those in the database (the targets).  For example:

``` json

{
    "ocean/global_ocean/cached/QU240/mesh/mesh/culled_mesh.nc": "global_ocean/QU240/mesh/mesh/culled_mesh.210727.nc",
    "ocean/global_ocean/cached/QU240/mesh/mesh/culled_graph.info": "global_ocean/QU240/mesh/mesh/culled_graph.210727.info",
    "ocean/global_ocean/cached/QU240/mesh/mesh/critical_passages_mask_final.nc": "global_ocean/QU240/mesh/mesh/critical_passages_mask_final.210727.nc",
    "ocean/global_ocean/cached/QU240/PHC/init/initial_state/initial_state.nc": "global_ocean/QU240/PHC/init/initial_state/initial_state.210727.nc",
    "ocean/global_ocean/cached/QU240/PHC/init/initial_state/init_mode_forcing_data.nc": "global_ocean/QU240/PHC/init/initial_state/init_mode_forcing_data.210727.nc"
}
```

A new command, `compass cache` has been added to aid in updating `cached_files.json`.  This command is only available on Anvil and Chrysalis, since you can only copy files from a compass work directory onto the LCRC server from these two machines.  You run `compass cache` from the base work directory, giving the relative paths to the step(s) that you want to cache output files from.  For example:
```
[...]

PR #189, opened 2021-07-30
Add a design document for cached output files

xylar:
This is being implemented in #184

### DIFF
--- a/docs/design_docs/cached_outputs.rst
+++ b/docs/design_docs/cached_outputs.rst
@@ -1,0 +1,570 @@
+.. _design_doc_cached_outputs:
+
+Caching outputs from compass steps
+==================================
+
+Date: 2021/07/30
+
+Contributors: Xylar Asay-Davis
+
+Summary
+-------
+
+We would like to have a way to download output files for ``compass`` steps from
+an online cache instead of generating them each time the step runs.  The
+primary motivation for this is to optionally avoid time-consuming steps for
+generating meshes and initial conditions for faster regression testing with
+MPAS components in "forward" mode.  Potential other uses could include cached
+results as baselines for validation.  A challenge for this capability is
+providing an easy way for both developers and users to control which steps in a
+test case or suite are cached and which are run as normal.
+
+
+Requirements
+------------
+
+.. _req_cached:
+
+Requirement: cached outputs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2021/07/30
+
+Contributors: Xylar Asay-Davis
+
+Each ``compass`` step defines its output files in the ``compass.Step.outputs``
+attribute. For selected steps (see :ref:`req_select`), we require a mechanism
+to download cached files for each of these outputs and to use these cached
+files for the outputs of the step instead of computing them.  
+
+.. _req_select:
+
+Requirement: selecting whether to use cached outputs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2021/07/30
+
+Contributors: Xylar Asay-Davis
+
+There needs to be a mechanism for developers and users to select which steps
+are run as normal and which use cached outputs.  For this mechanism to be
+practical, it should not be overly tedious or manual (e.g. manually setting a 
+flag for each step).
+
+.. _req_update:
+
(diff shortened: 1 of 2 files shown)


xylar:
@mark-petersen, as I mentioned in #184, I think we need a design document for this work.  Here it is.

@matthewhoffman, this work is general enough that it could affect the `landice` core.  If you are not interested in this capability, that's fine. Please just give the design doc a quick glance to make sure it seems to "do no harm" for you.  If you do want to have the capability, please look more closely and make sure it meets the needs you foresee having.

I will make some additional comments or ask questions in specific sections.

xylar:
To render the design doc in sphinx, set up the compass environment as usual.  Hopefully, you can set it up on your laptops because that makes it easier to open the results in a browser.  Once you have sourced the activation script, run:
```
cd docs
make html
```
Then, open `_build/html/design_docs/cached_outputs.html` in your browser.

docs/design_docs/cached_outputs.rst
```diff
@@ -0,0 +1,343 @@
...
+Requirement: either "normal" or "cached" versions of a step
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Date last modified: 2021/07/30
+
+Contributors: Xylar Asay-Davis
+
+We **do not** require the ability to set up a "normal" and a "cached" version
+of the same step within a ``compass`` test case or suite.  (If this is not the
+case, it would place important constraints on the design solution.)
```
xylar, line 95: This is the main requirement where I would like feedback.  Do we foresee needing both "normal" and "cached" version of steps?  If so, how do we avoid having a second copy of every test case where we want to have cached outputs (and every test case that depends on the "cached" test case)?  If we don't need that ability, the design I have here should serve us well.

docs/design_docs/cached_outputs.rst
```diff
@@ -0,0 +1,343 @@
...
+    # Options related to the current test case
+    [test_case]
+    
+    # The steps from the test case where output files should be downloaded
+    # from the server rather than computed
+    cached = mesh
+
+Both of these approaches assume that we always have either the "normal" or the
+"cached" version of a step within a test case or test suite and developers or
+users are free to choose between them, see :ref:`des_normal_or_cached`.
```
xylar, line 154: This is the main design choice where I could use some help brainstorming.  Can either of you think of better, cleaner ways of letting developers (e.g. via test suites) and users (e.g. via config files) determine which steps use cached outputs and which do not?

docs/design_docs/cached_outputs.rst
```diff
@@ -0,0 +1,343 @@
...
+flexibility about which steps are cached and which are not.  It is not clear
+to me how we achieve this flexibility without requiring that a given step
+either be set up as "normal" or "cached", and not both in the same work
+directory.
+
+
+
+
+Implementation
+--------------
```
xylar, line 237: Waiting until we iterate on requirements and design before I rework the implementation in #184

docs/design_docs/cached_outputs.rst
xylar: I came up with a better implementation.  It's reasonably efficient for adding cached outputs to test suites and for setting up test cases where *all* steps are cached.  it's not great for setting up individual test cases where only some of the steps should be cached, and we probably just want to provide test suites if such situations emerge very often.

xylar:
@mark-petersen, I believe this is finished but please feel free to commend and request changes.

mark-petersen (approved):
@xylar thanks, this looks great. I really like the brief notation of 'c' in 
```
compass setup -n 40c
```
and then the ability for a longer format with the `--cached` flag.

I agree that we do not need to require the ability to set up a cached version of every step in a test case. For some domains, like low-resolution regular hexes, it does not make sense to cache the input file.

xylar:
> I agree that we do not need to require the ability to set up a cached version of every step in a test case. For some domains, like low-resolution regular hexes, it does not make sense to cache the input file.

To be clear, the current implementation *does* support caching the outputs of every single step of every test case.  But I haven't actually run all tests and cached their outputs.  So in that sense, we will need to add cache files to the ``compass_cache`` database using the ``compass cache`` tool as we find we need them.

xylar:
Thanks very much for the review!  I think I'll merge this, and then we can make minor tweaks in #184 if need be.

PR #191, opened 2021-07-30
Make `compass setup` create a custom test suite

xylar:
This is convenient for running several test cases in one command even if they aren't part of a predefined test suite.  Test cases are run in the order they are supplied to `compass setup`.

Some functionality has been moved from the `suite` module into `setup` and `clean` to facilitate setting up and cleaning up custom test suites in the same way as predefined ones.

A small bug has been fixed in determining the number of cores that a suite requires.  Previously, all steps were used.  With the fix, only `steps_to_run` are used.

The docs have been updated to include new custom suites.

### DIFF
--- a/compass/clean.py
+++ b/compass/clean.py
@@ -7,7 +7,7 @@ from compass.mpas_cores import get_mpas_cores
 from compass import provenance
 
 
-def clean_cases(tests=None, numbers=None, work_dir=None):
+def clean_cases(tests=None, numbers=None, work_dir=None, suite_name='custom'):
     """
     Set up one or more test cases
 
@@ -21,6 +21,10 @@ def clean_cases(tests=None, numbers=None, work_dir=None):
 
     work_dir : str, optional
         A directory that will serve as the base for creating case directories
+
+    suite_name : str, optional
+        The name of the test suite if tests are being set up through a test
+        suite or ``'custom'`` if not
     """
 
     if tests is None and numbers is None:
@@ -28,6 +32,7 @@ def clean_cases(tests=None, numbers=None, work_dir=None):
 
     if work_dir is None:
         work_dir = os.getcwd()
+    work_dir = os.path.abspath(work_dir)
 
     mpas_cores = get_mpas_cores()
     all_test_cases = dict()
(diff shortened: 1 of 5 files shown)


xylar:
## Testing

With the fixes in #190, I was able to set up a "custom" test suite with all ocean test cases.

I also set up and successfully ran a custom suite with the 6 ISOMIP+ test cases.

xylar:
## Rationale

I believe this will be a helpful step toward resolving an issue with validation that we're having in #164.  We need a way to perform validation between outputs from different test cases, which is not currently supported.  But we need a way to only perform that validation if both steps of both test cases are actually being run.  The clearest way I can see to do this is to make sure both test case are run as part of a test suite.

I have also found it's pretty tedious to define a one-off test suite for temporary work but it's also tedious to `cd` into test cases or steps to run them one after another.  I think this will generally be helpful.

xylar:
@matthewhoffman, @mark-petersen and @cbegeman, first please let me know if you think this is a good idea.  Second, please give it a quick whirl when you can.  It should be very quick to test.  Let me know how it goes.

xylar:
I don't think this work is related to #164 after all (it isn't yet clear to me how to tell if a step within a test suite got run or not other than if the output exist, and the output from validation gets kind of buried when running a suite) but I think it will still be useful in general.

mark-petersen (approved):
Yes! This is a wonderful addition! I tried it with:
```
compass setup -p $PATH -w $WORKDIR -n 40 41 42
cd $WORKDIR
compass run  # (on a compute node)
```
and it works!  It runs just those three cases! This is like getting our first microwave when I was ten - you didn't know you wanted it before, and now it's indispensable!

xylar:
@mark-petersen, thanks!  I've been doing work in other branches and have been really tempted to rebase them onto this branch because it's really annoying not to have this feature.  I'm glad you agree.

mark-petersen:
@matthewhoffman and @cbegeman if you are comfortable approving based on our testing and comments, please go ahead. It would be nice to merge this in soon.

cbegeman:
Thanks for the reminder, @xylar. I tested it and it works fine for me as well. I personally don't find it cumbersome to add `compass/ocean/suites/custom-suite.txt`, but I don't see any harm in adding another option. And I like how you've tucked the minimum cores determination in `setup_cases`.

cbegeman (approved):

matthewhoffman (approved):
I looked over the changes and don't see anything that concerns me.  I can see how this can be a useful feature.  Given the other testing reported by others, I'm comfortable with this being merged without me doing more.

xylar:
Thanks @cbegeman and @matthewhoffman for your reviews!

PR #195, opened 2021-08-05
Remove f-strings in config script

xylar:
We want to be compatible with earlier versions of python in this particular script.


xylar:
This caused trouble on `compy`.

PR #198, opened 2021-08-06
Check git username and email during setup

xylar:
When setting up global ocean test cases, check for a git username and email address.  This way, any issues are caught during setup rather than showing up as relatively hidden errors at runtime.

### DIFF
--- a/compass/ocean/tests/global_ocean/configure.py
+++ b/compass/ocean/tests/global_ocean/configure.py
@@ -1,4 +1,6 @@
 from compass.config import add_config
+from compass.ocean.tests.global_ocean.metadata import \
+    get_author_and_email_from_git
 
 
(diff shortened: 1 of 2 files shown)


xylar:
This is following up on #196.  @katsmith133 is working on Compy and didn't happen to have git configured yet.  It would have been nice if this situation was pointed out during setup where the error message is relatively obvious rather than at runtime, where it got buried among several unrelated warnings.

xylar:
## Testing,

I set up the ocean nightly test suite with this branch.  When I don't explicitly set the user name and email address, they are picked up from git during setup, as expected, and show up in the config file.

xylar:
One of the checks seems to have stalled.  I have no reason to think it won't pass since the other 5 did.

PR #201, opened 2021-08-08
Make conda path dummy name more verbose

xylar:
This merge makes the dummy path name for the `--conda` flag in the deployment script more verbose: `conda_path` --> `base_path_to_install_or_update_conda`.  The hope is that this makes clear what is stated in the text immediately below, that the path is not just for an existing Miniconda3 installation but also for a new installation if Miniconda3 is not yet installed.


xylar:
@pbosler, would this change have helped with the confusion you had?

PR #206, opened 2021-08-12
Global ocean forward steps explicitly add output.nc

### LINKED ISSUE #205: Remove output.nc from expected testcase output
In Spinning up the WC14 with 80 layers I ran into an issue in every stage of the dynamic_adjustment stage looking for `output.nc`.  The test ran fun but the output file was not generated (as expected from streams.ocean) resulting in the error.  This required running each stage individually.  It would be helpful to remove that file from the expected output.

xylar:
Rather than assuming that every forward step has an output called `output.nc`, this merge adds that output explicitly when it is needed (which is true for most steps).  The only steps that currently don't produce this output are the shorter dynamic adjustment steps.

closes #205

### DIFF
--- a/compass/ocean/tests/global_ocean/mesh/wc14/dynamic_adjustment/__init__.py
+++ b/compass/ocean/tests/global_ocean/mesh/wc14/dynamic_adjustment/__init__.py
@@ -221,6 +221,7 @@ class WC14DynamicAdjustment(DynamicAdjustment):
 
         step.add_input_file(filename='../{}'.format(restart_filenames[5]))
         step.add_output_file(filename='../{}'.format(restart_filenames[6]))
+        step.add_output_file(filename='output.nc')
         self.add_step(step)
 
         self.restart_filenames = restart_filenames
--- a/compass/ocean/tests/global_ocean/daily_output_test/__init__.py
+++ b/compass/ocean/tests/global_ocean/daily_output_test/__init__.py
@@ -37,6 +37,7 @@ class DailyOutputTest(ForwardTestCase):
                            threads=1)
 
         module = self.__module__
+        step.add_output_file(filename='output.nc')
         step.add_namelist_file(module, 'namelist.forward')
         step.add_streams_file(module, 'streams.forward')
         self.add_step(step)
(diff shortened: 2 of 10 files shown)


xylar:
## Testing

I ran the ocean nightly test suite on Chrysalis with intel and intel-MPI, and it was bit-for-bit with a recent baseline run.

I was able to successfully run the dynamic adjustment test case for the QU240, QUwISC240 and WC14 run (though the WC14 timed out in the last 1.5 days of simulation, so I'm only 99% sure it works).  No errors were produced related to `output.nc` files.

However, I wasn't able to fully test with EC30to60 or SO12to60 meshes because of the spin-up issues I reported in #208.

compass/ocean/tests/global_ocean/mesh/wc14/dynamic_adjustment/__init__.py
```diff
@@ -221,6 +221,7 @@ def __init__(self, test_group, mesh, init, time_integrator):
 
         step.add_input_file(filename='../{}'.format(restart_filenames[5]))
         step.add_output_file(filename='../{}'.format(restart_filenames[6]))
+        step.add_output_file(filename='output.nc')
```
vanroekel, line 224: @xylar, I'm sorry if I'm being dense here, but it seems like you are adding the output.nc file to the expected  output?  I thought this was my problem in the WC14 spin up that it didn't actually create this file.  It seems it will now be expected?  I feel like i'm missing something obvious so sorry for the stupid question.

xylar: I can definitely understand the confusion.  In the `simulation` step, `output.nc` is always produced and we need this for validation.  In other dynamic-adjustment steps, it is not.

vanroekel: Oh, I see, I didn't notice this was just the `simulation` step.  I now see that the file above added `output.nc` to every step for global forward now it is explicit only for simulation.  Thanks for the help understanding this!

vanroekel (approved):
approved by visual inspection and testing on chrysalis.  Thanks @xylar!

xylar:
Thanks once again, @vanroekel!

PR #212, opened 2021-08-13
Remove mixed layer depth AM from global ocean steps

### LINKED ISSUE #211: Mixed layer depth analysis member in global ocean test cases outputs every time step
It appears that the MLD analysis member was turned on in legacy compass in this PR https://github.com/MPAS-Dev/MPAS-Model/pull/445 and set to output results every time step.  I presume this was a debugging step for that particular PR that got committed by mistake.  We inherited this in the current `compass`:
https://github.com/MPAS-Dev/compass/blob/36166e7f08c65f170e93abe11c91d8a225b802f5/compass/ocean/tests/global_ocean/namelist.forward#L19-L20

This creates some enormous output files an likely slows down the performance in `compass` fairly substantially.

xylar:
The mixed layer depth analysis member was enabled and writing out analysis at every time step for all global ocean forward model runs.  This appears to have been turned on for debugging in a pull request from more than a year ago (https://github.com/MPAS-Dev/MPAS-Model/pull/445).  During dynamic adjustment, this can lead to files that are many GB in size and presumably slows down all model runs significantly.

closes #211


xylar:
Okay, so clearly I misunderstood.  The MLD AM isn't optional if you want to use Redi.  (That seems confusing, since my understanding was that AMs are always supposed to be optional and if the MLD AM isn't optional anymore, it shouldn't be an AM anymore, but that's another discussion for another time.)

So it seems like the issue is more that we probably don't want the MLD AM to be writing output from every time step (except in the `analysis_test` test case).

xylar:
## Testing

The nightly test suite passed on Chrysalis with intel and impi, using a baseline from 2021/08/03.

The logs indicate that MLD output wasn't taking a significant fraction of the runtime so this change probably doesn't impact performance of the nightly suite very much but should improve longer runs like dynamic adjustment.

vanroekel:
@xylar sorry for that confusion, that was my change.  There is some tapering required for Redi and potentially soon GM that requires the MLD to be computed every time step.  I agree it is really now no longer "analysis".  It would seem reasonable to move this to part of diagnostics or at least just in shared to alleviate the confusion and would happy to put that on my list.

But I think the change to remove the output makes great sense for tests and will approve the PR.

vanroekel (approved):
approved by visual inspection

xylar:
@vanroekel, thanks for the review.  While I think the clean up you're mentioning would be a good idea, it's certainly a super low priority.  The solution you have is perfectly workable even if it did confuse me a bit.

xylar:
@mark-petersen, could you take a look at this when you have a chance?

PR #213, opened 2021-08-13
Fix timer comparison with baseline

xylar:
The method for finding the baseline path for the test case has been updated.  It should come from the `test_case` object, not from a config option.

The API of the `compare_timers()` function needed to change to accommodate change to where the baseline directory is stored.  The 3 test cases that use timers have been updated with the new API.


xylar:
## Testing

I ran the nightly ocean test suite on Chrysalis with intel and impi, using a run from 2021/08/03 as a baseline.  I verified that the expected timer output is in the log files for the following test cases:
* `ocean/global_ocean/QU240/PHC/performance_test`
* `ocean/global_ocean/QU240/PHC/RK4/performance_test`
* `ocean/global_ocean/QU240/PHC/analysis_test`
* `ocean/global_ocean/QU240/EN4_1900/performance_test`
* `ocean/global_ocean/QU240/PHC_BGC/performance_test`
* `ocean/global_ocean/QUwISC240/PHC/performance_test`
* `ocean/ziso/20km/default`

PR #214, opened 2021-08-13
Remove EN4_1900 and PHC_BGC test from nightly suite

xylar:
As discussed in https://github.com/MPAS-Dev/compass/discussions/188, we do not plan to use these initial conditions in E3SM runs so it doesn't make sense to test them on a regular basis.


xylar:
@maltrud, @mark-petersen, @sbrus89 and @vanroekel, I'm assuming you gave the feedback you wanted to give in #188 and there's no need to review.  Just letting you know that I'm removing these tests from `nightly`.

xylar:
## Testing

The reduced nightly test suite passed on Chrysalis with intel and impi, using a baseline from 2021/08/03.  The total runtime was about 7:30, as compared to about 9:00 for other recent runs.

PR #216, opened 2021-08-13
Add a new ocean "pr" test suite

xylar:
This test suite is heavier than the `nightly` suite and it intended to be run to test E3SM PRs that change MPAS-Ocean to more robustly test for unexpected changes.

See some discussion here https://github.com/MPAS-Dev/compass/discussions/188#discussioncomment-1086525

PR #217, opened 2021-08-13
Print steps of test cases in a suite

xylar:
This merge adds output indicating which step is currently running in a test suite.  This is helpful for test cases made up of many steps, particularly if each takes a long time.

Typical output looks like:
```
  * step: initial_state
  * step: forward
  test execution:      SUCCESS
  * step: initial_state
  * step: 1thread
  * step: 2thread
  test execution:      SUCCESS
  test validation:     PASS
```


xylar:
@mark-petersen and @matthewhoffman, I don't need a full review of this.  But I would like to know if you think this is an improvement or not.  I can make it off by default if you prefer, and turn it on with a flag like:
```
compass run -v
```
where `-v` is for "verbose".

xylar:
## Testing

I tested this with the ocean nightly test suite on my laptop (with gnu) and with the `pr` test suite from #216 on Chrysalis with intel/impi.

I also tested a test case and a step on their own.  The formatting for a step on its own is the same as before, but a test case on its own is slightly different:
```
performance_test$ compass run
Running steps: forward
  * step: forward
```
`* Running forward` has been changed to `* step: forward` because this makes things clearer in the test suite.  I'm no longer having it print "Complete" when a test succeeds because this seemed unnecessarily verbose in the test suite version.

matthewhoffman (approved):
Xylar, this looks helpful.  I think it should be the default.

xylar:
@mark-petersen, any feedback?

xylar:
Thanks @matthewhoffman!

mark-petersen:
Yes, I like having this as an option for larger meshes. I tested the nightly regressions suite, and would prefer for it to not be verbose for a long list of short tests. I would recommend adding either:
```
-v verbose
or
-q quiet
```
depending on what you prefer as the default. I guess `-v verbose` is a more standard option. In the thesaurus, the opposite of verbose is `succinct` and `laconic`, both of which sound a bit odd.

xylar:
Thanks @mark-petersen.  It sounds like @matthewhoffman and I both prefer that "verbose" be the default.  I would be happy to add a `-q`, `--quite` flag.  I think `-v` is more commonly used to display the version number, particular for python command-line tools where this is a default capability of the `argparse` package.  But that's only a secondary argument to support my preference for "verbose" to be the default.

mark-petersen:
I would be happy with the `-q, --quiet` (watch the spelling) and use that when needed. Yes, you are right about `-v --version`. Thanks!

xylar:
> watch the spelling

Quite so, sorry about that!

xylar:
@mark-petersen, I added the `-q` or `--quiet` flag to `compass run` as you requested.  Currently, it has no effect if you are running a test case or step on its own. (We could think about whether there is a useful "quiet" behavior for those situations, too.)

Please give this a quick test and let me know if that addresses your concern.

PR #219, opened 2021-08-13
A bit of cleanup to the docs

xylar:
1. State that only `bash` is supported in the developer's guide. We don't create activation scripts for `csh` or equivalent.
2. Remove `json` code tags in a design doc.  Sphinx doesn't recognize them.
3. Make section symbols consistent in framework section of the developer's guide.  Recent changes used tildes (`~`) instead of carrots (`^`).  Now using tildes for all subsections.

### DIFF
--- a/docs/developers_guide/quick_start.rst
+++ b/docs/developers_guide/quick_start.rst
@@ -3,6 +3,18 @@
 Quick Start for Developers
 ==========================
 
+.. _dev_shell:
+
+Unix Shell
+----------
+
+Currently, compass only supports ``bash`` and related unix shells (such as
+``ksh`` on the Mac).  We do not support ``csh``, ``tcsh`` or other variants of
+``csh``.  An activation script for those shells will not be created.
+
+If you normally use ``csh``, ``tcsh`` or similar, you will need to temporarily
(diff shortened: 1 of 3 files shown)


docs/developers_guide/quick_start.rst
```diff
@@ -3,6 +3,18 @@
...
+Unix Shell
+----------
+
+Currently, compass only supports ``bash`` and related unix shells (such as
+``ksh`` on the Mac).  We do not support ``csh``, ``tcsh`` or other variants of
+``csh``.  An activation script for those shells will not be created.
+
+If you normally use ``csh``, ``tcsh`` or similar, you will need to temporarily
+switch to bash by calling ``/bin/bash`` each time you want to use compass.
+
```
xylar, line 17: @milenaveneziani, feel free to have a look at this and make any suggestions.  I won't make you review it formally in case you don't have time.  I'll merge this in about half an hour so if you don't happen to have a chance to look at it, no worries.

milenaveneziani: sorry, I was away for a few hours. But looks good to me. Thanks!

PR #221, opened 2021-08-15
Make `conda_base` an absolute path

xylar:
In the script for setting up the compass conda environment, make sure the `conda_base` path (passed in from `--conda` or supplied as a config option) is an absolute path.  Recent tests showed that a relative path resulted in failures when trying to activate the environment.

### DIFF
--- a/conda/configure_compass_env.py
+++ b/conda/configure_compass_env.py
@@ -112,9 +112,8 @@ def get_conda_base(conda_base, is_test, config):
                                  'none could be inferred.')
         else:
             conda_base = config.get('paths', 'compass_envs')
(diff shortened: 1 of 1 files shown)


xylar:
@milenaveneziani, I think this should take care of the issue you saw recently.  If you want to test after I merge this, feel free.  But no need to review.

xylar:
## Testing

I tested this on my Linux laptop, supplying a relative path to `--conda`:
```
./conda/configure_compass_env.py --conda ../../../Desktop/conda_compass --mpi=mpich
```
I had some unrelated trouble with downloading the `pnetcdf` and `scorpio` packages (it took 3 tries) but the relative path worked fine.

PR #222, opened 2021-08-15
Fix dynamic adjustment test case for SOwISC12to60 mesh

### LINKED ISSUE #208: EC30to60, ECwISC30to60 and SOwISC12to60 dynamic adjustment test cases failing on Chrysalis
The `damped_adjustment_*` steps run successfully but there are NaNs in state validation during the `simulation` step.

xylar:
This merge
* uses the same namelist options for SOwISC12to60 as in E3SM 
* reduces the Rayleigh damping in the damped steps
* reduces the barotropic time step to match E3SM
* extends the duration of the existing damped steps
* adds an additional step that runs without damping but with the reduced (5-minute) time step for 10 days

The total adjustment time is extended from 20 to 50 days.  This appear to be necessary to get through some periods of high wave activity during the first 30 to 40 days.

Why exactly the dynamic adjustment that worked before no longer runs successfully is not clear.  It could be the new JIGSAW version, which generates a different mesh and could therefore change the interaction of waves with ice-shelf cavities, but that's pure speculation.

closes #208

### DIFF
--- a/compass/ocean/tests/global_ocean/mesh/so12to60/dynamic_adjustment/__init__.py
+++ b/compass/ocean/tests/global_ocean/mesh/so12to60/dynamic_adjustment/__init__.py
@@ -37,8 +37,8 @@ class SO12to60DynamicAdjustment(DynamicAdjustment):
                 mesh.mesh_name, time_integrator))
 
         restart_times = ['0001-01-03_00:00:00', '0001-01-11_00:00:00',
-                         '0001-01-21_00:00:00', '0001-01-31_00:00:00',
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I successfully ran dynamic adjustment on Chrysalis with intel/impi and cached versions of the `mesh` and `PHC/init` test cases.  During the `validate()` method for the test case, I saw errors related to a missing `simulation/output.nc` file, but I ran a separate test (not requiring rerunning the full spin-up) to verify that this has been fixed.

PR #225, opened 2021-08-16
Change file permissions after downloading files

### LINKED ISSUE #223: Shared files are ending up with permissions that don't allow collaborators to read/write
When @mark-petersen downloaded cached output files on Grizzly, both the files and directories ended up in his group, not the shared `climate` group.  We need a `chown/chmod` step for files downloaded to the shared space on supported machines.

xylar:
This is part of setting up compass steps that include downloading files from a database to a local cache on a supported machine.

This merge also moves the `group` config option to a new `permissions` section of the config file. On supported machines, we need group permissions not only during deployment of the compass conda environment but also when downloading files to a shared cache of the databases on the LCRC server.

Finally, this merge fixes a bug in how file permissions are changed during creation of a conda environments and building of system libraries (ESMF and SCORPIO).

closes #223


xylar:
## Testing

I deleted the `compass_cache` directory on LANL IC and then set up the nightly test suite.  The files and directories were downloaded and given the expected permissions: user and group read and write; world read; execute permission on directories.

PR #226, opened 2021-08-17
Avoid redundant activation script name

xylar:
If the environment name is given manually and it includes the mpi type, it should not redundantly be added to the activation script.


xylar:
## Testing

I created a `test_mpich` environment and got `load_test_mpich.sh`, not `load_test_mpich_mpich.sh`.

PR #227, opened 2021-08-17
Fix cores, min_cores, threads from config options

xylar:
In `global_ocean` `decomp_test` and `threads_test`, the number of cores and threads were incorrectly being retrieved from the config options rather than using the hard-coded values.

### DIFF
--- a/compass/ocean/tests/global_ocean/forward.py
+++ b/compass/ocean/tests/global_ocean/forward.py
@@ -24,6 +24,15 @@ class ForwardStep(Step):
 
     time_integrator : {'split_explicit', 'RK4'}
         The time integrator to use for the forward run
+
+    cores_from_config : bool
+        Whether to get ``cores`` from the config file
+
+    min_cores_from_config : bool
+        Whether to get ``min_cores`` from the config file
+
+    threads_from_config : bool
+        Whether to get ``threads`` from the config file
     """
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

This time, I'm seeing the expected number of cores in each step of the `decomp_test` test case:
```
$ ls */graph.info.part*
4proc/graph.info.part.4  8proc/graph.info.part.8
```
Fortunately, the results are still bit-for-bit with the fix in place on Ubuntu with gnu.

compass/ocean/tests/global_ocean/forward.py
```diff
@@ -195,13 +208,18 @@ def run(self):
         Run each step of the testcase
         """
         config = self.config
-        # get the these properties from the config options
         for step_name in self.steps_to_run:
             step = self.steps[step_name]
-            # get the these properties from the config options
-            step.cores = config.getint('global_ocean', 'forward_cores')
-            step.min_cores = config.getint('global_ocean', 'forward_min_cores')
-            step.threads = config.getint('global_ocean', 'forward_threads')
```
xylar, line 204: Previously, this was replacing the core and thread counts for all test cases with the values from the config file.  This meant that both steps of the decomp test case were running on 4 cores, and both steps of the thread test were running with 1 thread.

xylar:
@mark-petersen and @vanroekel, making you aware of this embarrassing bug and that I have now fixed it.

PR #232, opened 2021-08-28
Add support for tcsh/csh

### LINKED ISSUE #228: Consider supporting `csh` and not just `bash`
Currently, `compass` supports only `bash` and similar shells, not `csh` and its variants.  There has been some interest in support for `csh`.  Developers would like this should comment below and, if there is sufficient demand, I will add it.

xylar:
The conda/configure_compass_env.py script now detects what your preferred shell is via `$SHELL` and uses this to set up the compass environment and activation script either with `sh` or `csh`.

closes #228


xylar:
## Testing

I an environment successfully in `bash` as usual on Ubuntu and updated one on Chrysalis.

I artificially switched to `tcsh` by running:
```
/bin/tcsh
setenv SHELL /bin/tcsh
```
to simulate what would happen for someone using `/bin/tcsh` as their default shell.  Again, I was able to create an environment successfully under Ubuntu.

xylar:
@milenaveneziani, could you point Prajvala to this PR, too?  I don't have her GitHub username.

milenaveneziani:
I am not sure if she is on github actually, but I'll ask. I had actually suggested she should switch to bash on grizzly.

milenaveneziani:
Pinging @prajvala01123 here. FYI, she has also switched to bash on grizzly.

xylar:
Closing for now because @prajvala01123 is comfortable using bash.

PR #234, opened 2021-09-09
Check compass run for compute node

xylar:
Add a check to all types of calls to `compass run` (suites, test cases and steps) to make sure that, on machines with Slurm, we are running on a compute node (i.e. `$SLURM_JOB_ID` is defined).

### DIFF
--- a/compass/parallel.py
+++ b/compass/parallel.py
@@ -39,6 +39,36 @@ def get_available_cores_and_nodes(config):
     return cores, nodes
 
 
+def check_parallel_system(config):
+    """
+    Check whether we are in an appropriate state for the given queuing system.
+    For systems with Slurm, this means that we need to have an interactive
+    or batch job on a compute node, as determined by the ``$SLURM_JOB_ID``
+    environment variable.
+
+    Parameters
+    ----------
+    config : configparser.ConfigParser
+        Configuration options for the test case
+
+    Raises
+    -------
+    ValueError
+        If using Slurm and not on a compute node
+    """
+
+    parallel_system = config.get('parallel', 'system')
+    if parallel_system == 'slurm':
+        if 'SLURM_JOB_ID' not in os.environ:
+            raise ValueError('SLURM_JOB_ID not defined.  You are likely not '
+                             'on a compute node.')
+    elif parallel_system == 'single_node':
+        pass
+    else:
+        raise ValueError('Unexpected parallel system: {}'.format(
+            parallel_system))
+
+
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I ran the `nightly` test suite as well as a single test case and step from that suite on a Chrysalis login node and got the expected error message in each case:
```
SLURM_JOB_ID not defined.  You are likely not on a compute node.
```

Then, I ran each of these (the full suite, a test case and a step) on a compute node and they each complete as expected without errors.

xylar:
@matthewhoffman and @mark-petersen, this should be a really quick one to review.  I was getting annoyed that I would sometime forget to run the command to get an interactive node on Chrysalis.  Nearly all the test cases in a test suite would fail but I would need to open up the log files to find out that it was because I was on a login node.  It seems like it will be much more helpful to just be told you're not running in the right place from the very beginning.

matthewhoffman (approved):
@xylar , I looked this over and the solution makes sense.  This looks like a useful check.

mark-petersen (approved):
Thanks, this is useful. I sometimes hit `compass run` on the wrong tab. I tested on grizzly with a suite and a case. I get:
```
compass run
  File "/users/mpeterse/miconda3/envs/dev_compass_1.0.0/bin/compass", line 33, in <module>
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/__main__.py", line 62, in main
    commands[args.command]()
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/run.py", line 312, in main
    run_suite(suite, quiet=args.quiet)
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/run.py", line 55, in run_suite
    check_parallel_system(config)
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/parallel.py", line 63, in check_parallel_system
    raise ValueError('SLURM_JOB_ID not defined.  You are likely not '
ValueError: SLURM_JOB_ID not defined.  You are likely not on a compute node.
```
The traceback seems unnecessary, but I looked up `ValueError` and it appears to just come along for the ride.

xylar:
Thanks @mark-petersen 
> The traceback seems unnecessary

That's a standard behavior or python that definitely takes getting used to but raising an error always produces a traceback.  It's not really worth overriding that behavior and you will come to expect it.

xylar:
@matthewhoffman, thank you for your review as well!

PR #238, opened 2021-09-16
Add a planar convergence test group

xylar:
This merge includes an horizontal advection test case on planar, doubly periodic meshes.  It should be a useful complement to the `global_convergence` test group (notably the cosine bell test case) and should also be useful for debugging issues with the mesh converter related to periodic, planar meshes.

This merge also adds a method to the `Step` class for updating streams files at runtime. This is similar to the approach for updating namelist files, and can be used when config options that users might edit could affect things like output interval.


xylar:
This test group still needs to be documented and I likely will want modify the resolutions used, since the highest resolution (1 km) is too expensive.  I may also try some intermediate resolutions to see if they work correctly.

xylar:
## Testing

I tested with 1 km to 32 km resolution on Chrysalis and the test works as expected.

convergence

I do think 1 km is too expensive (requiring 3 nodes).  The whole test take 7:21.  I will re-test with fewer nodes and 2 km to 64 km resolution.

xylar:
## Further testing

I rebased this PR and ran the `pr` test suite as well as this test group.  The new convergence plot looks like this:

convergence

I set a lower threshold on the convergence rate of 1.9 and an upper bound (beyond which you get a warning) of 2.3.

xylar:
@vanroekel, there is not a big rush on this but it is ready to review when you can.  It is documented (if fairly minimally) and tested on Chrysalis.  It is based off of #250 so I will rebase as soon as that is merged.

xylar:
@vanroekel, just a follow-up to say that #250 has been merged and this branch has been rebased.  Again, no rush.  If this is something you don't have time to review, just let me know.  It's clearly not critically important.  I just figured:
1) it wouldn't hurt to have this in addition to the spherical tests and
2) it could help to debug the issue with the mesh converter on planar meshes, as I have mentioned.

vanroekel:
@xylar I took a look through the code and it looks great to me.  Would you like me to test as well?  Or are you okay with me approving based on your testing?

xylar:
@vanroekel, I don't think you need to test.  Approval by inspection is good enough.  Thank you!

vanroekel (approved):
Approving based on visual inspection and testing from @xylar 

This looks great.  Thanks for adding this test!

xylar:
Thanks @vanroekel!

PR #239, opened 2021-09-16
Add config options for E3SM supported machines with mache

xylar:
This merge introduces the new [mache](https://github.com/E3SM-Project/mache) package as a dependency for both setting up conda environments and running compass.  `mache` keeps track of paths, modules, compilers, MPI libraries, parallel environment information (such as number of cores per node) and more so we don't have to duplicate this info across packages.

In order to add `mache`, it was useful to change the conda build approach so a small, temporary conda environment is created (including `mache` as one of its packages) and then the temporary environment is used to create the `compass` environment.  The temporary environment is deleted at the end as long as the compass environment is created successfully.

`conda/configure_compass_env.py` has been broken into 3 python modules (the new ones are `conda/shared.py` and `conda/bootstrap.py`) to make the 2-step process for creating a conda environment easier.  Developers will not need to know about these changes, and the end result is the same as before except for the inclusion of the new `mache` package.


xylar:
## Testing

I ran the `pr` test suite with inte/impi on Chrysalis, using the current `master` as a baseline.  All tests are bit-for-bit.

xylar:
@matthewhoffman and @mark-petersen, as we discussed, what I'm hoping for in a review of this PR is just a sanity, no-harm check where you setup a development environment (perhaps using `--recreate`), build the model (from the submodule or elsewhere), and setup and run a test suite.  If you can do that by the end of the week, that would be ideal.

mark-petersen (approved):
This works for me. Tested on grizzly with gnu, ran `pr` suite with comparison to previous run.

matthewhoffman (approved):
I took a close look over these changes and the new mache package to get a general sense of how they integrate.  I'm approving by inspection and based on the testing that @xylar  and @mark-petersen have already done.

xylar:
@matthewhoffman and @mark-petersen, thank you both for getting to this so quickly after our meeting today.  That's very helpful.

PR #240, opened 2021-09-22
Update to libnetcdf 4.8.1

xylar:
Packages on conda-forge are migrating to this version so compass should, too.


xylar:
## Testing

I tested this on Chrysalis with intel/impi and on Ubuntu with gnu and both mpich and openmpi.  All tests were successful and bit-for-bit with a baseline with the previous libnetcdf.

I didn't have time to test on OSX but I have uploaded updated packages for pnetcdf and scorpio, both built with both mpich and openmpi and on linux and osx.  So I think we're good to go.

PR #244, opened 2021-09-23
Update legacy compass environment

xylar:
We should update the legacy environment once `mpas_tools 0.11.0` is released, including dropping `pyflann` and instead using `scipy.spatial.KDTree`.

We still need to support legacy for now because coastal tests have not been ported to the new package.

### DIFF
--- a/deploy/create_compass_env.py
+++ b/deploy/create_compass_env.py
@@ -15,8 +15,8 @@ def get_envs():
     # to use.
 
     envs = [{'suffix': '',
-             'version': '0.2.0',
-             'python': '3.8',
+             'version': '0.3.0',
+             'python': '3.9',
              'mpi': 'nompi'}]
 
     # whether to delete and rebuild each environment if it already exists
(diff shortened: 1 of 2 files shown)


xylar:
@sbrus89, I'd like this to be the last time I make a new legacy compass environment.  I'm willing to put in some fairly minimal effort  to make sure legacy is still usable for awhile but that would involve users creating their own environments as needed going forward.

So maybe this is a little push to start porting test cases to the new system.

sbrus89:
@xylar, sure. That's perfectly reasonable.

sbrus89 (approved):
Approved based on visual inspection.

xylar:
@sbrus89, thanks for understanding.  I'll take this out of draft and remove the `in-progress` tag as soon as I have created the legacy environments on Compy, Cori, Anvil/Chrysalis and Grizzly/Badger

xylar:
## Environments

- [ ] Compy
- [ ] Cori
- [ ] Anvil/Chrysalis
- [ ] Grizzly/Badger

PR #245, opened 2021-09-26
Add `add_config` method to test cases

xylar:
This new method is used to add a config file during test case
init in the same way that input, output, namelist and streams files
can be added to steps during their init.

This is useful because some config options should be computed in
the `configure()` method based on other config options, and it
may be helpful if (other than these computed config options) all
config options are loaded from files before `configure()` is
called.

Additionally, there is a list of dictionaries that keeps track of
the config files that are combined to get the config options for
a test case.  This may be useful in the future, e.g. for provenance.

Test-case setup has been modified to first call the `add_config`
method to build up a list of config files before loading them
more or less as usual.  The main difference is that all config
files are now loaded before calls to `configure()`, with the user
config file being added last.  This makes it more likely that
config options are as close as possible to their final state
before `configure()` is called.

### DIFF
--- a/docs/developers_guide/framework.rst
+++ b/docs/developers_guide/framework.rst
@@ -117,60 +117,58 @@ and debugging.
 Config files
 ------------
 
-The ``compass.config`` module includes functions for creating and manipulating
-config options and :ref:`config_files`.
+You can add a config file within an MPAS core with the same name as the MPAS
+core plus a ``.cfg`` extension (e.g. ``ocean.cfg``).  This file will
+automatically get loaded during setup.  Similarly, you can add config files
+within test groups or test cases with the same naming convention (e.g.
+``global_ocean.cfg`` for the ``global_ocean`` test group or ``cosine_bell.cfg``
+for the ``cosine_bell`` test case).  Each of these config files will be loaded
+automatically during setup without you needing to add them explicitly.
 
+If you need to add additional config files that serve other purposes, you can
+use the test case's In a test case's :py:meth:`compass.TestCase.add_config()`
+method within the test case's constructor (``__init__()`` method).  For
+example, a slightly simplified version of the
+:py:class:`compass.ocean.tests.global_ocean.mesh.Mesh` test case might load
+a config file specific to the global ocean mesh as follows:
+
+.. code-block:: python
+
+    def __init__(self, test_group, mesh_name):
+        ...
+        self.add_config(package=mesh_step.package,
+                        filename=mesh_step.mesh_config_filename,
+                        exception=True)
+        ...
+
+The ``package`` and ``filename`` arguments are the name of a package containing
+the config file and the name of the config file itself, respectively.  In this
+case, we know that the config file should always exist, so we would like the
+code to raise an exception (``exception=True``) if the file is not found.  This
+is the default behavior.  In some cases, you would like the code to add the
+config options if the config file exists and do nothing if it does not, in
+which case you should pass ``exception=False``.
+
+There is also a framework-level module, ``compass.config`` that includes
+functions for creating and manipulating config options and :ref:`config_files`.
 
 The :py:func:`compass.config.add_config()` function can be used to add the
 contents of a config file within a package to the current config parser.
-Examples of this can be found in most test cases as well as
-:py:func:`compass.setup.setup_case()`. Here is a typical example from
-:py:func:`compass.landice.tests.enthalpy_benchmark.A.A.configure()`:
+An example of this is the
+:py:class:`compass.ocean.tests.global_convergence.cosine_bell.CosineBell` test
+case:
 
 .. code-block:: python
 
-    def configure(self):
-        """
-        Modify the configuration options for this test case
-        """
-        add_config(self.config, 'compass.landice.tests.enthalpy_benchmark.A',
-                   'A.cfg', exception=True)
+    def __init__(self, test_group):
         ...
+        config = configparser.ConfigParser(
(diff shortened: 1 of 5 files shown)


xylar:
@mark-petersen and @matthewhoffman, the need for this change came up when I was trying to compute the number of cores (and min cores) as a function of mesh resolution for a global ocean test case.  The idea is that I want to add a quasi-uniform mesh with unspecified resolution.  Someone can specify the resolution in a user config file, and then other properties of the test case (like the number of cores and the time step) should follow from that choice.

The problem I ran into was that, before calling the `configure()` method of the test case, not all the config options had been loaded from config files.  Config options related to the mesh were loaded within `configure()`, *after* the user config file.  This meant that the user specified the resolution of the mesh, but then the mesh config file overrode that choice with a default value within `configure()` before I had a chance to compute the number of cores and time step.

I decided that the cleanest solution might be an approach for config files similar to what I have done for input, output, namelist and streams files in steps.  These can be "added" during step initialization but they are actually just added to a list of files to be added later during setup.  Similarly, this merge supports adding config files during init to a list that will be loaded during setup.  This way, during init I can specify that the mesh config file should be loaded.  Then, during setup, I can load all of the config files in the desired order, such that each config file overrides config options from the previous one: default, machine, MPAS-core, test group, test case, those specified at init, and finally the user config file.  With this change, I get the desired behavior.

I will make a draft PR for the QU mesh soon, and perhaps that will help to illustrate the problem.

xylar:
@matthewhoffman and @mark-petersen, given my explanation and the testing I am working on, all I would want from you in terms of a review is your consent to these changes (and perhaps some do-no-harm MALI testing).  Feel free to look over the code but don't feel obligated.

xylar:
## Testing

I tested this with the nightly test suite on Ubuntu with gnu/mpich and with the pr test suite on Chrysalis with intel/impi.  In both cases, all tests were bit-for-bit with a recent baseline.

mark-petersen:
I ran a quick test on this, and get:
```
compass suite -f Config_dir/conservation-am -s -c ocean -t pr -b $n/ocean_model_210910_aa70c750_gr_gfortran_openmp_master_openacc-diagnostics-new_double -w $n/ocean_model_210923_fe6876f1_gr_gfortran_openmp_conservation-am_t
Setting up test cases:
  File "/users/mpeterse/miconda3/envs/dev_compass_1.0.0/bin/compass", line 33, in <module>
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/__main__.py", line 62, in main
    commands[args.command]()
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/suite.py", line 124, in main
    setup_suite(mpas_core=args.core, suite_name=args.test_suite,
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/suite.py", line 48, in setup_suite
    setup_cases(tests, config_file=config_file, machine=machine,
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/setup.py", line 144, in setup_cases
    setup_case(path, test_case, config_file, machine, machine_info,
  File "/turquoise/usr/projects/climate/mpeterse/repos/compass/pr/compass/setup.py", line 222, in setup_case
    merge_other_config(config, machine_info.config)
UnboundLocalError: local variable 'config' referenced before assignment
```
but maybe I set it up wrong.

xylar:
@mark-petersen, sounds like a bug.  Sorry about that.

xylar:
@mark-petersen, I can't quite figure this out.  It seems like you are probably running with an incorrect version of the code, since line 222 of `setup.py` is a comment in this code.  Could you try running:
```
./conda/configure_compass_env.py ...
source ./load_dev_compass...
```
again and see if it works right?  It may be that your environment doesn't have the right version of the compass package in it.  Keep in mind that it isn't safe to test 2 different compass branches at the same time unless you make 2 separate development environments (using `--env_name`).  Otherwise, the last branch you call `source ./load_dev...` on will replace the version of `compass` installed in your environment.

docs/developers_guide/framework.rst
```diff
@@ -117,60 +117,58 @@ and debugging.
...
+You can add a config file within an MPAS core with the same name as the MPAS
+core plus a ``.cfg`` extension (e.g. ``ocean.cfg``).  This file will
+automatically get loaded during setup.  Similarly, you can add config files
+within test groups or test cases with the same naming convention (e.g.
+``global_ocean.cfg`` for the ``global_ocean`` test group or ``cosine_bell.cfg``
+for the ``cosine_bell`` test case).  Each of these config files will be loaded
+automatically during setup without you needing to add them explicitly.
+
+If you need to add additional config files that serve other purposes, you can
+use the test case's In a test case's :py:meth:`compass.TestCase.add_config()`
```
matthewhoffman, line 129: looks like a typo on this line.

matthewhoffman:
@xylar , are the ocean test case changes related this to PR in a separate branch?  Could you point me to those?  It would help me understand how the changes in this PR work in application.

docs/developers_guide/framework.rst
xylar:
```suggestion
use the test case's :py:meth:`compass.TestCase.add_config()`
```
It sure does.

xylar:
@matthewhoffman, they are in https://github.com/MPAS-Dev/compass/pull/246.  But that PR is kind of a monster so I'm not sure it will make things super clear.

xylar:
I'm closing this for now.  I'll re-open if a pressing need arises for this functionality.

PR #246, opened 2021-09-26
Add a general QU mesh to `global_ocean`

xylar:
The resolution of the QU mesh is 5 km by default but can be modified in a user config file if desired.

This merge also updates the `global_ocean` test group to use the new `add_config()` method of the `TestCase` class.  Rather than adding mesh config files in the `configure()` methods of each test case, they are added at init with the `add_config()` method.


xylar:
This branch is based off of #245 and will need to be rebased after that PR is merged.

xylar:
Documentation of the QU mesh is still needed.

xylar:
It doesn't seem like this general functionality is in demand at the moment.  I'm closing it for now.

xylar:
I don't want to use the solution from #245 at this point.  I think it's sloppy.  Instead, I think we need a way to have "immutable" config options (from a user config file) during setup that get read in at the beginning but can't be changed by any other config file.  That would solve this problem in what I think it a relatively intuitive way.

But that will require some framework development (implementing our variant of the config parser).  That work may also be able to take on #247 where we keep track of where each config option comes from.

xylar:
@dengwirda, I agree.  I think that would be much better.  However, I think we discussed that you provide a number of subdivisions rather than a fixed resolution for those routines?  That's not super easy to make compatible with the current approach for other meshes (not that we can't make QU a special case).

In any case, that would first requires some work in MPAS-Tools.

dengwirda:
@xylar yes you're absolutely correct --- it is a bisection-type approach for structured meshes, so a slightly different set of config. options here.
In terms of MPAS-Tools, one idea that I'd like to advocate for going forward generally is to more directly call `jigsaw` from within COMPASS, since this allows all of the different user-config. options + routines to be exposed, rather than working through MPAS-Tools wrappers. I'd say that to build more complex, non-distance related meshes this has some advantages that could be useful to make available.

xylar:
@dengwirda, I looked into this further and I agree.  It would make sense to make equivalent workflows to `mpas_tools.ocean.build_spherical_mesh()` that allow more flexible usage of `jigsaw` (and other tools).

dengwirda:
Thanks @xylar, I've been building new E3SM meshes/config.'s lately and some extra flexibility there would be excellent!

One option I thought could be to split the routines up a little, and allow users to (optionally) override the [mesh building step](https://github.com/MPAS-Dev/compass/blob/127b6d087d5c52d45b4a09361169cbab3ef7ae99/compass/ocean/tests/global_ocean/mesh/mesh.py#L87) itself, and call something other than the standard `build_spherical_mesh` should they wish to.

I feel that it's perhaps not necessary to try to support all of the `jigsaw` [namelist/config options](https://github.com/dengwirda/jigsaw/wiki/JIG-File-Format) in `compass` directly, and that a customisable workflow would allow folks to just call `jigsaw` direct via `jigsawpy` in some of these steps if they're looking to do something more detailed.

dengwirda:
Yes, sorry: users = developers in my last!

xylar:
#344 should make it possible for the user config options to take priority over all others, and thus make it easier to implement other config options that are based on the user-chosen mesh resolution (or, in this case, number of subdivisions).

xylar:
I'm closing this again because there are several more steps needed to make this possbile.

PR #248, opened 2021-09-28
Update to `mpas_tools` v0.11.0

xylar:
This gets rid of the dependency on `pyflann` and uses `scipy`'s `KDTree` instead.

With this change, `WC14` and `SOwISC12to60` meshes are expected to see small but non-bit-for-bit changes because the signed distance functions used to create them are computed with this new approach.

### DIFF
--- a/conda/compass_env/spec-file.template
+++ b/conda/compass_env/spec-file.template
@@ -17,7 +17,7 @@ jupyter
 lxml
 matplotlib-base
 metis
-mpas_tools=0.10.0
+mpas_tools=0.11.0
 nco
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I tested this with the `pr` test suite on Chrysalis with intel/impi and all tests were bit-for-bit.  This is because the PR test suite doesn't create either of the meshes that use signed-distance functions.

I also created just the meshes for WC14 and SOwISC12to60 to make sure they built correctly (knowing from my previous testing that they won't be bit-for-bit).

PR #249, opened 2021-09-28
Add internal wave test

cbegeman:
This PR ports the internal wave test group from legacy compass.


cbegeman:
@xylar I broke something recently. I'll let you know when it's ready again.

xylar:
No worries, I saw that this was in draft mode, so wasn't going to test it out until you gave the okay.

cbegeman:
@xylar Actually, I think it's ready for review as-is. It failed earlier because I linked to wrong MPAS-O build. Feel free to give suggestions on the viz too. Right now it plots the initial and final states.

.gitignore
```diff
@@ -8,9 +8,6 @@ local.config
 *.o
 *.mod
 
-# Ignore all PNGs
-*.png
-
```
xylar, line 13: Just noting that we discussed this.  Ignoring PNG files causes trouble for the documentation so we don't want to include these in the `.gitignore.

compass/ocean/__init__.py
```diff
@@ -3,6 +3,7 @@
 from compass.ocean.tests.global_convergence import GlobalConvergence
 from compass.ocean.tests.global_ocean import GlobalOcean
 from compass.ocean.tests.gotm import Gotm
+from compass.ocean.tests.internal_wave import InternalWave
```
xylar, line 6: I'm much less concerned about the imports but might as well keep them alphabetical, too.

docs/users_guide/ocean/test_groups/internal_wave.rst
```diff
@@ -0,0 +1,141 @@
...
+-------
+
+``ocean/internal_wave/default`` is the default version of the
+internal wave test case for a short (15 min) test run and validation of
+prognostic variables for regression testing.
+
+rpe_test
+--------
+
+Since mixing is a strong function of horizontal viscosity, this test case performs 20-day integrations of the model forward in time at 5 different values of the viscosity (with steps
```
xylar, line 130:
```suggestion
Since mixing is a strong function of horizontal viscosity, this test case performs 
20-day integrations of the model forward in time at 5 different values of the viscosity (with steps
```

docs/users_guide/ocean/test_groups/internal_wave.rst
```diff
@@ -0,0 +1,141 @@
...
+   :width: 500 px
+   :align: center
+
+By default, the 20 vertical layers each have 25-m uniform
+thickness. By default, all cases have a horizontal resolution of 5 km.
+
+The test group includes 3 test cases.  All test cases have 2 steps,
+``initial_state``, which defines the mesh and initial conditions for the model,
+and ``forward`` (given another name in many test cases to distinguish multiple
+forward runs), which performs time integration of the model. There is an optional ``viz`` step which performs visualization of vertical cross-sections through the center of the domain.
```
xylar, line 25:
```suggestion
forward runs), which performs time integration of the model. There is an optional ``viz`` 
step which performs visualization of vertical cross-sections through the center of the domain.
```

docs/users_guide/ocean/test_groups/internal_wave.rst
```diff
@@ -0,0 +1,141 @@
...
+
+internal_wave
+=============
+
+The ``ocean/internal_wave`` test group induces internal wave propagation and is documented in 
+`Ilicak et al. (2012) <https://doi.org/10.1016/j.ocemod.2011.10.003>`_.
+
+The domain is periodic on the zonal boundaries and solid on the meridional boundaries.
+Salinity is constant throughout the domain (at 35 PSU).  The initial
+temperature has a linear background stratification with a sinusoidal perturbation in the center of the domain. This perturbation initiates symmetric waves that propage out from the center and then back to the center after meeting the solid boundaries.
```
xylar, line 11:
```suggestion
temperature has a linear background stratification with a sinusoidal perturbation in the center 
of the domain. This perturbation initiates symmetric waves that propage out from the center and 
then back to the center after meeting the solid boundaries. 
```

xylar: I'd like to keep the lines easier to read from the terminal, so under 80 characters or so.

compass/ocean/tests/internal_wave/namelist.forward
```diff
@@ -0,0 +1,10 @@
+config_dt = '000:05:00'
+config_btr_dt = '000:00:15'
```
xylar, line 2:
```suggestion
config_dt = '00:05:00'
config_btr_dt = '00:00:15'
```
How you have it works fine but it's a little unusual to have 3-digit hours.

compass/ocean/tests/internal_wave/ten_day_test/namelist.forward
```diff
@@ -0,0 +1,10 @@
+config_dt = '000:05:00'
+config_btr_dt = '000:00:15'
```
xylar, line 2:
```suggestion
config_dt = '00:05:00'
config_btr_dt = '00:00:15'
```

compass/ocean/tests/internal_wave/ten_day_test/namelist.forward
```diff
@@ -0,0 +1,10 @@
...
+config_time_integrator = 'split_explicit'
+config_run_duration = '0010_00:00:00'
+config_use_cvmix = .true.
+config_pio_stride = 4
+config_use_mom_del2 = .true.
+config_mom_del2 = 10.0
+config_implicit_bottom_drag_coeff = 1.0e-2
+config_use_cvmix_convection = .true.
```
xylar, line 10:
```suggestion
config_implicit_bottom_drag_coeff = 1.0e-2
config_use_cvmix_convection = .true.
config_AM_globalStats_enable = .true.
config_AM_globalStats_compute_on_startup = .true.
config_AM_globalStats_write_on_startup = .true.
config_AM_globalStats_text_file = .true.
```
You may have good reason for not including global stats in this run, but the legacy test case includes it and sometimes it's useful.  I'd include it if you don't have a reason not to.

compass/ocean/tests/internal_wave/ten_day_test/streams.forward
```diff
@@ -0,0 +1,23 @@
...
+
+    <stream name="mesh"/>
+    <var_struct name="tracers"/>
+    <var name="xtime"/>
+    <var name="ssh"/>
+    <var name="normalVelocity"/>
+    <var name="layerThickness"/>
+</stream>
+
+</streams>
```
xylar, line 23: Assuming you want global stats:
```suggestion

<stream name="globalStatsOutput"
        output_interval="0000_01:00:00"/>

</streams>
```

compass/ocean/tests/internal_wave/rpe_test/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+
+        Parameters
+        ----------
+        test_group : compass.ocean.tests.internal_wave.InternalWave
+            The test group that this test case belongs to
+        """
+        name = 'rpe_test'
+        super().__init__(test_group=test_group, name=name)
+
+        nus = [1, 5, 10, 20, 200]
```
xylar, line 32: These are different from the viscosities in the legacy test case and in the original paper.  I think these might be copied from the baroclinic channel?

Following, https://github.com/MPAS-Dev/compass/blob/legacy/ocean/internal_waves/5km/rpe_test/plot.py#L10
```suggestion
        nus = [0.01, 1, 15, 150]
```
The formatting of the step names might get a little screwed up if `0.01` turns into something with 16 digits, so you might want to do some formatting magic there.

compass/ocean/tests/internal_wave/rpe_test/namelist.forward
```diff
@@ -0,0 +1 @@
+config_run_duration = '20_00:00:00'
```
xylar, line 1: Just a note to say that I agree with reducing the run duration from 100 days to 20 days.  The analysis is performed at 20 days so no added advantage in running 80 more days by default.  Also, the experiments in the paper were run for 200 days, so 100 days isn't consistent with the paper either.

xylar (commented):
Truly superb work as always.  All 3 test cases together ran perfectly on Ubuntu with gfortran in under 3 minutes.

My only major concern is that the viscosities for the RPE test case aren't the same as in the original test case or the paper.  Maybe that's just a question of having copied the viscosities from the baroclinic channel and not having updated them?

xylar:
@mark-petersen, since you wrote the legacy test case, could you give this a look as soon as possible?  We'd ideally like to merge this one soon.

xylar:
@cbegeman, when you get a chance, could you fix the PEP8 issues that the linter pointed out unless there's a good reason to keep the current formatting (e.g. long module names or URLs)?

compass/ocean/tests/internal_wave/rpe_test/__init__.py
xylar: I feel like I have a slight preference for sticking with the same set of nu values as in the paper, so the 4 values in my suggestion above.   That would make for an easier comparison with the model results from the paper if anyone wanted to do that. Maybe no one will actually want to do that and it's a moot point.

@mark-petersen, do you have an opinion?

xylar:
@cbegeman, let me know when you'd like me to take another look.  It seems form your commits like you've addressed all my concerns.

cbegeman:
@xylar You're welcome to take another look. I did make changes in response to all your suggestions. I figured I'd just wait for @mark-petersen's review.

xylar:
Sounds good!

mark-petersen:
Thank you @cbegeman for this helpful conversion from compass legacy. Tested with gnu optimized on grizzly. Created all plots after run. For example, here is a plot, required little fix I just pushed.
```
/lustre/scratch3/turquoise/mpeterse/runs/n/211004_internal_wave_1/ocean/internal_wave/rpe_test/analysis
sections_internal_wave.png
```
sections_internal_wave

compass/ocean/tests/internal_wave/namelist.forward
```diff
@@ -0,0 +1,10 @@
+config_dt = '00:05:00'
+config_btr_dt = '00:00:15'
+config_time_integrator = 'split_explicit'
+config_run_duration = '0000_00:15:00'
+config_use_cvmix = .true.
+config_pio_stride = 4
+config_use_mom_del2 = .true.
+config_mom_del2 = 10.0
+config_implicit_bottom_drag_coeff = 1.0e-2
+config_use_cvmix_convection = .false.
```
xylar, line 10: @mark-petersen, you pushed this change as part of your commit.  I want to make sure it was intentional, since it wasn't mentioned in the commit message.

In general, I think it's a better practice to suggest changes as part of a review rather than pushing them yourself.

cbegeman:
@mark-petersen Thanks for reviewing and catching the change needed to the viz script!

compass/ocean/tests/internal_wave/namelist.forward
mark-petersen: You are right, that was unintentional. See comment below, I was testing that at the same time. I just corrected it, but I'll do the `suggest changes` method next time.

mark-petersen (approved):
Also run successfully with intel 19 on grizzly, fails with debug as noted above. I think the cvmix convection failure is out of scope for this PR, so I'm approving it. Thanks!

mark-petersen:
I confirmed that my suggested fix in cvmix above,
```
components/mpas-ocean/src/cvmix/src/shared/cvmix_convection.F90:227
227                            CVmix_conv_params_in)
```
does fix this problem, i.e. I can run the test cases in this PR with gnu debug without errors.

mark-petersen:
Issue reported: https://github.com/CVMix/CVMix-src/issues/92

xylar (approved):
I reran with the latest changes using gfortran on Ubuntu and everything looks good to me.

xylar:
Thanks for the hard work on this one, @cbegeman!

PR #250, opened 2021-09-29
Add method for updating streams files at runtime

xylar:
This is similar to the approach for updating namelist files, and can be used when config options that users might edit could affect things like output interval.

### DIFF
--- a/compass/step.py
+++ b/compass/step.py
@@ -460,6 +460,40 @@ class Step(ABC):
             dict(package=package, streams=streams,
                  replacements=template_replacements, mode=mode))
 
+    def update_streams_at_runtime(self, package, streams,
+                                  template_replacements, out_name=None):
+        """
+        Update the streams files during the run phase of this step using the
+        given template and replacements.  This may be useful for updating
+        streams based on config options that a user may have changed before
+        running the step.
+
+        Parameters
+        ----------
+        package : Package
+            The package name or module object that contains the streams file
+
+        streams : str
+            The name of a Jinja2 template to be rendered with replacements
+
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I have tested this as part of #238

xylar:
@matthewhoffman, could you review this by inspection?  Since it affects the framework, I just want another pair of eyes on it.

matthewhoffman (commented):
@xylar, this makes sense.  Do you think there should be a message in the log file indicating the replacement that was made?  I could imagine a user being confused if they made their own changes to the streams file that then got modified during compass.run.

xylar:
@matthewhoffman, I agree that's a good suggestion (and the same thing for namelist options that are updated at runtime).  I'll add both changes to this PR before I merge.

xylar:
I retested this with #238 and with the `pr` test suite, and it worked as expected.  The planar convergence test gave the expected warnings about streams files being updated at runtime:
```
Warning: updating streams in streams.ocean using the following template and replacements:
compass.ocean.tests.planar_convergence streams.template
output_interval = 001_00:00:00
```

xylar:
@matthewhoffman, when you are ready to do so, could you approve this PR?  I need it to be able to merge #238.

matthewhoffman (approved):
@xylar , I'm approving the additional changes by inspection, because I don't want to take the time to set up a test that would use the new functionality.  Thank you for pasting the output snippet of these changes in use - that looks good to me.

xylar:
Yep, that's totally fair.  Thank you for the review!

PR #254, opened 2021-10-07
Fix python 2 bootstrapping issue

xylar:
If we're bootstrapping compass from python 2, we have to be more careful about what we import before creating the install environment.  This merge moves some `mache`-related imports out of the shared code and into the `bootstrap` module so it only gets run with python 3.

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -12,9 +12,32 @@ import grp
 import shutil
 import progressbar
 from jinja2 import Template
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I was having trouble setting put the `compass` conda environment on Grizzly before these changes.  With the changes, I can create an environment just fine.

I'm also able to create a compass conda environment on Chrysalis (as before) with these changes.

PR #257, opened 2021-10-11
Fix compass docs

hollyhan:
This merge 

- fixes typos in multiple places in the compass docs.
- updates the list of test cases shown as an example.
- clarifies the instruction on choosing the right option
   for the machine-specific MPAS target compiler and other flag options.


hollyhan:
@xylar I've requested a pull on the updated compass docs (please see the description on this merge). 
Thank you for reviewing it!

Holly

PR #258, opened 2021-10-12
Add a version of ISOMIP+ test cases with time-varying ice-shelf thickness (pressure)

xylar:
For now, this is just supported for `Ocean0` (a warm-cavity experiment) but could easily be added for other experiments.


xylar:
@cbegeman, no rush at all but this is ready for you to review when you have time.

PR #261, opened 2021-10-15
 Add abstract class functionality

altheaden:
...to step and global ocean mesh.


xylar:
@matthewhoffman, just a heads up about this since it affects framework.  We want to explicitly mark the `run()` method in `Step` as an `abstractmethod`, meaning that a child class *has* to override it or it's an error.  This is mostly just an exercise to help me and @altheaden learn about python abstract classes and methods.

Other methods like `setup()` in `Step` or `configure()` and `validate()` in `TestCase` aren't abstract methods because they don't have to be overridden. It's fine if they just do nothing.

xylar (approved):
Thanks for adding these, @altheaden!  I'm learning some better coding practices from you.

PR #262, opened 2021-10-20
Update MALI-Dev submodule to latest develop

xylar:
The current hash is: [0f8ff9ef09](https://github.com/MALI-Dev/E3SM/tree/0f8ff9ef09446be218bf7cb921c0384d8ac97a05)

The motivation for the update is to bring in recent framework changes needed to build with SCORPIO (inclusion of the `stdc++` library).


xylar:
Okay, great!  That explains the confusion.  @hollyhan and I will test this out on OSX tomorrow so we'll have some more testing there as well.

PR #267, opened 2021-11-03
Add RPE analysis to RPE tests

cbegeman:
This PR adds the computation of Reference Potential Energy (RPE) and a timeseries plot to the analysis step for existing RPE tests for internal wave and baroclinic channel test cases. The details of this computation are in Petersen et al. (2015) https://doi.org/10.1016/j.ocemod.2014.12.004. The timeseries plot is also modeled after figures in that publication.

### DIFF
--- a/compass/ocean/rpe.py
+++ b/compass/ocean/rpe.py
@@ -95,11 +95,8 @@ def compute_rpe(initial_state_file_name='initial_state.nc',
             thickness = np.divide(vol_sorted.tolist(), areaDomain)
 
             # --- RPE computation
-            z = 0.0
-            for i in range(nCells_1D):
-                zMid[i] = z-thickness[i]/2.0
-                z = z-thickness[i]
-            zMid = zMid + bottomMax
+            z = np.append([0.], -np.cumsum(thickness))
+            zMid = z[0:-1] - thickness/2. + bottomMax
             rpe1 = gravity * np.multiply(
                        np.multiply(density_sorted, zMid),
                        vol_sorted)
--- a/compass/ocean/rpe.py
+++ b/compass/ocean/rpe.py
@@ -1,92 +1,126 @@
 import xarray
 import xarray.plot
 import numpy as np
-import datetime
-import matplotlib.pyplot as plt
-from matplotlib.font_manager import FontProperties
+
 
 def compute_rpe(initial_state_file_name='initial_state.nc',
                 output_file_prefix='output_', num_files=5):
-    # --- Open and read vars from NC file 
-    
+    """
+    Computes the reference (resting) potential energy for the whole domain
+    Parameters
+    ----------
+    initial_state_file_name : string
+        Name of the netCDF file containing the initial state
+
+    output_file_prefix : string
+        Prefix for the netCDF file containing output of forward step
+
+    num_files : int
+        Number of files on which to compute rpe
+
+    Returns
+    -------
+    rpe : numpy array of size (num_timesteps) x (num_files)
+    """
+    gravity = 9.80616
+
+    # --- Open and read vars from netCDF file
     dsInit = xarray.open_dataset(initial_state_file_name)
     nCells = dsInit.sizes['nCells']
     nEdges = dsInit.sizes['nEdges']
     nVertLevels = dsInit.sizes['nVertLevels']
 
-    xCell        = dsInit['xCell']
-    yCell        = dsInit['yCell']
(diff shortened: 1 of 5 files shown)


cbegeman:
Example of the resulting rpe figure for the baroclinic channel test case
image

cbegeman:
@xylar and @mark-petersen This PR is ready for review when you have a chance. No rush on this. 

Thanks, @mark-petersen, for the original script. It was great to have that as a starting place.

By the way, since I pulled almost all the RPE computation out of loops and set the analysis interval at 1 day, the computation time is acceptable without threading. Thus, I think it's worth having this analysis as part of the `analysis` step rather than a separate, optional step.

xylar:
@cbegeman, this is very exciting!  I'm glad the performance is now acceptable.  I'll give this a closer look as soon as I can and review it next week.

compass/ocean/rpe.py
```diff
@@ -0,0 +1,126 @@
...
+            density_sorted = density_1D[sorted_ind]
+            vol_sorted = vol_1D[sorted_ind]
+
+            thickness = np.divide(vol_sorted.tolist(), areaDomain)
+
+            # --- RPE computation
+            z = 0.0
+            for i in range(nCells_1D):
+                zMid[i] = z-thickness[i]/2.0
+                z = z-thickness[i]
```
xylar, line 101: This is the kind of loop that one usually wants to avoid in python. It looks like the following might work instead:
```
z = np.append([0], -np.cumsum(thickness))
zMid = z[1:] - z[0:-1]
```

compass/ocean/rpe.py
```diff
@@ -95,11 +95,8 @@ def compute_rpe(initial_state_file_name='initial_state.nc',
             thickness = np.divide(vol_sorted.tolist(), areaDomain)
 
             # --- RPE computation
-            z = 0.0
-            for i in range(nCells_1D):
-                zMid[i] = z-thickness[i]/2.0
-                z = z-thickness[i]
-            zMid = zMid + bottomMax
+            z = np.append([0.], -np.cumsum(thickness))
+            zMid = z[0:-1] - thickness/2. + bottomMax
```
cbegeman, line 99: Thanks @xylar. This substitution worked out.

xylar: That looks reasonable.  My suggestion above should have been:
```
zMid = 0.5*(z[1:] + z[0:-1])
```
but this version looks like it's another way to accomplish the same.

mark-petersen:
Results for RPE plot for internal wave test case:
image
left is from this PR, right is Petersen et al. 2015 paper fig 7b. (Note days is 20 vs 100, vertical scale is 1e-7 vs 1e-6). So these RPE values are certainly on the right scale, and match the publication.

xylar (approved):
This looks good to me.  I don't feel like I need to do any testing you haven't already done.

xylar:
@cbegeman, do you see the "merge pull request" button for this PR?  If so, feel free to use it to merge once @mark-petersen has approved.  If not, I'll take care of it.

mark-petersen:
Here RPE from barotropic channel test case, left this PR, right Petersen et al. 2015 fig 11:
image
The new plot is 4x larger than the publication. I suspect this is the [advection bug](https://github.com/E3SM-Project/E3SM/pull/3809) from one year ago, and that the PR plot is correct and my published plot is not (unfortunately). See how irregular the published plot is, and the orange line goes negative the first 50 days. The new plot is much more regular, as you would expect for an idealized test case. I spent an extra 3 months trying to figure all that out in 2014, and even ran POP simulations for comparisons. I will rerun with the the advection bug put back in to check, but that is outside of the scope of this PR.

mark-petersen (approved):
Thank you @cbegeman. The addition of this analysis script is really helpful.

mark-petersen:
Never mind. On my [previous comment](https://github.com/MPAS-Dev/compass/pull/267#issuecomment-963487937) I was comparing 10km resolution (left) and 4km resolution (right). As a side note, I did test 10km with and without the advection bug and they looked identical, so scratch that too.

cbegeman:
@mark-petersen Thanks for doing those tests. I never did a test that ran out to 200 days but I also thought the magnitude seemed about right in comparison with your paper from 0-20 days.

@xylar I do not have permission to merge. You can do that at your convenience.

xylar:
Nice, thanks @cbegeman!

PR #271, opened 2021-11-10
Add Thwaites mesh creation step

trhille:
This PR adds mesh creation for a MALI Thwaites Glacier regional domain.
Config options `min_spac` and `max_spac` set minimum and maximum cell space.
The default is to use 1–8km resolution, but this can be changed as desired.
Uses ice surface speed, distance to grounding line, and distance to ice edge in mesh density function.


matthewhoffman:
I was thinking that before we add a 2nd mesh generation test case for landice to compass, we should generalize the mesh gen function.  Given that will take some additional work, I would be ok not having this test in the v1.0 release.  @trhille and @xylar , let me know if you disagree with any of that.  (Also I'm willing to help generalize the mesh gen function.)

xylar:
@matthewhoffman, I fully agree.  The primary purpose of making compass a python package was to have as much reusable code as possible.

trhille:
Force-push from e1d2ec3 to 52d1b49 was a rebase onto compass/master following merge of PR #322.

trhille:
Note: call to `compass.landice.mesh.get_dist_to_edge_and_GL()` took 72.52 seconds in this test, so @xylar's suggestion on PR #322 to use a faster external package like `scikit-fmm` will probably speed this case up considerably!

xylar:
Ha, ha!  Just saw this comment.  Okay, might be worth looking into `scikit-fmm` down the road but as long as this isn't the limiting step in the mesh creation, I still would let it be for now.

compass/landice/tests/thwaites/mesh.py
```diff
@@ -0,0 +1,218 @@
...
+                'thwaites_dehorned.nc', '-o',
+                'Thwaites_1to8km.nc', '-l', levels, '-v', 'glimmer',
+                '--beta', '--thermal', '--obs', '--diri']
+
+        check_call(args, logger=logger)
+
+        logger.info('calling interpolate_to_mpasli_grid.py')
+        args = ['interpolate_to_mpasli_grid.py', '-s',
+                'antarctica_1km_2020_10_20_ASE.nc',
+                '-d', 'Thwaites_1to8km.nc', '-m', 'b', '-t']
```
matthewhoffman, line 151: Speaking of potential speedup, switching the interpolation method from bilinear to delaunay (barycentric) might help here, but then I also see the `-t` option here that would only interpolate thickness.  Don't we want to interpolate everything in this step?

trhille: I remember bilinear being _much_ faster than barycentric, unless we save the barycentric weights somewhere.

trhille: We eventually want to interpolate everything, but we have not done that within COMPASS before, so I was thinking that should be in a separate PR. It will be a considerable effort, and will include switching over to the conservative remapping as well.

matthewhoffman: Yes bilinear is faster for one or a few fields, but the barycentric only calculates weights once, so it may be faster when there are a lot of fields.

I see about not doing the final step in this PR.  What do you think about dropping `-t` anyway so what comes out of COMPASS is at least a rough draft of all the data off the shelf?  Then we can still rerun the interpolation manually.

compass/landice/tests/thwaites/high_res_mesh/__init__.py
```diff
@@ -0,0 +1,25 @@
+from compass.testcase import TestCase
+from compass.landice.tests.thwaites.mesh import Mesh
+
+
+class HighResMesh(TestCase):
+    """
+    The high resolution test case for the thwaites test
+    group simply creates the mesh and initial condition.
```
matthewhoffman, line 8: Mention basal friction optimization occurs separately.

compass/landice/tests/thwaites/mesh.py
```diff
@@ -0,0 +1,218 @@
+import numpy as np
+import netCDF4
+import xarray
+from matplotlib import pyplot as plt
+
+from mpas_tools.mesh.creation import build_planar_mesh
+from mpas_tools.mesh.conversion import convert, cull
+from mpas_tools.planar_hex import make_planar_hex_mesh
```
matthewhoffman, line 8: This can be removed, right?

docs/users_guide/landice/test_groups/thwaites.rst
```diff
@@ -50,3 +83,11 @@ of the model forward in time (``full_run`` step).  Then, a second step
...
 are compared between the "full" and "restart" runs to make sure they are
 bit-for-bit identical.
+
+high_res_mesh
+-------------
+
+``landice/thwaites/high_res_mesh`` creates a variable resolution mesh based
+on the the config options listed above. This will not be the same as the
+pre-generated 4-14km mesh used in ``decomposition_test`` and ``restart_test``
+because it uses a newer version of Jigsaw.
```
matthewhoffman, line 93: Also mention the basal friction optimization is not handled.

compass/landice/tests/thwaites/mesh.py
```diff
@@ -0,0 +1,218 @@
...
+        # This step is only necessary because the GeoJSON region
+        # is defined by lat-lon.
+        logger.info('calling set_lat_lon_fields_in_planar_grid.py')
+        args = ['set_lat_lon_fields_in_planar_grid.py', '-f',
+                'ase_1km_preCull.nc', '-p', 'ais-bedmap2']
+
+        check_call(args, logger=logger)
+
+        logger.info('calling MpasMaskCreator.x')
+        args = ['MpasMaskCreator.x', 'ase_1km_preCull.nc',
```
matthewhoffman, line 119: @xylar , should this be replaced with the `mask` method in MPAS-Tools' mesh conversion package?

matthewhoffman (changes requested):
Looks good, @trhille .  A few small changes requested.

matthewhoffman:
@trhille , did you notice some regions of very high res near the Siple Coast divide (far right and bottom right corner)?  

image

Could that be from missing velocity obs that are getting filled with bad data values that happen to be really large?

image

We could either have this test case look out for and remove such regions, or maybe we need to do something to the data file we are using.

trhille:
@matthewhoffman Good catch. I will look into that.

xylar:
Subglacial lakes?

compass/landice/tests/thwaites/mesh.py
xylar: It seems like you might not want to have your own tool for this.  I think this is something that `ESMF_RegridWeightGen` and `ncremap` should be able to handle.  You can use `pyremap` to generate the MPAS script files, create the weight files, and do the remapping:
https://github.com/MPAS-Dev/pyremap/blob/master/examples/make_mpas_to_Antarctic_stereo_mapping.py
It makes calls to `E3SM_RegridWeightGen` and `ncremap`.

compass/landice/tests/thwaites/mesh.py
xylar: I don't think it matters too much.  `mask()` is just a wrapper that takes `xarray.Dataset` and `geometric_features.FeatureCollection` objects as its inputs.  In this case where the files are already written, the wrapper doesn't save work.

But you should maybe consider using the alternative tool: http://mpas-dev.github.io/MPAS-Tools/stable/mesh_conversion.html#mask-creation-with-python-multiprocessing
The tool is called `compute_mpas_region_masks` and seems to work better if called as a subprocess.  It should be significantly faster than `MpasMaskCreator.x` on HPC nodes because it uses multiprocessing.

xylar:
If that high-res region does turn out to be related to bad values, it would be worth considering interpolating a mask of bad values to the MALI mesh and normalizing the results based on that mask (I can walk you through the details) where you need a renormalization threshold below which there isn't enough valid data to use and the remapped data gets masked out as invalid.  This is something that `ncremap` will do for you automatically if you provide it with the right flags and the input data set has its data marked as invalid in the expected way (with a `_FillValue` attribute).

xylar (changes requested):
I have some suggestions, nearly all just clean-up.

compass/landice/tests/thwaites/mesh.py
trhille: > Yes bilinear is faster for one or a few fields, but the barycentric only calculates weights once, so it may be faster when there are a lot of fields.

@matthewhoffman I confirmed that barycentric takes ~25% longer than bilinear. That's not a huge deal for a smallish mesh like this, but I'd push to leave it as-is for this PR and then update interpolation for all real-geometry test-cases (i.e., Humboldt, Thwaites, AIS) in another PR following @xylar's  suggestion.

trhille:
Before (left) and after (right) commit c67a2dc:
image

The isolated pocket of high mesh density to the righthand side of the left panel was due to missing velocity values in the input dataset that were treating as 10.e36 when not using masked arrays in the Python script.

compass/landice/mesh.py
```diff
@@ -167,6 +167,18 @@ def set_cell_width(self, section, thk, vx=None, vy=None,
...
+        # Clean up where we have missing velocities. These are usually nans
+        # or the default netCDF _FillValue of ~10.e36
+        missing_data_mask = np.logical_or(
+                               np.logical_or(np.isnan(vx), np.isnan(vy)),
+                               np.logical_or(np.abs(vx) > 1.e5,
+                                             np.abs(vy) > 1.e5))
+        spacing[missing_data_mask] = max_spac
+        logger.info('Found {} points in input dataset with missing ' \
+                     'velocity values. Setting velocity-based spacing ' \
+                     'to maximum value.'.format(np.sum(missing_data_mask)))
```
xylar, line 180:
```suggestion
        logger.info('Found {} points in input dataset with missing '
                    'velocity values. Setting velocity-based spacing '
                    'to maximum value.'.format(np.sum(missing_data_mask)))
```
These are from https://github.com/MPAS-Dev/compass/pull/271#issuecomment-965526634

You don't need a backslash for continuing lines that are already in parentheses like here.

compass/landice/tests/thwaites/__init__.py
```diff
@@ -1,7 +1,7 @@
 from compass.testgroup import TestGroup
 from compass.landice.tests.thwaites.decomposition_test import DecompositionTest
 from compass.landice.tests.thwaites.restart_test import RestartTest
-
+from compass.landice.tests.thwaites.high_res_mesh import HighResMesh
```
xylar, line 4:
```suggestion
from compass.landice.tests.thwaites.high_res_mesh import HighResMesh

```
Two blank lines after imports.

xylar:
Two more very small things.

compass/landice/tests/thwaites/mesh.py
xylar: @trhille, do you want to try `compute_mpas_region_masks` or are you pretty convinced that's not a time-limiting step here?

trhille: @xylar , I added a timer on the call to MpasMaskCreator.x on a 2-8km test, and it only took 9 seconds out of 3:06 total runtime, so I'm happy to leave it as-is for now.

trhille:
> Ha, ha! Just saw this comment. Okay, might be worth looking into `scikit-fmm` down the road but as long as this isn't the limiting step in the mesh creation, I still would let it be for now.

The nice thing is that this is dependent on the input dataset size, not our MPAS mesh resolution. So this still only takes 72 seconds for a 500m–8km mesh that took a total of 23 minutes to run.

compass/landice/tests/thwaites/mesh.py
matthewhoffman: Thanks for testing that out.  I agree bilinear vs. barycentric doesn't matter for this case and the performance differences are too small to matter.

In the long run Xylar's suggestion of pyremap sounds wise.  Our current tool is highly specialized for our workflow, however, so we'll wait until we reach a point of specifically needing new capabilities before looking into replacing it.

compass/landice/mesh.py
```diff
@@ -167,6 +167,18 @@ def set_cell_width(self, section, thk, vx=None, vy=None,
...
+        # Clean up where we have missing velocities. These are usually nans
+        # or the default netCDF _FillValue of ~10.e36
+        missing_data_mask = np.logical_or(
+                               np.logical_or(np.isnan(vx), np.isnan(vy)),
+                               np.logical_or(np.abs(vx) > 1.e5,
+                                             np.abs(vy) > 1.e5))
+        spacing[missing_data_mask] = max_spac
+        logger.info(f'Found {np.sum(missing_data_mask)} points in input '
+                    f'dataset with missing velocity values. Setting '
+                    f'velocity-based spacing to maximum value.')
```
matthewhoffman, line 180: This does not seem like a general solution (what if the missing data was on an ice shelf or an ice stream?) but will work for this test case.  

The simplest thing I can think of to deal with the holes is the following (pseudocode):
```
do while still missing cells:
    loop over missing cells:
        if this cell is missing but has a nonmissing neighbor(s):
            assign it the average value of its nonmissing neighbors
```
As written, it would be dependent on the arbitrary order in which cells are looped over.  It also will probably be ugly if the hole is large.  But so far we've only seen small holes, and we don't care a lot about how we patch them - just that they are patched somehow.

The other way I can think to do it is to built an interpolation triangulation out of the good cells and use it to extrapolate into the missing cells.  Or build a kd tree and use that to take the value of the nearest good cell.
But I suspect either would be very computationally expensive relative to the previous idea, at least if we typically expect just a small holes.

matthewhoffman (commented):
@trhille , I looked over this PR, and I only had one topic to discuss further.  I added a comment about how to fill the holes instead of giving them the max value.  I was going to just say go with what you implemented (giving holes the coarsest value), but I suspect this issue may affect the AIS mesh, and perhaps we want to implement a more general solution now.  Let me know what you think.

compass/landice/mesh.py
trhille: My thought here was that it would be better to use the data that are actually available. For example, the cell spacing set by distance to the grounding line or ice edge would over-ride the maximum distance set by the `use_speed` config option. But I can see how that would be problematic if the missing data were in the middle of the Ross Ice Shelf (i.e., fast flowing, but far from both grounding line and ice edge). The best solution might actually be case dependent and require some judgment call based on the region in question.
I think a better solution than a loop would just be to use `scipy.interpolate.NearestNDInterpolator`. I will give that a try.

xylar: Would it be possible to patch the holes in the velocity data outside of compass and use the patched data set?  It seems like it would be worth doing a front propagation technique to fill in the holes.  I believe https://github.com/scikit-fmm/scikit-fmm provides such a capability but it likely wouldn't be fast enough to want to run it every time you make a mesh (especially for testing).

xylar: This is what I was thinking of in particular: 

Final example here:
https://scikit-fmm.readthedocs.io/en/latest/#examples
Code here:
https://scikit-fmm.readthedocs.io/en/latest/examples.html#extension-velocities-example-py

xylar: @trhille, your idea is also a good one.  I haven't had much luck with `scipy`'s interpolators in the past.  They're designed for interpolating a lot of points in a given field, rather than for interpolating a lot of fields at a given set of points, so they're not useful for remapping data.  But they might work fine for this particular case.  What I said still holds, though.  It would make more sense to do this once and have a filled-in velocity field that you use instead of the one with holes.

trhille: My concern with doing this outside of compass is that it adds another separate step that makes the reproducible workflow harder to track. However, maybe we should have a step that checks for missing values and fills them like I have done here, and then print a warning that this might not be a universal solution and that mesh quality should be checked manually? On the other hand, few of us are likely to look through the case output log thoroughly enough to see that warning.

xylar: I really think it's legit to have steps to preprocess the input data files that aren't part of the compass steps you run on a regular basis.  But if you can come up with a good way to fill in the gaps that doesn't make the test case take significantly longer, it's fine to keep it here.

We rerun the ocean mesh generation (at low resolution) as part of our regression testing.  I think you would want to do that in MALI, too (but that's up to you) so you presumably want a process that doesn't take more time than other regression tests if possible.

trhille: Okay, so maybe we create a generalized tool for pre-processing data that lives in compass/landice. So it would still be part of compass, but it doesn't need to be called during mesh generation.

trhille: Thanks @xylar. @matthewhoffman, what do you think? I feel like adding a pre-processing tool is beyond the scope of this PR. I can add some comments and/or a warning message that makes it clear that this isn't a universal fix.

trhille: And I can try to get another PR with that pre-processing tool opened in time for use in PR #324.

xylar (approved):
I think all my comments have been addressed and I'm okay with the temporary fix hear for missing velocity values as long as a longer-term fix goes in in another PR.

xylar:
@matthewhoffman and @trhille, if this is going into compass 1.0, it needs to be merged in the next few days.  Could you give me a sense of 1) if that seems likely and 2) if it's a priority.

matthewhoffman:
@xylar , sorry for letting this languish and thanks for the reminder.  I would be happy to approve the current version of this PR in order to get this into 1.0 and deal with the known issues later.  I think that's ok, as they could arguably be thought of as more general than what this PR is trying to accomplish.

@trhille , what do you think?

PR #272, opened 2021-11-10
Add Humboldt test group

trhille:
This PR adds a Humboldt Glacier test group. Currently, it contains only a 'default' test case that creates a 1-10 km Humboldt regional mesh. Future additions will add options for different resolutions.

Note: This mesh is not quite identical to the mesh from the old version of COMPASS. This could be due to different versions of Jigsaw, and thus we have decided not to pursue making these meshes identical unless a specific need arises.


xylar:
@trhille, thanks for the PEP8 work.  I promise that it isn't so bad once you get used to it... And that it's worth it in the end to have nice, consistently formatted code...

xylar:
@trhille, do you have an update on this PR?  One thing it would need before it can be merged is documentation.  Let me know if you need help on how to do that, but I can also follow up with that as part of my review.

trhille:
@xylar, my apologies for dropping the ball on this. I haven't had any time to work on this PR, but I should be able to find time soon to work on documentation. I'll look through examples elsewhere in the repo and let you know if I need help or clarification.

xylar:
@trhille, thanks for adding some documentation.  I realize you haven't asked for feedback yet but wanted to point to 2 other spots where documentation is still needed.  I'll point to the Thwaites equivalents.

First, the automatically generated API documentation:
https://github.com/MPAS-Dev/compass/blob/master/docs/developers_guide/landice/api.rst#thwaites
Second, the Developer's Guide:
https://github.com/MPAS-Dev/compass/blob/master/docs/developers_guide/landice/test_groups/thwaites.rst

In case it isn't clear, the User's Guide should describe only what someone needs to know to run the test without knowing anything about the code.  The Developer's Guide is more about how the test group is structured in the repository -- where the test cases and steps are and whatever shared code, namelist files, etc. there might be between test cases.

docs/users_guide/landice/test_groups/humboldt.rst
```diff
@@ -0,0 +1,30 @@
+.. _landice_humboldt:
+
+humboldt
+=========
```
xylar, line 4:
```suggestion
humboldt
========
```

docs/users_guide/landice/test_groups/humboldt.rst
```diff
@@ -0,0 +1,30 @@
...
+
+The test group uses the following default config options:
+
+.. code-block:: cfg
+
+    # number of levels in the mesh
+    levels = 10
+
+default
+------------------
```
xylar, line 27:
```suggestion
default
-------
```

trhille:
@xylar Okay, thanks!

docs/developers_guide/landice/test_groups/humboldt.rst
```diff
@@ -0,0 +1,35 @@
+.. _dev_landice_humboldt:
+
+humboldt
+========
+
+The ``humboldt`` test group (:py:class:`compass.landice.tests.humboldt.Humboldt`)
+creates a variable resolution 1-10 km mesh for a regional domain of Humboldt Glacier
+(see :ref:`landice_humboldt`).  Here, we describe the shared framework for
+this test group and the singe test case.
```
matthewhoffman, line 9:
```suggestion
this test group and the single test case.
```

docs/users_guide/landice/test_groups/humboldt.rst
```diff
@@ -0,0 +1,30 @@
+.. _landice_humboldt:
+
+humboldt
+========
+
+The ``landice/humboldt`` test group creates a 1-10km Humboldt Glacier mesh.
```
matthewhoffman, line 6:
```suggestion
The ``landice/humboldt`` test group creates a 1-10km mesh for Humboldt Glacier, Greenland.
```

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+        print('calling set_lat_lon_fields_in_planar_grid.py')
+        args = ['set_lat_lon_fields_in_planar_grid.py', '-f',
+                'Humboldt_1to10km.nc', '-p', 'gis-gimp']
+        check_call(args, logger=logger)
+
+        print('creating graph.info')
+        make_graph_file(mesh_filename='Humboldt_1to10km.nc',
+                        graph_filename='graph.info')
+
+    def build_cell_width(self):
```
matthewhoffman, line 160: Can you add a docstring for this function?  I think eventually this function could be abstracted as a general mesh gen function.  It's fine to not deal with that in this PR, but make a note of it in the docstring.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+        f.set_auto_mask(False)  # disable masked arrays
+
+        x1 = f.variables['x1'][:]
+        y1 = f.variables['y1'][:]
+        thk = f.variables['thk'][0, :, :]
+        topg = f.variables['topg'][0, :, :]
+        vx = f.variables['vx'][0, :, :]
+        vy = f.variables['vy'][0, :, :]
+
+        # subset data - optional
```
matthewhoffman, line 173: Add a note elaborating why this might be done.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+        vx = vx[::step, ::step]
+        vy = vy[::step, ::step]
+
+        dx = x1[1] - x1[0]  # assumed constant and equal in x and y
+        nx = len(x1)
+        ny = len(y1)
+
+        sz = thk.shape
+
+        # define extent of region to mesh
```
matthewhoffman, line 188: Make a note that these coords are specific to Humboldt.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+        iStart = sz[0] // 2
+        jStart = sz[1] // 2
+        floodMask[iStart, jStart] = 1
+
+        neighbors = np.array([[1, 0], [-1, 0], [0, 1], [0, -1]])
+
+        lastSearchList = np.ravel_multi_index([[iStart], [jStart]],
+                                              sz, order='F')
+
+        # flood fill -------------------
```
matthewhoffman, line 219: Add a note here again that we may want to generalize this into a separate function in the future.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+        # where ice exists and neighbors non-ice locations
+        marginMask = np.logical_and(marginMask, iceMask)
+        # optional - plot mask
+        # plt.pcolor(marginMask); plt.show()
+
+        # calc dist to margin -------------------
+        [XPOS, YPOS] = np.meshgrid(x1, y1)
+        distToEdge = np.zeros(sz)
+
+        # -- KEY PARAMETER: how big of a search 'box' (one-directional) to use.
```
matthewhoffman, line 275: Elaborate on this comment - what are we searching for?  (I realize you maybe didn't write this and maybe I did, but the details are foggy even to me...)

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+
+        print('calling interpolate_to_mpasli_grid.py')
+        args = ['interpolate_to_mpasli_grid.py', '-s',
+                'humboldt_1km_2020_04_20.epsg3413.icesheetonly.nc', '-d',
+                'gis_1km_preCull.nc', '-m', 'b', '-t']
+
+        check_call(args, logger=logger)
+
+        print('calling define_cullMask.py')
+        args = ['define_cullMask.py', '-f',
```
matthewhoffman, line 98: Is this step necessary if we are culling to geojson extent below? Or do we want the union of both masks?  If so, please add a comment pointing that out.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+                'gis_1km_preCull.nc', '-m'
+                'distance', '-d', '50.0']
+
+        check_call(args, logger=logger)
+
+        print('calling set_lat_lon_fields_in_planar_grid.py')
+        args = ['set_lat_lon_fields_in_planar_grid.py', '-f',
```
matthewhoffman, line 105: Add a note that this step is only necessary because the GeoJSON region is defined by lat/lon.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,347 @@
...
+        args = ['create_landice_grid_from_generic_MPAS_grid.py', '-i',
+                'humboldt_dehorned.nc', '-o',
+                'Humboldt_1to10km.nc', '-l', levels, '-v', 'glimmer',
+                '--beta', '--thermal', '--obs', '--diri']
+
+        check_call(args, logger=logger)
+
+        print('calling interpolate_to_mpasli_grid.py')
+        args = ['interpolate_to_mpasli_grid.py', '-s',
+                'humboldt_1km_2020_04_20.epsg3413.icesheetonly.nc',
```
matthewhoffman, line 142: Is this file just a subset of the whole GIS 1km dataset?  Did you create it separately just to allow the interpolation step to run faster?

matthewhoffman (changes requested):
@trhille , thank you for putting this together into COMPASS and pulling the PR together.  I know it was a lot of work, but this will be a great landice example to build other mesh generation test cases from.  Overall it looks good and there were no surprises for me.  I made a number of small comments/requests.  Let me know if anything is unclear or you want to discuss any of it.

compass/landice/tests/humboldt/mesh.py
trhille: Yes, I just cut it out of the whole GIS 1km dataset using ncks. I could instead replace it with the whole dataset and subsample it in this script. What would you prefer?

compass/landice/tests/humboldt/mesh.py
trhille: We could probably remove this. It's legacy code that was never used, and it's potentially dangerous if inadvertently changed.

compass/landice/tests/humboldt/mesh.py
trhille: That's a good question. I didn't write this, but it looks like this is finding the distance from every ice-filled cell to the nearest ice-sheet margin (distToEdge), which is then used as part of the mesh density function. I think the windowSize parameter is the length scale over which the distance from the ice-sheet margin affects the density function. But, as written this is pretty difficult to decipher. Not sure if we should just comment it much more thoroughly, or have a stab at rewriting the code to be more intuitive.

compass/landice/tests/humboldt/mesh.py
trhille: Testing confirms that this isn't necessary, as xCell and yCell are BFB between meshes created with and without this step. Deleting.

compass/landice/tests/humboldt/mesh.py
trhille: Added

compass/landice/tests/humboldt/mesh.py
trhille: Added.

compass/landice/tests/humboldt/mesh.py
trhille: Okay, I've improved this comment.

trhille: This calculation of the distance to the ice edge could probably also be separated out into its own function for general use.

compass/landice/tests/humboldt/mesh.py
trhille: Just adding a note here that this could still be useful if you want to cull a certain distance from the ice margin, within the geometry defined by the GeoJSON file. This step will be kept in the Thwaites mesh PR (PR #271), but it's fine to remove it here.

matthewhoffman: I'm ok with leaving it in here to make this a general example.  If you leave it, add some comments explaining when/why this could be useful and that it's otherwise not necessary.

compass/landice/tests/humboldt/mesh.py
matthewhoffman: As long as you already have the subset created, let's just stick with it.  But add some comments here that the whole ice sheet data file could be used instead and the subset was created just to make the test run faster.

matthewhoffman (changes requested):
@trhille , thanks for revising this PR.  It looks good.  I made two new small comments that ask for some extra comments.  Should only take a couple minutes.

compass/landice/tests/humboldt/mesh.py
trhille: Okay, I added this back in commit [8635048](https://github.com/MPAS-Dev/compass/pull/272/commits/8635048e9f664399f5681c23d8a2a2f40b8dd54b), and changed it to cull within 5 km of the ice margin so that it has an effect on the mesh. I added a note that it can safely be removed.

compass/landice/tests/humboldt/mesh.py
trhille: Okay I added this comment in commit [2e78220](https://github.com/MPAS-Dev/compass/pull/272/commits/2e782205d1300d4e4f436f7c8c01cdf9852aa080).

trhille:
@matthewhoffman Thanks for this review. I'm going to change mark this as Ready for Review.
@xylar, sorry this has taken so long!

matthewhoffman (approved):

compass/landice/tests/humboldt/mesh.py
xylar: Could you add a config option that determines if this step happens or not?  You should not plan to have users look at the code at all, the info they need should come through the documentation and the config options they have available to them.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,361 @@
...
+        args = ['set_lat_lon_fields_in_planar_grid.py', '-f',
+                'gis_1km_preCull.nc', '-p', 'gis-gimp']
+
+        check_call(args, logger=logger)
+
+        print('calling MpasMaskCreator.x')
+        args = ['MpasMaskCreator.x', 'gis_1km_preCull.nc',
+                'humboldt_mask.nc', '-f', 'Humboldt.geojson']
+
+        check_call(args, logger=logger)
```
xylar, line 123: You might want to use the faster python-based mask creation that uses multiprocessing

http://mpas-dev.github.io/MPAS-Tools/stable/mesh_conversion.html#mask-creation-with-python-multiprocessing

Here's and example of using it in `compass`:
https://github.com/MPAS-Dev/compass/blob/3e526c42e2e430f02833329534a81a3b173ab21c/compass/ocean/tests/global_ocean/mesh/cull.py#L134-L143

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,361 @@
...
+               'gis_1km_preCull.nc', '-m'
+               'distance', '-d', '5.0']
+
+        check_call(args, logger=logger)
+
+        # This step is only necessary because the GeoJSON region
+        # is defined by lat-lon.
+        print('calling set_lat_lon_fields_in_planar_grid.py')
+        args = ['set_lat_lon_fields_in_planar_grid.py', '-f',
+                'gis_1km_preCull.nc', '-p', 'gis-gimp']
```
xylar, line 115: We need @hollyhan's fix in https://github.com/MPAS-Dev/MPAS-Tools/pull/450 for the longitude variables to be in the desired range.  I don't know if this causes issues with `MpasMaskCreator.x`.  Presumably you would have noticed if it did.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,361 @@
...
+        floodMask[iStart, jStart] = 1
+
+        neighbors = np.array([[1, 0], [-1, 0], [0, 1], [0, -1]])
+
+        lastSearchList = np.ravel_multi_index([[iStart], [jStart]],
+                                              sz, order='F')
+
+        # flood fill -------------------
+        # In the future this might become a separate function
+        # to be used by multiple test groups.
```
xylar, line 232: I would definitely be in favor of pulling this out into the `compass.landice` framework in the future.

xylar (changes requested):
Could you do an interactive rebase onto the current `compass/master` and, in the process, change `pick' to `fixup` for the following 5 commits?
```
* 8b4a3b82 Update docs/users_guide/landice/test_groups/humboldt.rst
* 8bfc557a Update docs/users_guide/landice/test_groups/humboldt.rst
...
* fc36812c Fix PEP 8 issues in humboldt/mesh.py
* 0c302dfd Fix PEP 8 issues in humboldt/default/__init__.py
* 65fffe3f Fix PEP 8 issues in humboldt/__init__.py
```
You may wish to do other clean-up in the commit history but that's up to you.

I'm seeing `bedTopography` is all zeros in the output.  Is that expected? I thought that might be one of the fields that should have been interpolated from the full Greenland dataset or the cropped version.

xylar:
To build the documentation, do:
```
cd docs
make html
```
When I do that, I'm seeing:
```
/home/xylar/code/compass/add_Humboldt_test_group/docs/developers_guide/landice/api.rst:206: WARNING: Title underline too short.

humboldt
~~~~~~~
/home/xylar/code/compass/add_Humboldt_test_group/docs/developers_guide/landice/api.rst:206: WARNING: Title underline too short.

humboldt
~~~~~~~
/home/xylar/code/compass/add_Humboldt_test_group/docs/developers_guide/landice/test_groups/thwaites.rst:42: WARNING: duplicate label dev_landice_thwaites_decomposition_test, other instance in /home/xylar/code/compass/add_Humboldt_test_group/docs/developers_guide/landice/test_groups/humboldt.rst
looking for now-outdated files... none found
pickling environment... done
checking consistency... /home/xylar/code/compass/add_Humboldt_test_group/docs/developers_guide/landice/test_groups/humboldt.rst: WARNING: document isn't included in any toctree
/home/xylar/code/compass/add_Humboldt_test_group/docs/users_guide/landice/test_groups/humboldt.rst: WARNING: document isn't included in any toctree
```
Those underlines need to be the same length as the titles.  You need to add `humboldt` to the `index.rxt` files for both the `landice` user's and developer's guides.

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,361 @@
...
+
+        self.add_output_file(filename='graph.info')
+        self.add_output_file(filename='Humboldt_1to10km.nc')
+        self.add_input_file(
+                filename='humboldt_1km_2020_04_20.epsg3413.icesheetonly.nc',
+                target='humboldt_1km_2020_04_20.epsg3413.icesheetonly.nc',
+                database='')
+        self.add_input_file(filename='Humboldt.geojson',
+                            target='Humboldt.geojson',
+                            database=None, copy=True)
```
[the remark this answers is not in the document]
trhille: Oh great, thanks. I was having to run compass setup from the `compass/landice/tests/humboldt` directory, which worked but did seem wrong.
`copy=True` just resulting from me misunderstanding how this worked, so I will remove that.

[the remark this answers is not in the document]
xylar: Okay, great!  Glad we got that sorted out.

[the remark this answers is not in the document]
trhille: Fixed in commit [2fe87fb](https://github.com/MPAS-Dev/compass/pull/272/commits/2fe87fb99029b1f3ef56ed68d641e8f47957d613).

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,361 @@
...
+        Run this step of the test case
+        """
+        logger = self.logger
+        try:
+            config = self.config
+            section = config['humboldt']
+        except NameError:
+            print('No config file; skipping')
+
+        print('calling build_cell_wdith')
```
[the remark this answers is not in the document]
trhille: Fixed in commit [a16785e](https://github.com/MPAS-Dev/compass/pull/272/commits/a16785eb18497c759a7cbf2f921f4108d5151611).

compass/landice/tests/humboldt/mesh.py
trhille: Added cullDistance config option in commit [6bfb52b](https://github.com/MPAS-Dev/compass/pull/272/commits/6bfb52b2aa2e2c948fec1e4e8ff2261162216682).

compass/landice/tests/humboldt/mesh.py
```diff
@@ -0,0 +1,361 @@
...
+        if float(cullDistance) > 0.:
+           logger.info('calling define_cullMask.py')
+           args = ['define_cullMask.py', '-f',
+                   'gis_1km_preCull.nc', '-m'
+                   'distance', '-d', cullDistance]
+
+           check_call(args, logger=logger)
+        else:
+           logger.info('cullDistance <= 0 in config file. ' \
+                       'Will not cull by distance to margin. \n')
```
xylar, line 109:
```suggestion
        if float(cullDistance) > 0.:
            logger.info('calling define_cullMask.py')
            args = ['define_cullMask.py', '-f',
                    'gis_1km_preCull.nc', '-m'
                    'distance', '-d', cullDistance]

            check_call(args, logger=logger)
        else:
            logger.info('cullDistance <= 0 in config file. ' \
                        'Will not cull by distance to margin. \n')
```
This needs the proper level of indentation.

trhille:
@xylar bedTopography is not expected to be interpolated here. We have generally done all the interpolation outside of COMPASS after generating the mesh. We could change that, but I think it's beyond the scope of this PR. @matthewhoffman, do you agree with that?

matthewhoffman:
@trhille , I think historically we did the interpolation by hand post-compass because things were so tweaky, but ideally we'd include the interpolation step in COMPASS as well.  Depending on how complicated that is for this case, we can either add it here or leave it out and plan to incorporate that step into the next test case we set up.

xylar:
I'm fine either way.  Maybe it's time to let this PR go in and let that be future work?

xylar (approved):
@trhille, this looks good!

```
$ compass run
landice/humboldt/default
  * step: mesh
  test execution:      SUCCESS
Test Runtimes:
03:33 PASS landice_humboldt_default
Total runtime 03:33
```

xylar:
@matthewhoffman, this is assigned to you.  As far as I'm concerned, you can merge it as soon as you and @trhille are ready.

trhille:
@xylar and @matthewhoffman, thanks for your help and careful reviews!

PR #273, opened 2021-11-14
Port the SOMA test group

xylar:
The Simulating Ocean Mesoscale Activity (SOMA) test group are a set of double gyre tests documented in [Wolfram et al. 2015](https://doi.org/10.1175/JPO-D-14-0260.1).

The `default` test cases use 100 vertical layers distributed according to layer thicknesses defined for some E3SMv1 meshes.

Other SOMA test cases (3-layer, time-varying wind and surface-restoring) will be ported as needed at a later time.


xylar:
## Testing

- [x] `soma/32km/default` on Ubuntu with gnu/mich
- [ ] `soma/16km/default` on Ubuntu with gnu/mich
- [ ] `soma/32km/default` on Chrysalis with intel/impi
- [ ] `soma/16km/default` on Chrysalis with intel/impi
- [ ] `soma/8km/default` on Chrysalis with intel/impi
- [ ] `soma/4km/default` on Chrysalis with intel/impi

xylar:
I need to add documentation for the SOMA test group.

compass/ocean/tests/soma/default/analysis.py
```diff
@@ -0,0 +1,90 @@
...
+        drho = section.getfloat('density_difference')
+        t0 = section.getfloat('surface_temperature')
+        s0 = section.getfloat('surface_salinity')
+
+        # hard coded factors from namelist file
+        gamma = 1.0 / 1250.0  # psu /m
+        beta = 0.05
+        zt = 300  # m
+        hshelf = 100  # m
+        h0 = 2400  # m
```
xylar, line 64: These would ideally not be hard-coded but instead would come from config options

xylar:
Note to self: make it easier to turn off particles.

compass/ocean/tests/soma/default/namelist.init
```diff
@@ -0,0 +1,2 @@
+config_vert_levels = 100
+config_vertical_grid = '100layerE3SMv1'
```
xylar, line 2:
```suggestion
config_vert_levels = 60
config_vertical_grid = '60layerPHC'
```
Maybe this is better?

compass/ocean/tests/soma/namelist.init
```diff
@@ -0,0 +1,5 @@
+config_init_configuration = 'soma'
+config_ocean_run_mode = 'init'
+config_realistic_coriolis_parameter = .true.
+config_use_bulk_wind_stress = .true.
+config_realistic_coriolis_parameter = .true.
```
xylar, line 5: Redundant -- remove

compass/ocean/tests/soma/default/namelist.init
xylar: Maybe make this change in a follow-up PR?

compass/ocean/tests/soma/namelist.forward
```diff
@@ -0,0 +1,27 @@
...
+config_use_cvmix_shear = .true.
+config_cvmix_shear_mixing_scheme = 'KPP'
+config_use_cvmix_kpp = .true.
+config_use_bulk_wind_stress = .true.
+config_density0 = 1014.65
+config_eos_type = 'jm'
+config_eos_linear_alpha = 0.255
+config_eos_linear_beta = 0.764
+config_eos_linear_Tref = 19.0
+config_eos_linear_Sref = 35.0
```
xylar, line 19: It's odd that we're using JM but setting parameters for the linear EOS.  Also, the linear `alpha` here doesn't match the one used in the IC.

compass/ocean/tests/soma/forward.py
```diff
@@ -0,0 +1,122 @@
...
+                      '8km': {'cores': 4,
+                              'min_cores': 1,
+                              'dt': "'00:10:00'",
+                              'btr_dt': "'0000_00:00:15'",
+                              'mom_del4': "2.0e9"},
+                      '4km': {'cores': 4,
+                              'min_cores': 1,
+                              'dt': "'00:06:00'",
+                              'btr_dt': "'0000_00:00:10'",
+                              'mom_del4': "4.0e8"}}
```
xylar, line 63: We need to modify these.  Based on ~300 cells for the target number of cores and ~3000 for the minimum, we would have:
```suggestion
        res_params = {'32km': {'cores': 25,
                               'min_cores': 3,
                               'dt': "'00:10:00'",
                               'btr_dt': "'0000_00:00:25'",
                               'mom_del4': "2.0e11"},
                      '16km': {'cores': 100,
                               'min_cores': 10,
                               'dt': "'00:10:00'",
                               'btr_dt': "'0000_00:00:25'",
                               'mom_del4': "2.0e10 "},
                      '8km': {'cores': 400,
                              'min_cores': 40,
                              'dt': "'00:10:00'",
                              'btr_dt': "'0000_00:00:15'",
                              'mom_del4': "2.0e9"},
                      '4km': {'cores': 1600,
                              'min_cores': 160,
                              'dt': "'00:06:00'",
                              'btr_dt': "'0000_00:00:10'",
                              'mom_del4': "4.0e8"}}
```

xylar: These should also be made config options instead of just hard coded.

compass/ocean/tests/soma/initial_state.py
```diff
@@ -0,0 +1,106 @@
...
+
+        Parameters
+        ----------
+        test_case : compass.TestCase
+            The test case this step belongs to
+
+        resolution : str
+            The resolution of the test case
+        """
+        super().__init__(test_case=test_case, name='initial_state')
```
xylar, line 32: This is implicitly defaulting to running on one core.  We may want to run on multiple cores, particularly at higher res.

compass/ocean/tests/soma/default/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+
+        step = InitialState(test_case=self, resolution=resolution)
+        # add the namelist options specific ot this test case
+        package = 'compass.ocean.tests.soma.default'
+        for out_name in ['namelist_mark_land.ocean', 'namelist.ocean']:
+            step.add_namelist_file(package, 'namelist.init', mode='init',
+                                   out_name=out_name)
+        self.add_step(step)
+
+        with_particles = self.resolution == '32km'
```
xylar, line 45: This should be the default but we want a config option for turning particles on or off at all resolutions.

compass/ocean/tests/soma/namelist.forward
```diff
@@ -0,0 +1,27 @@
+config_ocean_run_mode = 'forward'
+config_run_duration = '0000_00:01:30'
```
xylar, line 2: `dt` is at least 6 minutes, usually 10 minutes.  This should be updated to be several time steps.

compass/ocean/tests/soma/forward.py
xylar: It would make sense to have `dt` be approximately proportional to the grid spacing.

compass/ocean/tests/soma/default/namelist.init
xylar: I think this should be
```suggestion
config_vert_levels = -1
config_vertical_grid = '60layerPHC'
config_soma_vert_levels = 60
```
I'll take a more careful look that the code to be sure.

ytakano3: @xylar Just to note (for myself) but as we talked, config_vert_levels = 60 did not work, did not change the information in the namelist so need to set config_coma_vert_levels = 60 as you suggested. This worked.

mark-petersen:
@xylar I would like to preserve the original SOMA settings in the new repo in the history, and see you did that in 2f884182bf9f35ec1c697ce32efda113c2d13e97. As long as you keep that, you can proceed from there with additional commits, we can see the diffs, and you can add changes that make sense for your current setup in this PR. Since I have no personal experience with the settings, go ahead and change the settings above as you see fit.

xylar:
>  would like to preserve the original SOMA settings in the new repo in the history, and see you did that in 2f88418.
@mark-petersen, thanks for this comment.  In that commit "initial state" refers to the `initial_state` step, not the full test case.

I could squash some commits so that there is one that is a straight-up port of SOMA like you were expecting and then modify from there.  I think that's a good idea.

compass/ocean/tests/soma/default/__init__.py
xylar: It turns out to be difficult to make this a config option. Primarily, this is because the analysis step shouldn't happen if you're not using particles.  So I've made 2 versions of the test case at 32km resolution instead, one with and one without particles.  I think that's okay for now and we could add particles at other resolutions if we want to.

xylar:
I re-tested the 32km test case with and without particles after the many changes I just made and they both run fine.  The analysis plots look the same as before.

compass/ocean/tests/soma/streams.analysis
```diff
@@ -0,0 +1,83 @@
...
+    <var name="buoyancySurfaceVelocityZonal"/>
+    <var name="buoyancySurfaceDepth"/>
+    <var name="particleTemperature"/>
+    <var name="particleSalinity"/>
+</stream>
+
+<stream name="lagrPartTrackRestart"/>
+
+<stream name="mixedLayerDepthsOutput"
+        precision="double"/>
```
xylar, line 81: We don't want to be writing this out, at least not every time step.

```suggestion
<stream name="mixedLayerDepthsOutput"
        precision="double"/
        output_interval="none"/>
```

xylar:
## Longer tests (2 years)

I'm also going to run all resolutions (without particles) for 2 years.
- [x] `soma/32km/default` on Chrysalis with intel/impi
- [ ] `soma/16km/default` on Chrysalis with intel/impi
- [ ] `soma/8km/default` on Chrysalis with intel/impi
- [ ] `soma/4km/default` on Chrysalis with intel/impi

I'll put relative vorticity plots here as they are done.

32 km:
rel_vort_2yrs

xylar:
@mark-petersen, while I'm running some longer tests, I think this is ready to review.

xylar:
@vanroekel, please feel free to review this, too, if you like and have time.  I won't add you but feel free to add yourself.

xylar:
@ytakano3, you had asked me how long you would need to run this simulation to reach quasi-equilibrium.  It seems like 2 years is not enough and you might need to run for 3 or 4.

You also asked me how long the simulation might take.  The answer so far (for 2 year runs) was about 1 hour 15 min for 32 km; 3 hours 30 minutes for 16 km.  I did a little optimization to reduce the output that might speed things up some.  I haven't finished the 8 or 4 km runs yet, but they're taking awhile even on a lot of cores.

xylar:
Here's a movie at 16 km so you can see what I mean about not quite being equilibrated:
https://drive.google.com/file/d/1cq5kA0cuTF8ck1D0Ktc3jBUfr9q6i7Ha/view?usp=sharing

xylar:
I rebased onto the latest `master` to get the updated E3SM-Project submodule (and other updates).

I also added a set of 3-year `long` test cases at each resolution.  I'm retesting with these runs now on Chrysalis (with intel and impi).

xylar:
@mark-petersen, I believe you haven't had a chance to test this PR yet.  I hope that's the case because I ported the remaining test cases today.  I was trying to see if this might be a good project for a student but it turned out not to be.  It required a lot of re-engineering for good code reuse.

There is now a single `SomaTestCase` class that can be used to create all SOMA test cases.  This class supports:
* different resolutions (4km, 8km, 16km and 32km)
* with or without particles
* 3 or 60 layers
* with or without surface restoring
* a short (3 time-step) or long (3 year) simulation time

For now, 4 variants are defined at most resolutions (`default`, `three_layer`, `surface_restoring` and `long`), with a 5th variant at 32km resolution that includes `particles`.  It would be very easy to add other variants if they turn out to be useful (e.g. `long_with_particles` or `three_layer_with_surface_restoring`) to SOMA's `__init__.py` module in the future.

It's up to you how you want to review this.  One option would be to set up the `legacy` versions of the test cases and compare the namelists and streams files with these test cases.  That's a little tricky because they are purposefully different in some cases (e.g. 60 layers instead of 100) but we could talk those through.

xylar:
@ytakano3, for your review, I think the best test would be if you could run the `ocean/soma/16km/long` test case for the full 3 years.  If you want to do that in one go, you can ask for 1 or 2 nodes on Chrysalis (or 2 or 3 on Anvil) for 24 hours.  If you want to run in shorter increments, you will need to change the following namelist options in order to restart the run:
```
&time_management
    config_do_restart = .true.
    ...
    config_start_time = 'file'
    ...
```

xylar:
Following up on https://github.com/MPAS-Dev/compass/pull/273#issuecomment-974444026, here is a movie at 4km resolution, showing that 3 years does seem to be long enough to reach quasi-equilibrium:
https://drive.google.com/file/d/1GHGwdefY9tJuSAcCiZAOB-cYmgL0Tf-A/view?usp=sharing

mark-petersen (commented):
@xylar thanks for all your work here. It is an impressive addition. I tested on cori with:
```
compass list|grep soma
 118: ocean/soma/4km/default
 119: ocean/soma/4km/long
 120: ocean/soma/4km/surface_restoring
 121: ocean/soma/4km/three_layer
 122: ocean/soma/8km/default
 123: ocean/soma/8km/long
 124: ocean/soma/8km/surface_restoring
 125: ocean/soma/8km/three_layer
 126: ocean/soma/16km/default
 127: ocean/soma/16km/long
 128: ocean/soma/16km/surface_restoring
 129: ocean/soma/16km/three_layer
 130: ocean/soma/32km/default
 131: ocean/soma/32km/long
 132: ocean/soma/32km/particles
 133: ocean/soma/32km/surface_restoring
 134: ocean/soma/32km/three_layer

compass setup -w $n/211206_test_soma_PR273 -n 130 131 132 133 134 -p $e/master/components/mpas-ocean/
```
I ran all the 32km cases on a single node on cori successfully, so I trust the mechanics work. I also looked through @xylar's comments above. Visually, things look right.

mark-petersen (approved):
I looked through the code, and I'm approving based on your testing, which looks great. I think if we find small flag settings to change in the next few months, it's best to just make another PR.  Thanks!

xylar:
I have rebased this branch and updated the documentation to include the 4 other variants of the test case besides `default`.  In the process of documenting `three_layer`, I realized that I missed a step in which 2 config options need to be set so that the continental shelf is removed.  I have fixed that and will re-test the 4 `three_layer` test cases to make sure they look right.

xylar:
Indeed, all `three_layer` test cases still run sucessfully on Chrysalis (with Intel and Intel-MPI).

The domain (`4km/three_layer` in blue) is smaller than for the other test cases (`4km/long` in white):
three_layer_vs_long

xylar:
@mark-petersen, I assume you don't feel the need to revise your review based on these small changes?

mark-petersen:
no, please proceed. Thanks.

ytakano3:
@xylar I finally managed to run SOMA 16km, long run and I will attach the relative vorticity figure from the year 3. I could not find the same color scale as yours but the range is the same. I think the structure looks similar and the simulation is working well (figure attached at the bottom). As you know, I had minor issue compiling but that is the character encoding setting on local machine and manage to change settings on cluster (where I ran SOMA). 

The other error is Low-level I/O error towards the end of the simulation but this does not impact the results as we communicate. SOMA 32km (short) test also looks good. Thank you!
vtk_16km_year3

xylar:
Thanks @ytakano3! I appreciate you taking the time to run the model.

Just to document it, the errors @ytakano3 saw are:
```
PIO: ERROR: Changing the define mode for file (analysis_members/okuboWeiss.0003-01-01_00.00.00.nc) failed. Low-level I/O library API failed. NetCDF: Operation not allowed in data mode (error num=-38), (/gpfs/fs1/home/ac.xylar/mpas-work/compass/compass/conda/build_chrysalis_intel_impi/scorpio/src/clib/pioc_support.c:3377)
PIO: ERROR: Changing the define mode for file (analysis_members/highFrequencyOutput.0003-01-01.nc) failed. Low-level I/O library API failed. NetCDF: Operation not allowed in data mode (error num=-38), (/gpfs/fs1/home/ac.xylar/mpas-work/compass/compass/conda/build_chrysalis_intel_impi/scorpio/src/clib/pioc_support.c:3377)
PIO: ERROR: Changing the define mode for file (analysis_members/eliassenPalm.0003-01-01.nc) failed. Low-level I/O library API failed. NetCDF: Operation not allowed in data mode (error num=-38), (/gpfs/fs1/home/ac.xylar/mpas-work/compass/compass/conda/build_chrysalis_intel_impi/scorpio/src/clib/pioc_support.c:3377)
PIO: ERROR: Changing the define mode for file (analysis_members/mixedLayerDepths.0003-01-01.nc) failed. Low-level I/O library API failed. NetCDF: Operation not allowed in data mode (error num=-38), (/gpfs/fs1/home/ac.xylar/mpas-work/compass/compass/conda/build_chrysalis_intel_impi/scorpio/src/clib/pioc_support.c:3377)
PIO: ERROR: Changing the define mode for file (output/output.0003-01-01_00.00.00.nc) failed. Low-level I/O library API failed. NetCDF: Operation not allowed in data mode (error num=-38), (/gpfs/fs1/home/ac.xylar/mpas-work/compass/compass/conda/build_chrysalis_intel_impi/scorpio/src/clib/pioc_support.c:3377)
```
While I don't know exactly what the cause is, I have seen similar error messages and the output has still been okay.

@ytakano3, I don't know if you have permission, but the usual procedure to "officially" approve the PR would be to go to the "Files changes" tab at the top right, click on the green "Review changes", and enter your review.  That might be worth trying next time.

PR #274, opened 2021-11-17
Fix machine name arg in conda configure script

xylar:
Before this fix, specifying a machine name to the configure script in fact resulted in `machine = None`, which is obviously not the desired behavior.

This likely went unnoticed for 2+ months because we typically let the machine name be discovered automatically.

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -622,6 +622,8 @@ def main():
     if not args.env_only:
         if args.machine is None:
             machine = discover_machine()
+        else:
+            machine = args.machine
         if machine is not None:
             machine_info = MachineInfo(machine=machine)
 


xylar:
@matthewhoffman, please let me know if this fix works for you when building an environment for Cori-Haswell with gnu.  It worked for me.  Please merge if/when you're happy with it.

Very sorry about the bug and especially that I didn't do any testing that would have exposed this.

matthewhoffman:
@xylar , no worries and thanks for fixing this so quickly.  Cori was down yesterday, but I tested this branch this morning and everything worked.  When I load the Cori-haswell gnu load script I see a couple warnings that I think aren't critical but may need to be addressed eventually:
```
Installing collected packages: compass
  Attempting uninstall: compass
    Found existing installation: compass 1.0.0
    Uninstalling compass-1.0.0:
      Successfully uninstalled compass-1.0.0
  Running setup.py develop for compass
Successfully installed compass-1.0.0
craype-hugepages2M(43):ERROR:102: Tcl command execution failed: set             CRAYPE_DIR      $env(CRAYPE_DIR)

Module Deprecation Warning: upgrade to 3.21.3 (new default on 2021-09-21)
```

I approve of this PR unless you think we should try to resolve that hugepages error and module deprecation warning.

matthewhoffman (approved):

xylar:
> When I load the Cori-haswell gnu load script I see a couple warnings that I think aren't critical but may need to be addressed eventually

This is something to do with module switching.  The module commands come directly from E3SM so I don't know why E3SM doesn't also get these warnings (maybe it just suppresses them?).  It has never caused trouble as far as I can tell.  E3SM-Unified has the same warnings.  But I'm very much open to fixing this if there's a plausible way to do this while still getting the module commands directly from E3SM.

xylar:
@matthewhoffman, thank you for the review!

matthewhoffman:
I think it makes sense to leave this alone if it's coming from the E3SM module selections.  Presumably the machine POC will fix them if they become a problem.

PR #280, opened 2021-12-04
Set the cores per node from Slurm command

xylar:
Rather than relying on the `mache` config file, get the number of cores per node from Slurm.  Warn users if this number differs from what `mache` suggests.

### DIFF
--- a/compass/parallel.py
+++ b/compass/parallel.py
@@ -1,6 +1,7 @@
 import os
 import multiprocessing
 import subprocess
+import warnings
 
 
 def get_available_cores_and_nodes(config):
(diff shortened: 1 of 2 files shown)


xylar:
I tested this with the `cosine_bell` test case on Chrysalis (with Intel and Intel-MPI).  I saw the expected warning:
```
/gpfs/fs1/home/ac.xylar/mpas-work/compass/get_cpus_per_node_from_slurm/compass/parallel.py:90: UserWarning: Slurm found 128 cpus per node but config from mache was 64
```
in the log file for the test case (because I was running the `custom` test suite) and I got the expected PIO tasks and stride on 4 nodes and 512 cpus (for the `QU60/forward` step):
```
    config_pio_num_iotasks = 4
    config_pio_stride = 128
```

PR #284, opened 2021-12-07
Check parallel file system before sinfo call

xylar:
This will produce a more helpful error message if a user tries to run on a login node.


xylar:
## Testing

I tested this for a test suite, test case and step on a Chrysalis login node.  In each case, I got the expected output:
```
SLURM_JOB_ID not defined.  You are likely not on a compute node.
```

PR #287, opened 2021-12-11
Add support for compass test release

xylar:
Move the config options related to deployment of the conda env. to the conda directory.

Make sure the shared environment gets updated and mamba/boa installed.


xylar:
## Testing

I did test releases using this branch on all supported machines with supported combinations of compilers and MPI libraries.  For some machines, I used a local build of `mache 1.3.1` (since that has fixes for Grizzly, Cori-KNL and Chrysalis).

PR #292, opened 2021-12-12
Remove `grizzly`

### LINKED ISSUE #290: Remove Grizzly from compass machines
It will no longer be supported by LANL IC.

xylar:
This machine is no longer supported at LANL IC.

closes #290


xylar:
@mark-petersen, I'm just looking for a quick review by inspection here.  I searched the code and docs for "grizzly" and the only remaining reference is a design document, which we do not try to update.

I have not tried to run any tests yet with this code because I do not see how it would cause any trouble but obviously I would run tests before the 1.0.0 release.

xylar:
The corresponding PR for `mache` is here: https://github.com/E3SM-Project/mache/pull/19

mark-petersen:
Grizzly will be decommissioned sometime in 2022, but this has not happened yet. People with allocations >2M CPU-hours still have access to submit jobs, including some COSIM people. We should close this PR, and reopen when grizzly is decommissioned.

PR #294, opened 2021-12-12
Update User's Guide for 1.0.0 release

xylar:
Changes needed to bring the User's Guide up to date before the release.

The `load_latest_compass*.sh` script won't exist until the release happens and deployment takes place.  For now, the `test_compass_*.sh` activation scripts are the stand-ins for these scripts.  

To install `compass` on an unknown machine, you will need to use the following:
```
conda create -n compass -c conda-forge -c e3sm/label/compass -c e3sm/label/compass_dev python=3.9 "compass=1.0.0rc1=mpi_mpich*"
```
instead of the instructions in the user's guide.  This is because the `compass=1.0.0` packages haven't been created or uploaded yet.

Other than that, the user's guide should work as described.


xylar:
@matthewhoffman and @trhille  I realize the timing is terrible.  Even so, would you be willing to either make edits to the User's Guide to make sure it is useful for MALI users or alternatively, at various points, link MALI users to the appropriate equivalent instructions on the MALI-Dev wiki?

xylar:
This PR shouldn't be merged until after #292 and a rebase.

mark-petersen:
Grizzly will be decommissioned sometime in 2022, but this has not happened yet. People with allocations >2M CPU-hours still have access to submit jobs, including some COSIM people. So best to leave grizzly in. Otherwise, all these changes look good.

xylar:
Okay, I'll do my best to put grizzly back without losing my other changes.

matthewhoffman:
@xylar , sorry for the delay last year on looking at this.  I skimmed through the User's Guide again, and didn't see anything particular to MALI that needed changing.  The one thing that would be different is the instructions for compiling, but the only instructions I could find on that are in the Developer's Guide here:
https://mpas-dev.github.io/compass/latest/developers_guide/quick_start.html#building-mpas-components

Is that correct?

xylar:
@matthewhoffman, that's a good point.  We at least need to point to those instructions from the User's Guide.  Because folks will definitely need to build MPAS components.

xylar:
Okay, I put Grizzly back.  I still need to take care of the issue @matthewhoffman brought up, though.

xylar:
Okay, my last commit should add a little section to the User's Guide pointing over to the Developer's Guide for building MPAS components.  @matthewhoffman and @mark-petersen, let me know if you notice anything else.  But no big rush.

docs/users_guide/machines/badger.rst
```diff
@@ -11,8 +11,9 @@ login: ``ssh -t $my_username@wtrw.lanl.gov ssh ba-fe``
 config options
 --------------
 
-Here are the default config options added when you choose ``-m badger`` when
-setting up test cases or a test suite:
+Here are the default config options added when when Badger is automatically
```
matthewhoffman, line 14:
```suggestion
Here are the default config options added when Badger is automatically
```

docs/users_guide/quick_start.rst
```diff
@@ -58,25 +76,37 @@ miniconda is not already installed), then add the
...
 follows:
 
 .. code-block:: bash
 
-    conda create -n compass python compass
+    conda create -n compass -c conda-forge -c e3sm/label/compass python=3.9 "compass=*=mpi_mpich*"
+
+This will install the version of the package with MPI from conda-forge's MPICH
+package.  If you want OpenMPI, use ``"compass=*=mpi_openmpi*"`` instead.  If
+you do not want MPI from conda-froge (e.g. because you are working with a
```
matthewhoffman, line 88:
```suggestion
you do not want MPI from conda-forge (e.g. because you are working with a
```

matthewhoffman (approved):
Hi @xylar , I left a couple typo suggestions, and I also pushed a commit about compiling MALI.  It's kind of a punt, because it doesn't directly address compiling MALI with Albany, but that is still an evolving target, so I was thinking we could add stable details about that later.  Let me know if you think that's ok.

xylar:
@mark-petersen, still looking for your review on this.

mark-petersen:
@xylar thanks for pinging me. This documentation is an impressive piece of work. I decided to review by looking up some questions that I've had lately about compass.

Under `DEVELOPER'S GUIDE/Command-line interface/compass run`:
1. How do you know the steps available, that can be written in as `STEPS`? Do you look at the subdirectories within that directory, or is there some compass command to use that lists the steps?
2. When I am in the lowest level directory (e.g. `initial_state` or `forward`) I would like to know exactly what `compass run` is doing, for example the `srun` or any other commands. How do I do that? I think you've told me, but I couldn't find it, but this section would be a good place to say that. I miss the scripts in `compass/legacy` that specified this, because `compass run` seems more opaque. But most likely I just don't understand how to retrieve that information.

Under `DEVELOPER'S GUIDE/Quick Start for Developers`
1. If I alter something in a compass development branch on disk, do I need to rerun `./conda/configure_compass_env.py` every time I make an alteration? My assumption is that if I change python code then I need to rerun it, but if I change a `config_` line in a `namelist.*` file I don't. 
2. If I am switching between different compass branches, I have to be sure to source the `load_dev_compass*` in that branch, and then you can run compass commands anywhere, and it always refers to that branch, right? Is there a way for compass to tell us what directory it was loaded from, or we just need to be careful to track that ourselves?

@xylar you can decide if these questions deserve an answer in the documentation. The docs have a very clean feel right now.

xylar:
@mark-petersen, thanks for the suggestions.  These seem important to address.

>  When I am in the lowest level directory (e.g. `initial_state` or `forward`) I would like to know exactly what `compass run` is doing, for example the `srun` or any other commands. How do I do that? I think you've told me, but I couldn't find it, and this section would be a good place to say that. I miss the scripts in `compass/legacy` that specified this, because `compass run` seems more opaque. But most likely I just don't understand how to retrieve that information.

Hmm, I don't think I'm going to be able to address this one to your satisfaction.  What we have is the info that goes into the log files.  It doesn't look like the `srun` (or `mpirun`) commands are getting written out, and we should add those.  But I gather this isn't what you are after.

Short of that, there is no useful way to provide a step-by-step similar to the script in `legacy`.  I really have wanted to get away from code-writing code for one and for another that method doesn't lend itself to clean code reuse.  What you have is the `run()` method (function) of each test case class and and its step classes.  That is what make up the "script" you used to have.  This is not information that the current compass makes accessible to "users" in the work directory and I don't see an easy way to do that.

I am certainly open to having a discussion about how to make this easier to follow.

xylar:
@mark-petersen, I just added changes to address 3 or your 4 requests.

Regarding logging of `srun` commands, I need to make a change in MPAS-Tools for that and I'm working on testing it now.  If just logging the `srun` commands isn't sufficient, we'll need to talk about another solution.

xylar:
@mark-petersen, another possibility in terms of logging of `compass run` would be to write out which `run()` methods are getting called.  I will have to look into how to do this from the framework, since it would be too clumsy to put this in every single test case.  I think it's a good idea.

docs/developers_guide/quick_start.rst
```diff
@@ -44,6 +44,14 @@ you're working on.
 The ``conda`` directory in the repository has a tool ``configure_compass_env.py``
 that can get you started.
 
+You will need to run ``./conda/configure_compass_env.py`` each time you check
+out a new branch or create a new worktree with ``git``.  Typically, you will
+*not* need to run this command when you make changes to files within the
+``compass`` python package.  These will automatically be recognized because
+``compass`` is installed into the conda environment in "editable" mode.  You
+*will* need to run the command if you add new code or data files to the
```
mark-petersen, line 52: Wow, that's good to know. I would change this to "new code files or data files" to be clear, as my experience with compilers led me to think that any code change required a rerun of `./conda/configure_compass_env.py`

xylar: There aren't any compilers in python, and the "editable" install is a handy way not to need to reinstall the code after every change.  I didn't know about that feature until relatively recently (about a year ago).

I made the change you requested ("code or data files" --> "code files or data files").  I agree, that's clearer.

mark-petersen (approved):
This is great, thanks!

xylar:
Thanks again!

Issue #298, opened 2022-01-11
60 layer PHC interface locations incorrect in MPAS-Ocean init mode


xylar:
For nearly 7 years, the interface locations in MPAS-Ocean init mode have been incorrect:
https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/mode_init/mpas_ocn_init_vertical_grids.F#L165-L225
Except for the first interface value, these are the locations of the *centers* of layers, not the interfaces.

The correct interface locations are available in compass:
https://github.com/MPAS-Dev/compass/blob/master/compass/ocean/vertical/60layerPHC.json

Fortunately, `60layerPHC` has not been used very much.  For example, the global initial conditions get their vertical coordinate from the PHC data files, not from this piece of code.

The only test case from legacy COMPASS that is affected is:
https://github.com/MPAS-Dev/compass/blob/legacy/ocean/Gaussian_hump/USDEQU120cr10rr2/delaware/config_init.xml#L12

In the current `compass`, the only affected test case is the new SOMA, which was switched from the 100-layer vertical coordinate to a 60-layer coordinate as part of the port:
https://github.com/MPAS-Dev/compass/blob/master/compass/ocean/tests/soma/initial_state.py#L86-L87

This should be fixed in E3SM, but I'm documenting it here for now because it primarily (exclusively?) impacts compass.

xylar:
Fixed in https://github.com/E3SM-Project/E3SM/pull/4729, but I will keep this open until we update the `E3SM-Project` submodule to bring in that fix.

xylar:
Closed via #295

PR #299, opened 2022-01-11
Switch SOMA init to python

xylar:
The initial conditions and forcing are computed in python instead of in init mode in MPAS-Ocean.  This should make it easier to generalize the test case, e.g. to include BGC tracers.


xylar:
The initial condition is not expected to be bit-for-bit because of this bug: #298

Even with that bug fixed, the way the SOMA init-mode code handles bottom cells seems incorrect (or at least undesirable) in that it is only moving `bottomDepth` up to the nearest interface, rather than rounding either up or down to the nearest interface location as `compass` is doing here.  As a result, `maxLevelCell <= 59` in SOMA init mode but `maxLevelCell <= 60`, as expected, here.

xylar:
I'll first fix #299, but then I'm not going to bother with bit-for-bit checks with the existing SOMA results, I'm just going to run some long simulations to make sure things look qualitatively similar to #273

xylar:
## Testing

I ran the `32km/long` and `16km/long` test cases, which run for 3 years.  Below are first the results from #273 and then the results from this branch:

old 32km:
image
new 32km:
rel_vort_32km

old 16km:
image
new 16km:


For the `32km/surface_restoring` test case, here are plots of the difference between the old and new temperature restoring and zonal wind stress fields, with differences at machine precision:

temp_rest_diff
zonal_wind_diff

The Coriolis parameter is also the same to machine precision.  Other fields, such as T and S, in the initial state are affected by the vertical grid and `bottomDepth` being handled differently in the python version vs. MPAS-Ocean's init mode, so differences are expected.

xylar:
@ytakano3 and @mark-petersen, this is ready for review.

The primary purpose of this port is so it will be easier for @ytakano3 to add BGC tracers to the initial condition and surface forcing.  In the meantime, @ytakano3 and @maltrud are working on modifications to MPAS-Ocean to allow the time-variable forcing that they would need for this work.

@mark-petersen, for your review, it would be very helpful to have another set of eyes checking if I ported https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/mode_init/mpas_ocn_init_soma.F into `initial_state.py` accurately.  See the discussion above about the vertical grid and how full cells are handled differently in https://github.com/MPAS-Dev/compass/blob/master/compass/ocean/vertical/__init__.py#L8 than in the SOMA Fortran code. It would also be helpful if you could check that `soma.cfg` contains the same options and values as https://github.com/E3SM-Project/E3SM/blob/master/components/mpas-ocean/src/mode_init/Registry_soma.xml.  I don't think there's any need for you to run tests yourself but you can if you like. Let me know if you'd like to chat on Slack.

@ytakano3, for your review, I would like you to go through the new `initial_state.py` with me and make sure you understand how it works.  If you are willing to run the 8 km long test again (for the full 3 years), that would be really helpful.  Another helpful test would be to run the `surface_forcing` test at any resolution of your choice and compare the results (qualitatively, since they are definitely not identical) with results from the `compass /master` branch.  We can chat about this on Slack or in person next week.

xylar:
I'm taking @ytakano3 off of this because the Kuroshio simulations are going to be his priority going forward.

xylar:
Thanks @mark-petersen!  I'll test again as soon as I can.  I'm putting this work on hold until v1.1.0, though, because @ytakano3 will not be making refinements to SOMA after all.

xylar:
I'm closing this for now because it's a low priority

PR #275, opened 2021-11-18
Update E3SM-Project to latest master

xylar:
commit hash: 5fa795e1adbf94795445b81bdb290bee8a18ba57

Among other things, this update is intended to bring in the [addition of MARBL](https://github.com/E3SM-Project/E3SM/pull/4587) so that code can be used in compass.

PR #303, opened 2022-01-22
Set E3SM-Project submodule back to last update

xylar:
Set E3SM-Project back to hash `5fa795e1adbf94795445b81bdb290bee8a18ba57`.

This was update in #275 but inadvertently reverted in #273.


xylar:
@mark-petersen, could you give this a quick look when you can?  This is causing @ytakano3 trouble because MARBL went away when this got reverted by mistake.

mark-petersen (approved):
Yes, this is fine. I can see that this reverts to the hash before the inadvertent change in #273.

xylar:
Tanks @mark-petersen.

PR #305, opened 2022-01-24
Fix location for mpas-seaice initial condition

xylar:
The files are placed in `...ice/mpas-seaice` whereas they were previously in `...ocn/mpas-seaice`.

### DIFF
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/seaice_initial_condition.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/seaice_initial_condition.py
@@ -58,7 +58,7 @@ class SeaiceInitialCondition(Step):
             creation_date = ds.attrs['{}_Version_Creation_Date'.format(prefix)]
 
         try:
-            os.makedirs('../assembled_files/inputdata/ocn/mpas-seaice/{}'.format(
+            os.makedirs('../assembled_files/inputdata/ice/mpas-seaice/{}'.format(
                 mesh_short_name))
         except OSError:
             pass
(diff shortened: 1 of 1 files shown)


xylar:
The second commit here is just some clean-up to help with PEP8 and such.

xylar:
@mark-petersen, I'll run a full QU240 test (through `files_for_e3sm`) and ping you to review once I post the results.

xylar:
## Testing

I ran the QU240 mesh, init, performance test, dynamic adjustment and files-for-E3SM test cases, using the version of MPAS-Ocean from https://github.com/E3SM-Project/E3SM/pull/4750.

As expected, the resulting sea-ice initial condition is in:
```
assembled_files/inputdata/ice/mpas-seaice/QU240E2r1/mpassi.QU240E2r1.220124.nc
```
(It also contains `fCell` that is not all zeros.)

xylar:
@mark-petersen, this is ready to review.  You can just review by inspection if you like.  The first commit is the important one, the second is clean up.

xylar:
@mark-petersen, could you have a quick look at this when you have time?

mark-petersen (approved):
Yes, looks good. I've never seen an f-string, but I found it with a web search. That's handy!

xylar:
Thanks, @mark-petersen!  Yeah, f-strings are almost always more readable and have been around since early python 3, so they're supported in all versions of python that folks *should* be using (currently python >=3.7).

@altheaden, deserves the credit for pushing me to use them.

PR #307, opened 2022-01-27
Fix `config_mom_del2` for the WC14 mesh

xylar:
There was a typo in the name of the namelist option.


mark-petersen (approved):
Thanks. This is an obvious typo to correct. I included this in my testing, and it passes the nightly regression suite, gnu optimized and debug, and I checked that the flag comes through on the namelists. Also works fine for WC14.

xylar:
Thanks again, @mark-petersen!

Issue #308, opened 2022-01-30
SST in WC14 stand-alone spin-up warms to 60C


mark-petersen:
An unphysical warming to 60C on the Indonesian coast occurs in the WC14 stand-alone spin-up using compass/master 2270c038, even after fixing the Coriolis bug noted in #304. This can be seen with:
```
compass/ocean/tests/global_ocean/init/namelist.init
-config_expand_sphere = .false.
+config_expand_sphere = .true.

compass list|grep WC14
  95: ocean/global_ocean/WC14/mesh
  96: ocean/global_ocean/WC14/PHC/init
  97: ocean/global_ocean/WC14/PHC/performance_test
  98: ocean/global_ocean/WC14/PHC/dynamic_adjustment

compass setup -f Config_dir/e3sm_master -w $DIR -n 95 96 97 98
```
After running these steps on grizzly with `load_dev_compass_1.0.0_grizzly_gnu_mvapich.sh`, the error can be seen here:
```
cd /lustre/scratch4/turquoise/mpeterse/runs/220126_compass_master_E3SM_head_gr_gfortran_openmp_WC1402
cd ocean/global_ocean/WC14/PHC/dynamic_adjustment/simulation/analysis_members/

ncdump -v temperatureMax globalStats.*.nc |tail -n 2
    60.5092380065623, 60.5138480143581, 60.5184557681456, 60.5230611583427 ;
```
It reaches 60C by day 25 of the simulation.

mark-petersen:
I conducted a large number of tests, but I understand it now so can make a cleaner presentation. I started with:
1. WC14 on head of compass/legacy: OK, max remains at 30.1C
2. WC14 on head of compass/master with expand_sphere on: described above, max is 60.5

The namelist differences from (1) to (2) on the simulation step are:
```
<     config_mom_del2 = 400
>     config_mom_del2 = 1.0e3
<     config_use_activeTracers_surface_restoring = .true.
>     config_use_activeTracers_surface_restoring = .false.
```
The `config_mom_del2` was unintentional, and is fixed in #307. This exacerbates the problem, but is not the root cause of the 60C temperature. 

Turning off surface restoring was intentional, in #99. The surface restoring was the change that allowed the 60C problem to occur. But it was essentially 'covering up' some other mechanism that causes non-monotonic warming. To show this, I did the additional tests:

[...]

PR #309, opened 2022-01-31
Turn on surface restoring for all dynamic adjustment

xylar:
This merge re-enables surface restoring of tracers in the dynamic adjustment test cases from the global ocean test group. This choice was discussed at length in #308.

This merge also enables GM in the QU240 configuration to be (somewhat) more consistent with other resolutions.  QU240 also enables many features that are not used at other resolutions.

### DIFF
--- a/compass/ocean/tests/global_ocean/mesh/ec30to60/dynamic_adjustment/__init__.py
+++ b/compass/ocean/tests/global_ocean/mesh/ec30to60/dynamic_adjustment/__init__.py
@@ -48,9 +48,11 @@ class EC30to60DynamicAdjustment(DynamicAdjustment):
 
         module = self.__module__
 
-        global_stats = {'config_AM_globalStats_enable': '.true.',
-                        'config_AM_globalStats_compute_on_startup': '.true.',
-                        'config_AM_globalStats_write_on_startup': '.true.'}
+        shared_options = \
+            {'config_AM_globalStats_enable': '.true.',
+             'config_AM_globalStats_compute_on_startup': '.true.',
+             'config_AM_globalStats_write_on_startup': '.true.',
+             'config_use_activeTracers_surface_restoring': '.true.'}
 
         # first step
         step_name = 'damped_adjustment_1'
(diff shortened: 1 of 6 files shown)


xylar:
@mark-petersen, I have not yet had any time to test this.  I will post as soon as I do.

xylar:
@mark-petersen and @vanroekel, related to #310, we should revisit whether or not we want surface restoring.  @mark-petersen seemed to want it before but then seems to be changing his mind in recent discussions of results using the Redi fix in https://github.com/E3SM-Project/E3SM/pull/4785.

I'm fine either way but would like to know whether to close this or retest once https://github.com/E3SM-Project/E3SM/pull/4785 goes in.

mark-petersen:
@xylar I gave this some more thought, and I would still like to have surface restoring on, as it keeps spin-up closer to climatology and is more physical for any longer stand-alone simulations.

vanroekel:
For the short dynamic adjustment phase, I have no preference either way.  If we are spinning up for only a month or two in stand alone, I can't imagine it having a long term impact in E3SM on or off.    So I'm fine with going with it on as desired by @mark-petersen.

mark-petersen:
@xylar, I tested this PR with:
```
./conda/configure_compass_env.py --conda /usr/projects/climate/mpeterse/miconda3 --compiler gnu
source load_dev_compass_1.0.0_grizzly_gnu_mvapich.sh
compass suite -s -c ocean -t nightly -f Config_dir/master -w $WORKDIR
```
but it still shows
```
    config_use_activeTracers_surface_restoring = .false.
```
in all the `global_ocean` cases. I might have made a mistake, but this is the normal process I use. @xylar can you double-check that your test changes this flag?

xylar:
@mark-petersen, thanks.  I'll check this.

xylar:
@mark-petersen, I don't think we run any `dynamic_adjustment` tests in either `nightly` or `pr` test suites.  I didn't turn on surface restoring for performance tests etc. It's not clear to me why that would be important to do.

xylar:
I'm seeing `config_use_activeTracers_surface_restoring = .true.` as expected within the steps in `dynamic_adjustment` where I looked.

mark-petersen (approved):
OK, that was just my misunderstanding about which case we turn it on for. Thanks, this looks good.

xylar:
Tested along with #310.

vanroekel (approved):
approving based on visual inspection and testing by @xylar and @mark-petersen

PR #310, opened 2022-02-03
Add bathymetry smoothing to global ocean init

xylar:
At @mark-petersen's request, we perform 6 iterations of smoothing with a weight of 0.9.

For more info on the smoothing approach, see:
https://github.com/MPAS-Dev/MPAS-Model/pull/440
For more on some slightly more aggressive values used previously (related to ice-shelf cavities), see:
https://github.com/MPAS-Dev/MPAS-Model/pull/532#issuecomment-620760208

Partially addresses #308

### DIFF
--- a/compass/ocean/tests/global_ocean/init/namelist.init
+++ b/compass/ocean/tests/global_ocean/init/namelist.init
@@ -63,3 +63,5 @@ config_global_ocean_depth_conversion_factor = 1.0
 config_global_ocean_minimum_depth = 10
 config_global_ocean_deepen_critical_passages = .false.
 config_block_decomp_file_prefix = 'graph.info.part.'
+config_global_ocean_topography_smooth_iterations = 6
+config_global_ocean_topography_smooth_weight = 0.9


xylar:
## Testing

I am running on Anvil with Intel and Intel-MPI through `dynamic_adjustment` with this PR and #309.
- [x] `QU240` max temperature: 29.8256095996756
- [x] `QUwIsC240` max temperature: 29.8256253687571
- [ ] `EC30to60` max temperature:
- [ ] `ECwISC30to60` max temperature:
- [ ] `SOwISC12to60` max temperature:
- [ ] `WC14` max temperature:

xylar:
@vanroekel and @mark-petersen, following a successful merge of https://github.com/E3SM-Project/E3SM/pull/4785 (in the likely event that occurs), I'd like to revisit bathymetry smoothing before the compass 1.0.0 release.

I would like to have consistent bathymetry and smoothing techniques across `global_ocean` meshes unless there is a compelling reason to deviate.  The trouble case seems to be SORRM (it requires the most smoothing to be well behaved under ice shelves).  Since it has been a long time since the values used here were chosen (https://github.com/MPAS-Dev/MPAS-Model/pull/532#issuecomment-620760208), we could see if we can back off (say, to 6 iterations instead of 10).  Or we could stick with these values for now and revisit this when we are actually working on new production meshes for E3SMv3 or some intermediate project.

I think this last option would be my preference.  But we definitely do need to turn on some level of smoothing or I can't test the SORRM infrastructure at all.

mark-petersen:
@xylar, I agree that no smoothing leads to very rough bathymetry, and that was never the intention. I think that 10 iterations is too much, but setting it to a mid-value between 4 and 6 is fine for now, so that all the COMPASS cases produce reasonable meshes. As @xylar said, when we sign off on our next production mesh, we should run some tests on that value.

vanroekel:
@xylar I agree that the last option seems best for now -- stick with current values and revisit for v3 production meshes.  I also agree that we want some level of smoothing in all meshes going forward and we should be consistent amongst them all.

xylar:
@mark-petersen, I'm fine with backing off to 6 iterations with a weight of 0.9 if SORRM can be spun up with that value (it couldn't previously, which is how I ended up at 10 iterations with a weight of 0.92).  I do want to caution that the number of iterations means very little on its own.  The combination of the weight and the number of iterations determines how many effective cell widths the smoothing happens over.  But I'll change and retest, since it seems important to you.

mark-petersen:
@xylar, thanks for letting me know about the combination of number and weight for smoothing iterations. I'm happy to go with your best judgement on the combination of the two.

xylar:
I'd like to wait for https://github.com/E3SM-Project/E3SM/pull/4785 to go in and test with the final result that we end up with.  Then, I'll see if I can get away with 6 iterations of smoothing instead of 10, and experiment with the associate weight.

xylar:
@mark-petersen, in retesting this I'm seeing some discouraging results.  In the EC30to60 dynamic adjustment (during the second damped step), I see:
ec

Temperatures reach 51 C before damping back down.  We need to find out if this is still Redi or if it's something else (smoothing, restoring, not enough Rayleigh friction or too small a time step).

xylar:
I'm running dynamical adjustment for EC30to60 with `compass/master` and temperature is well behaved, so that's good news!

xylar:
The SORRM mesh is behaving perfectly well with this branch and #309, so I don't know what the deal is with EC.

xylar:
Just smoothing is definitely making things worse, with `temperatureMax = 42.437262634545` at the end of dynamic adjustment (and a maximum over the process of 52.6819303618259).

xylar:
Looking back at https://github.com/E3SM-Project/E3SM/pull/4785, I didn't test EC as part of that review.  I just can't believe I didn't test all the meshes!  That's just supremely negligent!

xylar:
The good news is that EC tests with `config_Redi_min_layers_diag_terms = 8` do seem to work fine (both with and without smoothing).  I'll try `config_Redi_min_layers_diag_terms = 7` tomorrow to see if that works as well.

milenaveneziani:
Do we understand why the SORRM is better behaved with a lower value of `config_Redi_min_layers_diag_terms`, considering that the SORRM resolution is the same as the EC30to60 at those latitudes? And SORRM also has 60 vertical levels, right?

xylar:
> Do we understand why the SORRM is better behaved with a lower value of config_Redi_min_layers_diag_terms, considering that the SORRM resolution is the same as the EC30to60 at those latitudes? And SORRM also has 60 vertical levels, right?

I don't think we know that SORRM is better behaved.  We only tested WC14 with `config_Redi_min_layers_diag_terms = 6`.  I can't do SORRM dynamic adjustment right now at all, since there's not enough smoothing of the topography to even get through the `init` test case.

milenaveneziani:
oh sorry, I thought you mentioned that the SORRM was well behaved above? maybe you meant WC14?

xylar:
> oh sorry, I thought you mentioned that the SORRM was well behaved above? maybe you meant WC14?

Oh, I see what you're referring to:

> The SORRM mesh is behaving perfectly well with this branch and https://github.com/MPAS-Dev/compass/pull/309, so I don't know what the deal is with EC.

I've tested so many things in the last few days that I completely lose track.  It could just be a coincidence of where the cells end up at different resolutions.  Or it could have to do with different settings.  I'm really not sure.  Since SORRM spin-up doesn't work without smoothing, it's hard to make a clean comparison but it's true that smoothing alone didn't prevent EC30to60 from going nuts.

mark-petersen:
@xylar, reading the comments above, it looks like once you set `config_Redi_min_layers_diag_terms=8` for EC60to30, then the  6 iterations of smoothing with a weight of 0.9 proposed in this PR is still a good choice. Is that correct?

xylar:
@mark-petersen, I'm about to re-test this and #309 on all the meshes.  But I think, yes, 6 iterations of smoothing with a weight of 0.9 will likely be fine for SORRM (which is the mesh that tends to have trouble if there isn't enough smoothing).  I'll keep you posted and ask for a review based on my testing later today or tomorrow.

vanroekel:
@xylar I'm happy to approve this, but have a very small question first in a comment you mention

> But I think, yes, 6 iterations of smoothing with a weight of 0.9 will likely be fine for SORRM (which is the mesh that tends to have trouble if there isn't enough smoothing)

From the files changed you seem to use 10 iterations though.  Is 10 the value you found necessary?  I just wanted to confirm.

mark-petersen:
@xylar I can't easily find the updated results above. Could you summarize? 

But generally, I think we are finding that we need to be aware that there is no silver bullet to these spin-up maxima when Redi is on, and be aware of it as we create and spin up new meshes. I think #328 is a huge help for preventing mistakes, and makes it much safer as we go forward.

xylar:
> From the files changed you seem to use 10 iterations though. Is 10 the value you found necessary?

@vanroekel, shoot!  Thanks for noticing that!  I swear I change that but it looks like my changes got lost in one rebase or another.  I'll fix that and rerun my tests tomorrow.

>  Could you summarize?

@mark-petersen, it's a little unclear to me what you're looking for in a summary.  I'll be rerunning anyway after I fix this branch, since @vanroekel pointed out that my changes got lost.  But the gist is that I am running `dynamic_adjustment` for each of the 6 ocean meshes we support using this branch and #309, and looking at the maximum temperature to see whether it gets worse (it does for EC meshes).

If you want some specific summary of what the smoothed meshes look like or something that, I'd be happy to oblige.  Maybe a comparison between smoothed and unsmoothed in some reason of interest?  I'm honestly a little exhausted by this whole process and not that keen on playing "bring me a rock".

mark-petersen:
Sorry, didn't mean to ask for extra busywork. I just couldn't find the post you meant by "my updated test results from above". If smoothing the bathymetry generally reduces the problem, then it seems like a good idea.

xylar:
@mark-petersen, sorry for the confusion.  I was referring to https://github.com/MPAS-Dev/compass/pull/310#issuecomment-1029546997, which I have been updating continuously.  I will make sure to post the path to my results on Anvil when I update that comment again today.

> If smoothing the bathymetry generally reduces the problem, then it seems like a good idea.

Unfortunately, smoothing the bathymetry seems to slightly aggravate the problem with temperature extrema.  It is necessary for other reasons -- meshes that include ice shelf cavities tend to blow up without any smoothing, typically during the adjustment of the `landIcePressure` field to match the `ssh`.  It is my feeling that we should be using the same smoothing algorithm and values on all meshes for consistency, and I also feel strongly that it isn't a good idea to have completely unsmoothed bathymetry like we have in the current `master`.

As a side note, I worked pretty hard to design a procedure for smoothing bathymetry before it is sampled to the MPAS mesh, and I know that @dengwirda has an alternative approach for accurately averaging the bathymetry over an MPAS cell.  Both of these approaches are clearly preferable to the approach used here -- smoothing after sampling to the MPAS mesh.  But these approaches are time consuming and require further investigation and discussion.  I think the approach used here will suffice in the interim.

vanroekel (approved):
This looks good @xylar and am happy to approve.  I agree with you that this approach is more than sufficient in the near term and that we should be consistent in the smoothing amongst all meshes (and that there should be smoothing).

xylar:
Thanks @vanroekel.  and thanks for approving #309 as well.  Much appreciated.

xylar:
Thanks @mark-petersen.  I'm still waiting for the WC14 test to finish (Anvil was busier than usual today).  I'll merge both this and #309 when that's done.

xylar:
Just a final note to say that, in my testing, `EC30to60`, `ECwISC30to60` and `SOwISC12to60` are all producing temperatures that exceed the threshold of 33 C that I set in #328. This will need to be addressed before we produce any new initial conditions for E3SM for any of these meshes.

PR #313, opened 2022-02-07
Constrain pyamg (instead of scipy)

xylar:
A recent update to pyamg fixes issues with the latest scipy.


xylar:
In a bit over an hour, the package should be available and CI should work with this update.

PR #315, opened 2022-02-07
Generalize flood-fill routine for use by multiple test groups

trhille:
Create module compass.landice.landice that will contain functions for general use in creating landice meshes, and add flood_fill() function. Also add usage example in Humboldt mesh creation step. The flood_fill() function will likely be used by all landice test groups that interpolate gridded observational data onto an MPAS mesh.

### DIFF
--- a/compass/landice/landice.py
+++ b/compass/landice/landice.py
@@ -1,0 +1,71 @@
+import numpy as np
+
+
+def flood_fill(thk, vx, vy):
+    """
+    Remove glaciers and ice-fields that are not connected to the ice sheet.
+
+    Parameters
+    ----------
+    thk : numpy.ndarray
+        Ice thickness from gridded dataset
+    vx : numpy.ndarry
+        Velocity x-component from gridded dataset
+    vy : numpy.ndarray
+        Velocity y-component from gridded dataset
+
+    Returns
+    -------
+    thkTrimmed : numpy.ndarray
+        Ice thickness after flood-fill is applied
+    vxTrimmed : numpy.ndarray
+        Velocity x-component after flood-fill is applied
+    vyTrimmed : numpy.ndarry
+        Velocity y-component after flood-fill is applied
+    """
+    sz = thk.shape
+    searchedMask = np.zeros(sz)
+    floodMask = np.zeros(sz)
+    iStart = sz[0] // 2
+    jStart = sz[1] // 2
+    floodMask[iStart, jStart] = 1
+
+    neighbors = np.array([[1, 0], [-1, 0], [0, 1], [0, -1]])
+
+    lastSearchList = np.ravel_multi_index([[iStart], [jStart]],
+                                          sz, order='F')
+
+    cnt = 0
+    while len(lastSearchList) > 0:
+        cnt += 1
+        newSearchList = np.array([], dtype='i')
+
+        for iii in range(len(lastSearchList)):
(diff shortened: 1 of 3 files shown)


compass/landice/landice.py
```diff
@@ -0,0 +1,71 @@
...
+    Remove glaciers and ice-fields that are not connected to the ice sheet.
+
+    Parameters
+    ----------
+    thk : numpy.ndarray
+        Ice thickness from gridded dataset
+    vx : numpy.ndarry
+        Velocity x-component from gridded dataset
+    vy : numpy.ndarray
+        Velocity y-component from gridded dataset
```
xylar, line 15: Could you explain in the docstring why velocity components are needed for the flood fill?  It might be obvious to you but it was not to me.  If the idea is to trip 3 arrays, I would generalize this to just take thickness and a list of arrays (or better yet an `xarray.Dataset` to trim.  That way, you don't have to generalize if you need other trimmed variables in the future.

xylar (changes requested):
This is really good to have generalized and pulled out into the framework!

In addition to my comment about maybe having a list of variables rather than specific ones to trim, my main comment is that `landice.py` is too generic a module name.  I would name it something like `mesh.py` or `connectivity.py` or something that leave room for a small number of other related framework functions in the future (but not every single piece of land-ice framework).

compass/landice/landice.py
trhille: Thanks for the suggestion. That's a much better approach.

trhille: I could also have the only input be thickness, and then have it return floodMask. Then the actual trimming would be performed after calling flood_fill(). That might be more flexible and a bit easier to understand.

xylar: I like that better!

trhille: Fixed in commit [52d2c97](https://github.com/MPAS-Dev/compass/pull/315/commits/52d2c97c6362d6bc8a4cadd647130174ce44fbb5)

xylar (approved):
Excellent, looks great to me!

compass/landice/mesh.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+
+
+def flood_fill(thk):
+    """
+    Remove glaciers and ice-fields that are not connected to the ice sheet.
```
matthewhoffman, line 6: I'm trying to avoid getting too semantic about this, but do want this to be a general flood fill?  In which case, you replace `thk` with `field` or something and it can be applied to other situations (but leave a note in the docstring that the typical use is for removing glaciers and ice-fields, etc.  On the hand, I'm not sure if there are any other uses I can imagine in COMPASS.  Also, in the docstring mention that this is on a regular grid and not and mpas grid.  Actually how about calling fhe method `gridded_flood_fill` or something?

And while you're at, you could mention there may be more efficient algorithms.

matthewhoffman (approved):
@trhille , thanks for taking the time to generalize this routine.  I feel like landice is finally starting to take advantage of the new COMPASS design!

trhille:
@xylar, is it okay to merge this with these failed checks? I don't think they are related to this PR.

xylar:
@trhille, in general, no, please only merge when checks work. In this particular case I know that there is something going wrong with Linux CI that is nothing to do with this branch so I will allow it this once.

PR #319, opened 2022-02-14
Add set_geom_points_and_edges() function

trhille:
Add set_geom_points_and_edges() function to compass.landice.mesh module, and call from build_cell_widths(). This will likely be used by all landice mesh-generating test cases.

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -63,7 +63,7 @@ def gridded_flood_fill(field):
     return floodMask
 
 
-def set_geom_points_and_edges(xmin, xmax, ymin, ymax):
+def set_rectangular_geom_points_and_edges(xmin, xmax, ymin, ymax):
     """
     Set node and edge coordinates to pass to
     :py:func:`mpas_tools.mesh.creation.build_mesh.build_planar_mesh()`.
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -63,20 +63,20 @@ def gridded_flood_fill(field):
     return floodMask
 
 
-def set_geom_points_and_edges(xx0, xx1, yy0, yy1):
+def set_geom_points_and_edges(xmin, xmax, ymin, ymax):
     """
-    Set node and edge coordinates to pass to 
+    Set node and edge coordinates to pass to
     :py:func:`mpas_tools.mesh.creation.build_mesh.build_planar_mesh()`.
 
     Parameters
     ----------
-    xx0 : int or float
+    xmin : int or float
         Left-most x-coordinate in region to mesh
-    xx1 : int or float
+    xmax : int or float
         Right-most x-coordinate in region to mesh
-    yy0 : int or float
+    ymin : int or float
         Bottom-most y-coordinate in region to mesh
-    yy1 : int or float
+    ymax : int or float
         Top-most y-coordinate in region to mesh
 
     Returns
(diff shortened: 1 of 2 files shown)


compass/landice/mesh.py
```diff
@@ -60,3 +61,43 @@ def gridded_flood_fill(field):
         lastSearchList = newSearchList
 
     return floodMask
+
+
+def set_geom_points_and_edges(xx0, xx1, yy0, yy1):
+    """
+    Set node and edge coordinates to pass to build_planar_mesh().
```
xylar, line 68:
```suggestion
    Set node and edge coordinates to pass to 
    :py:func:`mpas_tools.mesh.creation.build_mesh.build_planar_mesh()`.
```
If you do this, it should automatically link to that function in the MPAS-Tools API, which can add clarity and make it easier to navigate.

xylar (approved):
This looks good to me!

The names `xx0`, `xx1`, etc. are a little odd but there's presumably a history to that.

I have a minor suggestion for supporting cross-linking between the documentation for MPAS-Tools and for this package, but it's fine either way.

compass/landice/mesh.py
trhille: Oh great, thanks!

trhille:
> This looks good to me!
> 
> The names `xx0`, `xx1`, etc. are a little odd but there's presumably a history to that.
That is what they were called in the legacy code, but I could call them xMin, xMax, yMin, yMax for clarity.

xylar:
> That is what they were called in the legacy code, but I could call them xMin, xMax, yMin, yMax for clarity.

I'd be very happy with `x_min`, etc. or `xmin`, etc.  We don't use camel case for python variables.

xylar (approved):
Perfect!

xylar:
@trhille, could you rebase onto `master` now that #320 is in?

trhille:
> @trhille, could you rebase onto `master` now that #320 is in?

Rebased on force-push to https://github.com/MPAS-Dev/compass/commit/a09ebb1f7d34e65f736eed10fa9e453f6f8cf30c

compass/landice/mesh.py
```diff
@@ -60,3 +61,44 @@ def gridded_flood_fill(field):
         lastSearchList = newSearchList
 
     return floodMask
+
+
+def set_geom_points_and_edges(xmin, xmax, ymin, ymax):
```
matthewhoffman, line 66: I propose to rename this to `set_rectangular_geom_points_and_edges` or similar.  I suspect defining the region to mesh by a rectangle is what we'll always do for convenience, but in principle we could create any custom polygon shape to mesh inside of.  When we get to ultra high res or possibly if we try to mesh to exact boundaries, we may introduce other, similar functions.

compass/landice/mesh.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import jigsawpy
```
matthewhoffman, line 2: Now that this has moved here, it looks like the corresponding import in `compass/landice/tests/humboldt/mesh.py`can be safely deleted.

matthewhoffman (changes requested):
@trhille , thanks for working on abstracting this out for general use.  I made two small comments, but otherwise this looks great.

@xylar, after I'm happy with those changes (and assuming CI still passes), are you ok with me merging myself?

xylar:
Sure, feel free to merge, though ideally someone should set up and run the Humboldt test case one last time after all the changes come in.

compass/landice/mesh.py
trhille: Renamed in [68a8205](https://github.com/MPAS-Dev/compass/pull/319/commits/68a82051daed908e1ebfbb3051b4a862c27aaa59)

trhille:
I ran the Humboldt test case after addressing @matthewhoffman's requested changes in commit [68a8205](https://github.com/MPAS-Dev/compass/pull/319/commits/68a82051daed908e1ebfbb3051b4a862c27aaa59). The test passed and the mesh is identical to the previous version.

PR #321, opened 2022-02-15
Create set_cell_width() function

trhille:
Create a function to determine cell_width based on input parameters in the config file. This currently contains three different cell spacing functions based on log10(speed), distance to ice edge, and distance to the grounding line that can be turned on or off in the config file. Minimum and maximum cell width and the distances over which cell width is varied are also set in the config file. Implement a use case for Humboldt and update User's Guide. This will be used by all mesh-generating land ice test cases based on real-world geometries.

Note: The mesh created here is not bit-for-bit with the previous version. It is not clear why this is the case, but the differences are trivial.

PR #322, opened 2022-02-15
Create function to calculate distance to ice edge and grounding line

trhille:
Add new function compass.landice.mesh.get_dist_to_edge_and_GL() that calculates the distance from each grid point in gridded dataset to the ice margin and grounding line, which is then passed to compass.landice.mesh.set_cell_width() as input to the cell spacing functions. Add use case for Humboldt.

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -223,6 +223,11 @@ def get_dist_to_edge_and_GL(thk, topg, x, y, windowSize=1.e5):
         x coordinates from gridded dataset
     y : numpy.ndarray
         y coordinates from gridded dataset
+    windowSize : int or float
+        Size (in meters) of a search 'box' (one-directional) to use
+        to calculate the distance from each cell to the ice margin.
+        Bigger number makes search slower, but if too small, the transition
+        zone could get truncated.
 
     Returns
     -------
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -18,18 +18,18 @@ def gridded_flood_fill(field):
 
     Returns
     -------
-    floodMask : numpy.ndarray
-        Mask calculated by the flood fill routine,
+    flood_mask : numpy.ndarray
+        _mask calculated by the flood fill routine,
         where cells connected to the ice sheet (or main feature)
         are 1 and everything else is 0.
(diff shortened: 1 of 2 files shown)


trhille:
Force-push from  1e3a14e to 3d8d6d7 was a rebase onto origin/master following merge of PR #321.

xylar:
@trhille and @matthewhoffman, as a future optimization, you might want to look into faster packages for computing distances such as:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.html (used for signed distance functions in MPAS-Tools)

xylar:
Something went glitchy with testing but I don't think it's an actual problem.  Assuming it eventually times out, I'll rerun it.  Or if there are requested changes, that will trigger it to start again.

xylar (approved):
Other than my comment that an external package (probably `scikit-fmm`) would almost certainly be much faster, this looks fine to me.

Thanks for generalizing this, @trhille!

trhille:
Thanks @xylar! I'll address your comment in another PR.

xylar:
I'm going to re-trigger the testing by closing this and reopening it.  Just so you're not surprised...

matthewhoffman (changes requested):
@trhille , this generalization looks useful, and I don't see any concerns.  To Xylar's performance comment, I have one small request:  Add a comment at the top of the new routine mentioning the alternate methods that Xylar suggested, and then add a python timer to the function and a log message on how long the function took.  Then we'll have the info we need to improve performance if in practice it becomes limiting.  (But we won't worry about changing the implementation until then.)

trhille:
@matthewhoffman, thanks for your review. I added a note in the docstring and a log message with the time. It only takes 8 seconds for the Humboldt case. I'm sure that can be improved, but it's certainly not prohibitive for now. But large meshes like the forthcoming Antarctica mesh will probably benefit from @xylar's suggested improvement.

matthewhoffman (approved):
Thanks, Trevor.  That will be useful.

xylar:
@trhille, thanks for doing the timing!  It sounds like it's unlikely to become a serious problem.  Get back to me if you need something more efficient down the road.

PR #327, opened 2022-03-19
Update to `mache >=1.2.1`

xylar:
This version brings in important module update on Cori-Haswell and Chrysalis.

This merge also makes a change that helps with testing `mache`:  In addition to getting packages for the compass env from local builds when `--use_local` is provided, do the same for the temporary install environment so a local build of `mache` can be used.  This allows me to build `mache` locally and use it for making the `compass` conda environment.

### DIFF
--- a/conda/compass_env/spec-file.template
+++ b/conda/compass_env/spec-file.template
@@ -15,7 +15,7 @@ jigsaw=0.9.14
 jigsawpy=0.3.3
 jupyter
 lxml
-mache >=1.1.4
+mache >=1.2.1
 matplotlib-base
 metis
 mpas_tools=0.13.0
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I reran the same tests I used for `mache` before releasing v1.2.1 (see https://github.com/E3SM-Project/mache/pull/31#issuecomment-1072763053).

I have successfully run the ocean `nightly` test suite in `compass` using this branch with the following machines, compilers and MPI libraries:
- [ ] Chrysalis with Intel and OpenMPI
- [ ] Chrysalis with Gnu and OpenMPI
- [ ] Cori-Haswell with Intel and mpt
- [ ] Badger with Intel and Intel-MPI
- [ ] Anvil with Intel and Intel-MPI

xylar:
@dengwirda, this PR should fix the issues you reported to me on Slack from Cori-Haswell.

dengwirda:
Thanks @xylar.

PR #328, opened 2022-03-21
Add validation of max temp. during dynamic adjustment

xylar:
This will hopefully catch issues like #308

### DIFF
--- a/compass/ocean/tests/global_ocean/global_ocean.cfg
+++ b/compass/ocean/tests/global_ocean/global_ocean.cfg
@@ -71,6 +71,13 @@ email = autodetect
 pull_request = <<<Missing>>>
 
 
+# config options related to dynamic adjustment
+[dynamic_adjustment]
+
+# the maximum allowed value of temperatureMax in global statistics
+temperature_max = 33.0
+
+
 # config options related to initial condition and diagnostics support files
 # for E3SM
 [files_for_e3sm]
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I'm testing dynamic adjustment on Anvil and Intel with this test.
- [ ] EC30to60: this fails as expected
```
ValueError: Max of temperatureMax > allowed threshold: 54.22333235731241 > 33.0 in /lcrc/group/e3sm/ac.xylar/compass_1.0/anvil/test_20220321/dyn_adj_ec_validate_temp_redi_min_6/ocean/global_ocean/EC30to60/PHC/dynamic_adjustment/damped_adjustment_2/analysis_members/globalStats.0001-01-11_00.00.00.nc
```
- [ ] ECwISC30to60
- [ ] QU240
- [ ] QUwISC240
- [ ] SOwISC12to60
- [ ] WC14

compass/ocean/tests/global_ocean/global_ocean.cfg
```diff
@@ -71,6 +71,13 @@ email = autodetect
 pull_request = <<<Missing>>>
 
 
+# config options related to dynamic adjustment
+[dynamic_adjustment]
+
+# the maximum allowed value of temperatureMax in global statistics
+temperature_max = 33.0
```
xylar, line 78: I need to make sure this threshold is reasonable.

xylar:
@mark-petersen, this is ready for your review.  Is 33.0 C a good threshold in you opinion?

mark-petersen (approved):
@xylar this is a great idea! Thanks for implementing it. Yes, I think 33C is an appropriate value. We can be conservative here, as it will tell us if the max is increasing at all.

Approving based on this catching increasing temperatures in the tests above.

xylar:
Thanks again @mark-petersen!

PR #329, opened 2022-03-21
Increase `config_Redi_min_layers_diag_terms` to 8

xylar:
Based on testing with the EC30to60 mesh, this seems necessary to avoid spurious surface temperature extrema.

### DIFF
--- a/compass/ocean/tests/global_ocean/namelist.forward
+++ b/compass/ocean/tests/global_ocean/namelist.forward
@@ -15,6 +15,7 @@ config_implicit_bottom_drag_coeff = 1.0e-3
 config_use_bulk_wind_stress = .true.
 config_use_bulk_thickness_flux = .true.
 config_use_Redi = .true.
+config_Redi_min_layers_diag_terms = 8
 config_use_GM = .true.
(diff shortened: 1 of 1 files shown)


xylar:
In my testing, of EC30to60, this change is needed to fully address #308.

xylar:
## Testing

When I cherry-pick #328 onto this branch, EC30to60 now passes the validation of `dynamic_adjustment` on Anvil with Intel and Intel-MPI

xylar:
I also tried `config_Redi_min_layers_diag_terms = 7` but that was not sufficient to prevent new temperature extrema.

xylar:
@mark-petersen, would you like to see further testing of this PR?  I will test #309 and #310 again once this and #328 are merged.

mark-petersen (approved):
It makes sense that some meshes need a higher value for `config_Redi_min_layers_diag_terms`. I tuned that setting to a low value that worked for the WC14 mesh. Approving based on @xylar's experience with the EC60to30 here. Thanks.

xylar:
Thanks @mark-petersen!

PR #334, opened 2022-03-24
Fix out-of-range problem with permissions progress bar

xylar:
This happens when the number of files or directories changes in the middle of updating permissions on the conda or spack environment (often because the same user or someone else is modifying it).


xylar:
@trhille, this should fix the progress bar problem you just saw, once it goes in...

PR #337, opened 2022-03-25
Update to mache>=1.3.2

xylar:
This version adds a missing environment variable needed on Compy


xylar:
## Testing

I deployed on Compy with a test build of `mache` 1.3.2 and it fixed the runtime issue I was having with `mache` 1.3.1.  With this fix, I successfully run the `pr` test suite with Intel on that machine.

PR #338, opened 2022-03-27
Add dev tools to the conda package

xylar:
This merge adds the development tools (compilers, netcdf libraries and build tools like cmake) as dependencies of the compass package if using conda-forge MPI.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -46,14 +46,16 @@ jobs:
       eval "$(conda shell.bash hook)"
       # workaround based on recent failures
       rm /usr/share/miniconda/pkgs/cache/*.json
-      conda mambabuild -m ci/mpi_$(mpi)_python${PYTHON_VERSION}.yaml conda/recipe
+      conda mambabuild -m ci/mpi_$(mpi)_python${PYTHON_VERSION}.yaml \
+          -c conda-forge -c e3sm/label/compass conda/recipe
     displayName: Build COMPASS metapackage
 
   - bash: |
       set -e
       eval "$(conda shell.bash hook)"
       mamba create --yes --quiet --name compass -c ${CONDA_PREFIX}/conda-bld/ \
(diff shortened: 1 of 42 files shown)


xylar:
## Linux Testing

I built the `compass` conda package with MPICH on Linux and used that package to build MPAS-Ocean and run the `nightly` ocean test suite.  For that purpose, I needed the following load script:
```
source /home/xylar/miniconda3/etc/profile.d/conda.sh
conda activate test

export NETCDF=$(dirname $(dirname $(which nc-config)))
export NETCDFF=$(dirname $(dirname $(which nf-config)))
export PNETCDF=$(dirname $(dirname $(which pnetcdf-config)))

export MPAS_EXTERNAL_LIBS="-lgomp"
export PIO=/home/xylar/miniconda3/envs/test/
export USE_PIO2=true
export HDF5_USE_FILE_LOCKING=FALSE
export LOAD_COMPASS_ENV=/home/xylar/code/compass/add_dev_tools_to_conda_package/load_compass_1.0.0_mpich.sh
```
A follow-up PR (#340) will add command-line tool for generating such a load script without the need for a user to check out the `compass` repo.

xylar:
## OSX Testing

I did similar testing on OSX.  I'm seeing:
```
ERROR: MPAS IO Error: Bad return value from PIO
CRITICAL ERROR: Core init failed for core ocean
```
on the `ocean/global_ocean/QU240/PHC/restart_test` test case in the `restart_run` step.  

The `ocean/global_ocean/QU240/PHC/analysis_test` test case stalled in the `forward` step and the `ocean/global_ocean/QU240/PHC/RK4/restart_test` test case stalled in the `full_run` step.

Some other test cases are failing only because I don't have 8 cores (the minimum for some `decom` test cases).

For now, I'm just going to let these issues be because they only affects a few test case and OSX isn't a main target for us.  But this deserves another look after the 1.0.0 release.

PR #340, opened 2022-03-27
Add load package to compass

xylar:
The `compass.load` package and the `create_compass_load_script` command-line tool can be used to generate a load script when `compass` is installed as a conda package, rather than cloning the git repository.


xylar:
## Testing

Tested along with #338

PR #306, opened 2022-01-26
Add 2.5km, 5km and 10 km resolutions to ZISO test group

yariseidenbenz:
This merge changes the ZISO test group to support multiple resolutions. New resolutions range from 2.5 km, 5 km, 10 km, 20 km. Each resolution has a `default`, `long` and `particles` test case, and there is also still a `with_frazil` test case at 20 km resolution. These new resolutions and test cases were developed using the same approach as the SOMA test group.

PR #341, opened 2022-03-28
More updates and corrections to the documentation

xylar:
This merge includes updates to the documentation, particularly related to recent expansion of the ZISO test cases (#306) and the way to use compass on unknown (Linux or OSX) machines, following #340.


xylar:
## Testing

I have successfully built the documentation locally and checked the ZISO links to make sure they are now correct.

PR #349, opened 2022-04-01
Fix how docs detects compass version

xylar:
This got missed in #343


xylar:
## Testing

With these changes, I'm able to build the docs locally whereas they fail without this change.

CI didn't catch this because it assigns the version number rather than trying to detect it.

PR #352, opened 2022-04-05
Add drying slope case

cbegeman:
This PR ports the MPAS-Ocean `drying_slope` test case from legacy, which is based on 

Warner, J. C., Defne, Z., Haas, K., & Arango, H. G. (2013). A wetting and
drying scheme for ROMS. Computers & geosciences, 58, 54-61. 

Currently, we support one vertical coordinate (`sigma`) and two horizontal resolutions (250m, 1km). The test case was verified to produce the same results as the legacy case. 
image
image


xylar:
This is exciting!  I'll give this a look tomorrow.  

I took off the "legacy" tag because we use that to indicate PRs into the `legacy` branch.  I'll update the description of that tag to make that clear.

xylar (commented):
This is really great! I'm able to run the test cases successfully through both forward runs but the `viz` step is giving me some trouble.

`viz` is failing because the csv files are not getting downloaded automatically.  It will be necessary to add all of the csv files as input files to the step, using `database=drying_slope` as one of the arguments.

Then, I think it would be easier if you just moved the functions from MoviePlotter and TimeSeriesPlotter into Viz.  This is because there need to be some shared attributes (`times`, `damping_coeffs` and `datatypes` seem like some obvious ones.  These could be passed from Viz over to each of the other classes but I think it just doesn't seem clearly advantageous to have 3 classes in this case.

I do realize that this approach was inherited either from `IsomipPlus` or from the legacy version of `drying_slope` (or perhaps both) and predates the current compass.

One more thing is that documentation needs to be added before this gets merged.

compass/ocean/tests/drying_slope/viz/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+        ----------
+        test_case : compass.TestCase
+            The test case this step belongs to
+        """
+        super().__init__(test_case=test_case, name='viz')
+
+        for damping_coeff in [0.0025, 0.01]:
+            self.add_input_file(filename='output_{}.nc'.format(damping_coeff),
+                                target='../forward_{}/output.nc'.format(
+                                       damping_coeff))
```
xylar, line 31: Here is a rough idea of how the csv files would be added.  However, please don't hard-code these arrays as I have done here.
 
``` python
        for damping_coeff in [0.0025, 0.01]:
            for time in ['0.50', '0.05', '0.40', '0.15', '0.30', '0.25']:
                for datatype in ['analytical', 'roms']:
                    filename = f'r{damping_coeff}d{time}-{datatype}.csv'
                    self.add_input_file(filename=filename, target=filename,
                                        database='drying_slope')
```

compass/ocean/tests/drying_slope/viz/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+        Run this step of the test case
+        """
+        section = self.config['paths']
+        datapath = section.get('ocean_database_root')
+        section = self.config['drying_slope_viz']
+        frames_per_second = section.getint('frames_per_second')
+        movie_format = section.get('movie_format')
+
+        tsPlotter = TimeSeriesPlotter()
+        tsPlotter.plot_ssh_validation(datapath)
```
xylar, line 44: `datapath` should just be `'.'` with my suggestion above.  You don't need to find the files in the data base, there will be local symlinks to them automatically.

compass/ocean/tests/drying_slope/viz/plot.py
```diff
@@ -0,0 +1,324 @@
...
+
+                # Plot comparison data
+                for atime, ay in zip(times, locs):
+                    ax2[i].text(1, ay, atime + ' days', size=8,
+                                transform=ax2[i].transAxes)
+
+                    for datatype in ['analytical', 'ROMS']:
+                        datafile = '{}/drying_slope/r{}d{}-{}.csv'.format(
+                                    datapath, damping_coeff, atime,
+                                    datatype.lower())
```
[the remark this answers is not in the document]
xylar: Again, you shouldn't need `datapath`, just use the symlinks to these files in the current directory.

cbegeman:
@xylar Thanks for the detailed review and suggestions! I implemented all those changes and will get the documentation together by Monday.

xylar (commented):
This is really excellent work!  The documentation is very clear.  The new organization of the `viz` step seems to work really well, too.  There are a few new PEP8 details to deal with (see https://github.com/MPAS-Dev/compass/pull/352#issuecomment-1089283514) and I have some additional very minor ones.

I ran both test cases and the resulting movie is really fun to watch!

cbegeman:
@xylar I believe I've now made all the changes you mentioned and resolved PEP issues.

@sbrus89 Please give this a whirl when you have the chance! 

Thanks to you both!

sbrus89:
@cbegeman, super. I'll test this today.

compass/ocean/tests/drying_slope/namelist.forward
```diff
@@ -0,0 +1,28 @@
+config_run_duration='0000_12:00:01'
+config_dt='0000_00:00:30'
```
sbrus89, line 2: The legacy case uses `config_dt='0000_00:00:20'` for the 250m resolution. With the current 30s dt, there are some large errors on the right boundary for the 250m case:
ssh_depth_section_30s

For a 20 second dt:
ssh_depth_section_20s


I also ran it with a 10 second dt:
ssh_depth_section_10s

The smaller dt progressively improves the boundary mismatch and also improves the drying front (especially for r=0.0025). Maybe we should with with 10s (or even 5 since there's still a slight mismatch with 10?)

sbrus89 (commented):
@cbegeman, this is a great setup! I have a few minor comments, hopefully they are useful.

compass/ocean/tests/drying_slope/namelist.forward
```diff
@@ -0,0 +1,28 @@
...
+config_vert_coord_movement='impermeable_interfaces'
+config_ALE_thickness_proportionality='weights_only'
+config_use_wetting_drying=.true.
+config_prevent_drying=.true.
+config_drying_min_cell_height=1.0e-3
+config_zero_drying_velocity=.true.
+config_verify_not_dry=.true.
+config_thickness_flux_type='upwind'
+config_cvmix_background_viscosity=1e4
+config_use_variable_drag=.true.
```
sbrus89, line 22: We should set `config_use_variable_drag=.true.`. I don't think this option does anything unless other options in the `bottom_drag` namelist section are active

compass/ocean/tests/drying_slope/namelist.forward
```diff
@@ -0,0 +1,28 @@
...
+config_eos_linear_beta=0.0
+config_vert_coord_movement='impermeable_interfaces'
+config_ALE_thickness_proportionality='weights_only'
+config_use_wetting_drying=.true.
+config_prevent_drying=.true.
+config_drying_min_cell_height=1.0e-3
+config_zero_drying_velocity=.true.
+config_verify_not_dry=.true.
+config_thickness_flux_type='upwind'
+config_cvmix_background_viscosity=1e4
```
sbrus89, line 21: I think we should set `config_use_cvmix = .false.`.

compass/ocean/tests/drying_slope/namelist.init
```diff
@@ -0,0 +1,13 @@
...
+config_tidal_boundary_layer_type='sigma'
+config_tidal_start_dry=.true.
+config_tidal_boundary_left_bottom_depth=0.0
+config_tidal_boundary_use_distances=.true.
+config_tidal_forcing_monochromatic_baseline=10.0
+config_use_wetting_drying=.true.
+config_drying_min_cell_height=1.000001e-3
+config_use_tidal_forcing=.true.
+config_write_cull_cell_mask=.false.
+config_use_variable_drag=.true.
```
sbrus89, line 13: `config_use_variable_drag=.false.` here too

compass/ocean/tests/drying_slope/drying_slope.cfg
```diff
@@ -0,0 +1,27 @@
...
+
+# The type of vertical coordinate (e.g. z-level, z-star)
+coord_type = sigma
+
+# config options for drying slope test cases
+[drying_slope]
+
+# the number of grid cells in x and y
+nx = 6
+ny = 28
```
sbrus89, line 15: Is `ny` here necessary since it's computed based on `resolution`? It seems like the calculation in `Default.configure` could be moved to `InitialState.run` and then `ny` can be removed here.

compass/ocean/tests/drying_slope/drying_slope.cfg
```diff
@@ -0,0 +1,27 @@
...
+
+# the size of grid cells (m)
+dc = 1000.0
```
sbrus89, line 18: Similarly, this doesn't seem to be needed. Can `resolution` be passed as an argument to `self.add_step(InitialState(test_case=self))` so this can get set in `InitialState.run`?

compass/ocean/tests/drying_slope/drying_slope.cfg
```diff
@@ -0,0 +1,27 @@
...
+
+# config options for visualizing drying slope ouptut
+[drying_slope_viz]
+
+# frames per second for movies
+frames_per_second = 30
+
+# movie format
+movie_format = mp4
```
sbrus89, line 27: What do you think about an option to skip making the movie?

compass/ocean/tests/drying_slope/namelist.forward
cbegeman: @sbrus89 Good idea. How about 8s since that is roughly 1/4 the dt of the lower resolution case?

sbrus89: Sounds good to me.

compass/ocean/tests/drying_slope/drying_slope.cfg
xylar: @sbrus89, the philosophy is that users could reasonably change this.  This is just the default.  If you hard code it in `run()`, a user can't change it.

xylar: But you are right.  If it is set internally in the code, it shouldn't also be here.

compass/ocean/tests/drying_slope/drying_slope.cfg
xylar: The same, if this is set in the code (in `configure()`), it shouldn't be set here, too.

compass/ocean/tests/drying_slope/drying_slope.cfg
xylar: Hmm, that's not a bad idea.  The default should probably be to *not* make the movie if we're using this as a regression test.

compass/ocean/tests/drying_slope/drying_slope.cfg
cbegeman: I'll defer to @xylar on whether it makes more sense to take the original approach of having these be config options that are modified in `Default.configure` or remove these as config options and set them in `InitialState.run` as you suggest. My thinking behind the use of config options was making these dimensions more externally visible, but I can see how it might be misleading since `dc` and `ny` are not independent.

compass/ocean/tests/drying_slope/namelist.init
cbegeman: I just want to double check that you think we should skip these lines. I believe this would mean we would also remove `bottomDrag` as forcing input for the forward run. I think this would be a deviation from the legacy test case.
```
      if (config_use_variable_drag) then
        call mpas_pool_get_array(forcingPool, 'bottomDrag', bottomDrag)
        do iCell = 1, nCellsSolve
          bottomDrag(iCell) = config_tidal_forcing_left_Cd_or_n &
            + (yCell(iCell) - yMin) / (yMax - yMin)  * &
            (config_tidal_forcing_right_Cd_or_n - config_tidal_forcing_left_Cd_or_n)
        end do
      end if
```
https://github.com/E3SM-Project/E3SM/blob/4435bfa0e188f8419a4c2dc0dbcb88ceacea7e4f/components/mpas-ocean/src/mode_init/mpas_ocn_init_tidal_boundary.F#L220-L227

compass/ocean/tests/drying_slope/drying_slope.cfg
xylar: I think it's better to include it in configure() and not in the config file, to make a bit clearer that the configure() version always takes priority. You can use the brand new "comment" argument to set() to have the comment included there instead.

compass/ocean/tests/drying_slope/namelist.forward
cbegeman: I agree with you on this. I only see this config option being used in init mode.

compass/ocean/tests/drying_slope/drying_slope.cfg
cbegeman: Good idea. I'll add that.

compass/ocean/tests/drying_slope/drying_slope.cfg
cbegeman: Thanks @xylar! I didn't refresh in time so my comment was a bit lagged but it all worked out.

compass/ocean/tests/drying_slope/namelist.init
sbrus89: That's true. I d true. I don't think that bottomDrag is usedon't think that bottomDrag is used in forward mode for this case, since it just uses Rayleigh drag (set via`config_Rayleigh_damping_coeff = ... `). So it should be fine to remove, right? If I recall correctly. I think the variable drag only used in the marsh test cases in legacy, which would require `config_use_variable_drag=.true.`

cbegeman: Yes, you're right. I just double-checked the `ocn_vmix` routines. I'll remove `config_use_variable_drag` (it's false by default) and `bottomDrag` from the forcing stream.

sbrus89: great, that sounds good to me

cbegeman:
I've made the changes @sbrus89 suggested. I also rebased to make use of the new comment feature.

Note: https://github.com/MPAS-Dev/compass/pull/352/commits/140c15b49c15aa0e7bb975039f23a9b818184695 makes use of additional namelist files to set `config_dt` for each resolution. I'm happy to change this to the `add_namelist_options` method in `Forward` step if that's preferred.

xylar:
> Note: [af3b8bb](https://github.com/MPAS-Dev/compass/pull/352/commits/af3b8bb6657e1f9b97bc092ba13d8d5af5fedb8a) makes use of additional namelist files to set `config_dt` for each resolution. I'm happy to change this to the `add_namelist_options` method in `Forward` step if that's preferred.

If it were me, I would use `add_namelist_options` but I don't think it's necessary to change it.

xylar (approved):
I'm good with this.  Hopefully, one of you can run one last set of tests?

xylar:
@dengwirda, we're ready for your review on this so we can get it in.

cbegeman:
I did run these test cases with the latest commits on anvil.

compass/ocean/tests/drying_slope/namelist.init
sbrus89: @cbegeman, Is removing this option and the streams variable something we still would like to do?

docs/users_guide/ocean/test_groups/drying_slope.rst
```diff
@@ -0,0 +1,85 @@
...
+   dc = 1000.0
+   
+   # config options for visualizing drying slope ouptut
+   [drying_slope_viz]
+   
+   # frames per second for movies
+   frames_per_second = 30
+   
+   # movie format
+   movie_format = mp4
```
sbrus89, line 75: Should this be updated with the changes to the config file?

compass/ocean/tests/drying_slope/namelist.init
cbegeman: Thanks for reminding me. Let me test that out now.

cbegeman: Looks good. No differences in the forward solution after the change.

docs/users_guide/ocean/test_groups/drying_slope.rst
cbegeman: Good catch. I made those updates. Thanks, @sbrus89!

xylar:
Thanks, @cbegeman!  Such a good feature to have!

PR #326, opened 2022-03-18
Update E3SM-Project submodule

xylar:
This merge updates the E3SM-Project submodule to match today's [E3SM/master](https://github.com/E3SM-Project/E3SM/tree/7b87d1faa545f7da4e792058401ad3bc04434c85)
Commit hash: 7b87d1faa545f7da4e792058401ad3bc04434c85

Merges of interest to compass:
* [Add mostRecentAccessTime attribute to streams](https://github.com/E3SM-Project/E3SM/pull/4752) - MALI
* [GPU port of thickness tendency with OpenACC](https://github.com/E3SM-Project/E3SM/pull/4792) - MPAS-Ocean
* [Extending OpenACC diagnostics up to time integration](https://github.com/E3SM-Project/E3SM/pull/4813) - MPAS-Ocean
* [Fix non-standard declarations of 8-byte integers in ESMF in MPAS-Framework](https://github.com/E3SM-Project/E3SM/pull/4825) - MPAS Framework
* [Add -fallow-argument-mismatch to gfortran in MPAS standalone builds](https://github.com/E3SM-Project/E3SM/pull/4821) - MPAS Framework
* [Turn off Redi diagonal terms for shallow columns](https://github.com/E3SM-Project/E3SM/pull/4785) - MPAS-Ocean

closes #308

PR #354, opened 2022-04-08
Add scripts for using `git bisect`

xylar:
A config file determines the "good" and "bad" commits, and provides the paths to the MPAS code, the load script, etc.  It
also provides the commands for building the code, setting up the test case(s) and running them.

A driver script sets things up and makes the calls to `git bisect`, passing along the `utils/bisect/bisect_step.py` script for running each step of the bisection.

The `utils/bisect/bisect_step.py` script updates and builds the MPAS code, then sets up the compass test case(s) and runs compass.

### DIFF
--- a/utils/bisect/README.md
+++ b/utils/bisect/README.md
@@ -1,0 +1,36 @@
+Using `git bisect`
+====================
+
+The command `git bisect` is a handy tool for finding the first commit that
+breaks a code in some way.  `git bisect run` can call a script that
+succeeds when a given commit is "good" but fails when it is "bad".  The script
+`bisect_step.py` provided here is one such script.
+
+To further encapsulate the process of using `git bisect`, we provide a driver
+script `bisect.py` that makes use of config options in a file similar to
+`example.cfg`.
+
+Instructions
+------------
+
+1. Copy `example.cfg` to the base of the branch:
+   ```shell
+   cp utils/bisect/example.cfg bisect.cfg
+   ```
+2. Modify the config options with the appropriate "good" and "bad" E3SM commit hash
+or tag.
+
+3. Modify the various paths and commands as needed.
+
+4. In a batch job or interactive session on a compute node, run:
+   ```shell
+   ./utils/bisect/bisect.py -f bisect.cfg
+   ```
+
+Note
+----
+
+Since the code will be compiled on a compute node, any baseline use for
+comparison should also be built on a compute node.  Otherwise, you may get
+non-bit-for-bit results simply because of where the code was compiled.  This
+has been seen with Intel on Anvil.
--- a/utils/bisect/bisect.py
+++ b/utils/bisect/bisect.py
@@ -58,6 +58,10 @@ def bisect(good, bad, e3sm_path, load_script, config_file, first_parent):
                f'git bisect bad {bad}; ' \
                f'git bisect run {cwd}/utils/bisect/bisect_step.py' \
                f'  -f {config_file}'
+    print('\n')
+    print(72*'-')
+    print('Biscect Initialization')
+    print(72*'-')
     print('\nRunning:')
     print_commands = commands.replace('; ', '\n  ')
     print(f'  {print_commands}\n\n')
--- a/utils/bisect/bisect.py
+++ b/utils/bisect/bisect.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+
+import argparse
+import configparser
+import os
+import subprocess
+
+
+def bisect(good, bad, e3sm_path, load_script, config_file, first_parent):
+    """
+    The driver function for calling ``git bisect`` to find the first "bad"
+    commit between a known "good" and "bad" E3SM commit.
+
+    The function uses ``git bisect run`` to call
+    ``utils/bisect/bisect_step.py`` repeatedly to test whether a given commit
+    is good or bad.
+
+    Parameters
+    ----------
+    good : str
+        The hash or tag of a "good" E3SM commit that passes the test(s)
+        specified in the config file
+    bad : str
+        The hash or tag of a "bad" E3SM commit that fails the test(s)
+    e3sm_path : str
+        The path to the E3SM branch to test.  If you are pointing to the
+        ``E3SM-Project`` or ``MALI-Dev`` submodules, make sure they have been
+        initialized with ``git submodule update --init``.
+    load_script : str
+        The relative or absolute path to the load script used to activate
+        the compass conda environment and set environment variables used to
+        build the MPAS component to test.
+    config_file : str
+        The relative or absolute path to a config file containing config
+        options similar to ``utils/bisect/example.cfg`` that control the
+        bisection process.
+    first_parent : bool
+        Whether to only follow the first parent for merge commits.  This is
+        typically desirable because there may be broken commits within a branch
+        that are fixed by the time the branch is merged.
+    """
+
+    e3sm_path = os.path.abspath(e3sm_path)
+    load_script = os.path.abspath(load_script)
+    config_file = os.path.abspath(config_file)
+
+    cwd = os.getcwd()
+
+    if first_parent:
+        flags = '--first-parent'
+    else:
+        flags = ''
+
+    commands = f'source {load_script}; ' \
+               f'cd {e3sm_path}; ' \
+               f'git bisect start {flags}; ' \
+               f'git bisect good {good}; ' \
+               f'git bisect bad {bad}; ' \
+               f'git bisect run {cwd}/utils/bisect/bisect_step.py' \
+               f'  -f {config_file}'
+    print('\nRunning:')
+    print_commands = commands.replace('; ', '\n  ')
+    print(f'  {print_commands}\n\n')
+    subprocess.check_call(commands, shell=True)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='Use "git bisect" to find the first E3SM commit for which '
+                    'a given test fails')
+    parser.add_argument("-f", "--config_file", dest="config_file",
+                        required=True,
+                        help="Configuration file with bisect options",
+                        metavar="FILE")
+
+    args = parser.parse_args()
+
+    config = configparser.ConfigParser(
+        interpolation=configparser.ExtendedInterpolation())
+    config.read(args.config_file)
+
+    section = config['bisect']
+
+    bisect(good=section['good'], bad=section['bad'],
+           e3sm_path=section['e3sm_path'],
+           load_script=section['load_script'],
+           config_file=args.config_file,
+           first_parent=section.getboolean('first_parent'))
+
+
+if __name__ == '__main__':
+    main()
--- a/utils/bisect/bisect.py
+++ b/utils/bisect/bisect.py
@@ -51,8 +51,7 @@ def bisect(good, bad, e3sm_path, load_script, config_file, first_parent):
     else:
         flags = ''
 
-    commands = f'source {load_script}; ' \
-               f'cd {e3sm_path}; ' \
+    commands = f'cd {e3sm_path}; ' \
                f'git bisect start {flags}; ' \
                f'git bisect good {good}; ' \
                f'git bisect bad {bad}; ' \
--- a/utils/bisect/bisect_step.py
+++ b/utils/bisect/bisect_step.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+
+import argparse
+import configparser
+import os
+import subprocess
+
+
+def run(launch_path, mpas_path, work_base, load_script, make_command,
+        setup_command, run_command):
+    """
+    This function runs a single step in the bisection process.  It is typically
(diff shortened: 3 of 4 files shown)


xylar:
@sbrus89, I believe this is working now.  It is ready for you to test and review.  I don't plan to document it beyond `bisect/README.md` unless you would like it to be in the main documentation.

sbrus89:
@xylar, very cool! I'll take a look.

xylar:
I've been trying to clean this up a bit (e.g. make sure the `make` output doesn't overwhelm everything else).  I'll re-test tomorrow or Monday.

xylar:
Using this approach, I was able to trace the unexpected non-BFB changes in baroclinic channel runs in https://github.com/MPAS-Dev/compass/pull/326 to https://github.com/E3SM-Project/E3SM/pull/4752.  That's good to know but not concerning.

xylar:
@sbrus89, please don't forget to review this when you can.

sbrus89:
@xylar, will do.

sbrus89:
Now I'm not seeing it. Probably something on my end.

utils/bisect/bisect.py
```diff
@@ -0,0 +1,92 @@
...
+import argparse
+import configparser
+import os
+import subprocess
+
+
+def bisect(good, bad, e3sm_path, load_script, config_file, first_parent):
+    """
+    The driver function for calling ``git bisect`` to find the first "bad"
+    commit between a known "good" and "bad" commit.
```
xylar, line 12:
```suggestion
    commit between a known "good" and "bad" E3SM commit.
```

utils/bisect/bisect.py
```diff
@@ -0,0 +1,92 @@
...
+    ``utils/bisect/bisect_step.py`` repeatedly to test whether a given commit
+    is good or bad.
+
+    Parameters
+    ----------
+    good : str
+        The hash or tag of a "good" commit that passes the test(s) specified
+        in the config file
+    bad : str
+        The hash or tag of a "bad" commit that fails the test(s)
```
xylar, line 24:
```suggestion
    good : str
        The hash or tag of a "good" E3SM commit that passes the test(s) specified
        in the config file
    bad : str
        The hash or tag of a "bad" E3SM commit that fails the test(s)
```

utils/bisect/README.md
```diff
@@ -0,0 +1,28 @@
...
+`example.cfg`.
+
+Instructions
+------------
+
+1. Copy `example.cfg` to the base of the branch:
+   ```shell
+   cp utils/bisect/example.cfg bisect.cfg
+   ```
+2. Modify the config options with the appropriate "good" and "bad" commit hash
```
xylar, line 20:
```suggestion
2. Modify the config options with the appropriate "good" and "bad" E3SM commit hash
```

utils/bisect/example.cfg
```diff
@@ -0,0 +1,30 @@
+# config options related to using "git bisect" to find the first commit that
```
xylar, line 1:
```suggestion
# config options related to using "git bisect" to find the first E3SM commit that
```

utils/bisect/example.cfg
```diff
@@ -0,0 +1,30 @@
...
+# causes a set of one or more tests to fail
+[bisect]
+
+# The hash or tag of a good commit where the tests pass
+good = 44814ae
+# The hash or tag of a bad commit where the tests fail
+bad = 7b87d1f
```
xylar, line 8:
```suggestion
# The hash or tag of a good E3SM commit where the tests pass
good = 44814ae
# The hash or tag of a bad E3SM commit where the tests fail
bad = 7b87d1f
```

utils/bisect/bisect.py
```diff
@@ -0,0 +1,92 @@
...
+
+    The function uses ``git bisect run`` to call
+    ``utils/bisect/bisect_step.py`` repeatedly to test whether a given commit
+    is good or bad.
+
+    Parameters
+    ----------
+    good : str
+        The hash or tag of a "good" E3SM commit that passes the test(s) specified
+        in the config file
```
xylar, line 22:
```suggestion
        The hash or tag of a "good" E3SM commit that passes the test(s)
        specified in the config file
```

xylar:
@sbrus89, any chance of giving this a look in the next few weeks?  It would be nice to have it in.

sbrus89:
@xylar, sure I'll put it on my todo list.

xylar:
@sbrus89, I'm not sure what changed but I'm not having any luck with this.  My runs just hang when I try to run on an interactive node and they fail in ways I haven't figured out yet when I try to run a job.

xylar:
Oh, I take that back.  The job script worked fine, it just returned an error code for reasons I don't understand.  It seems like it found the bad commit and then tried to run again or something.  I need to investigate sometime but don't have time right now.

utils/bisect/bisect.py
```diff
@@ -0,0 +1,92 @@
...
+    config_file = os.path.abspath(config_file)
+
+    cwd = os.getcwd()
+
+    if first_parent:
+        flags = '--first-parent'
+    else:
+        flags = ''
+
+    commands = f'source {load_script}; ' \
```
sbrus89, line 54: Is this first `source {load_script}` necessary?

utils/bisect/bisect_step.py
```diff
@@ -0,0 +1,120 @@
...
+        os.makedirs(work_path)
+    except FileExistsError:
+        pass
+
+    os.chdir(mpas_path)
+    commands = f'source {load_script}; ' \
+               f'git submodule update --init --recursive; ' \
+               f'make clean >& {work_path}/clean.log; ' \
+               f'{make_command} >& {work_path}/make.log; ' \
+               f'{setup_command} -p {mpas_path} -w {work_path}; ' \
```
sbrus89, line 64: Adding the `--copy_executable` option from #360 would be useful here.

utils/bisect/bisect_step.py
```diff
@@ -0,0 +1,120 @@
...
+        The command (typically just ``compass run``) use to run the compass
+        test case(s)
+    """
+
+    mpas_path = to_abs(mpas_path, launch_path)
+    work_base = to_abs(work_base, launch_path)
+    load_script = to_abs(load_script, launch_path)
+
+    commands = f'cd {mpas_path}; ' \
+               f'source {load_script}; ' \
```
sbrus89, line 46: Is this `source {load_script}` necessary?

utils/bisect/bisect.py
sbrus89: I suppose this is to ensure that  the `git` version is recent?

utils/bisect/bisect_step.py
sbrus89: Same here. Is this to make sure `git` is reasonably up-to-date?

utils/bisect/bisect_step.py
```diff
@@ -0,0 +1,120 @@
...
+
+    os.chdir(mpas_path)
+    commands = f'source {load_script}; ' \
+               f'git submodule update --init --recursive; ' \
+               f'make clean >& {work_path}/clean.log; ' \
+               f'{make_command} >& {work_path}/make.log; ' \
+               f'{setup_command} -p {mpas_path} -w {work_path}; ' \
+               f'cd {work_path}; ' \
+               f'{run_command}'
+    print('\nRunning:')
```
sbrus89, line 67: Before this line, I suggest adding something to the effect of:
```
    print('\n')
    print(72*'-')
    print('Biscect Step')
    print(72*'-')
```
I think this would be helpful in cleaning up the screen output to make it more clear where the bisect steps begin. This isn't quite perfect, because `git bisect run` does the bisect between calls to this script. Maybe those lines should go at the end then. But, then it wouldn't show up for the first call...

utils/bisect/bisect_step.py
```diff
@@ -0,0 +1,120 @@
...
+    print_commands = commands.replace('; ', '\n  ')
+    print(f'  {print_commands}\n\n')
+    subprocess.check_call(commands, shell=True)
```
sbrus89, line 70: Maybe add a `print('\n')` after this line so the `compass run` screen output is separate from the `git bisect run` output

utils/bisect/bisect.py
```diff
@@ -0,0 +1,92 @@
...
+               f'cd {e3sm_path}; ' \
+               f'git bisect start {flags}; ' \
+               f'git bisect good {good}; ' \
+               f'git bisect bad {bad}; ' \
+               f'git bisect run {cwd}/utils/bisect/bisect_step.py' \
+               f'  -f {config_file}'
+    print('\nRunning:')
```
sbrus89, line 61: Before this line maybe add something like:

```
    print('\n')
    print(72*'-')
    print('Biscect Initialization')
    print(72*'-')
```

to clearly indicate (along with the similar comment in `bisect_step.py`) what is going on at the highest level in the process.

utils/bisect/bisect_step.py
sbrus89: If so, that's fine. But if we can get rid of it, it would help speed things up a little.

utils/bisect/bisect_step.py
sbrus89: If you agree, then we should merge #360 before this PR.

utils/bisect/bisect_step.py
sbrus89: Although, this only works when a test passes.

utils/bisect/bisect.py
xylar: Yes, I think that was the idea.  Is it causing trouble.

utils/bisect/bisect.py
xylar:
```suggestion
    print('\n')
    print(72*'-')
    print('Biscect Initialization')
    print(72*'-')
    print('\nRunning:')
```

utils/bisect/bisect_step.py
xylar: Okay, we can let it go and assume the user has called `module load git` or something like that if need be.

utils/bisect/bisect_step.py
xylar:
```suggestion
    print('\n')
    print(72*'-')
    print('Biscect Step')
    print(72*'-')
    print('\nRunning:')
```

utils/bisect/bisect.py
sbrus89: No, it doesn't cause any trouble. It can just slow things down a bit.

utils/bisect/bisect_step.py
xylar: I'll add this after the `run()` call.

utils/bisect/bisect_step.py
```diff
@@ -0,0 +1,120 @@
...
+    config.read(args.config_file)
+
+    launch_path = os.path.dirname(args.config_file)
+
+    section = config['bisect']
+    run(launch_path=launch_path, mpas_path=section['mpas_path'],
+        work_base=section['work_base'], load_script=section['load_script'],
+        make_command=section['make_command'],
+        setup_command=section['setup_command'],
+        run_command=section['run_command'])
```
xylar, line 116:
```suggestion
        run_command=section['run_command'])
    print ('\n')
```

utils/bisect/bisect.py
```diff
@@ -0,0 +1,96 @@
...
+
+    cwd = os.getcwd()
+
+    if first_parent:
+        flags = '--first-parent'
+    else:
+        flags = ''
+
+    commands = f'source {load_script}; ' \
+               f'cd {e3sm_path}; ' \
```
xylar, line 55:
```suggestion
    commands = f'cd {e3sm_path}; ' \
```

utils/bisect/bisect_step.py
```diff
@@ -0,0 +1,125 @@
...
+        The command (typically just ``compass run``) use to run the compass
+        test case(s)
+    """
+
+    mpas_path = to_abs(mpas_path, launch_path)
+    work_base = to_abs(work_base, launch_path)
+    load_script = to_abs(load_script, launch_path)
+
+    commands = f'cd {mpas_path}; ' \
+               f'source {load_script}; ' \
```
xylar, line 46:
```suggestion
    commands = f'cd {mpas_path}; ' \
```

utils/bisect/bisect_step.py
xylar: I think it makes the most sense to include this in `setup_command`, which will do in the example.

xylar:
> The only other comment I have is it may be convenient to create a summary log file that just prints the commit information along with the PASS/FAIL status of each of the commits that is checked-out in the process. I think that is very optional though...

@sbrus89, I don't feel like I have time to do that right now.  I would welcome a follow-up PR to add this, though!

Could you take a look and see if I've addressed your suggestions?  I think so.  If so, I'll merge after CI has finished unless you want to run another test.

xylar:
@sbrus89, great, thanks!

PR #356, opened 2022-04-12
Fix docs deployment on tag

xylar:
The v1.0.0 tag didn't create docs as it was supposed to.  These missing quotes should help.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -100,7 +100,7 @@ jobs:
         deploy=False
       fi
 
-      if [ -n ${branch} ]; then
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I made a new `test_docs` tag successfully deployed docs here:
https://dev.azure.com/MPAS-Dev/compass%20testing/_build/results?buildId=2576&view=logs&j=2910401f-f502-52f6-c37a-26a24a040b37&t=8df40b6d-1bee-5150-a8fe-6bc0b420fd44

I then deleted the docs and the tag.

PR #353, opened 2022-04-07
Switch to building system libraries with Spack

xylar:
This merge switched to using Spack to build system libraries.  This change:
* adds support for Albany wherever possible (see `conda/albany_support.txt` for supported machines, compilers and MPI libraries)
* harnesses maintenance from the community for updating package versions and recipes
* provides a path to adding future dependencies built with system compilers (e.g. PETSC)

To support Albany, we have moved away from E3SM supported compilers and modules on Badger.  In the future, it may make sense to update the E3SM configuration for Badger to match.

To do:
- [x] Remove `grizzly` and `cori-knl` from the list of machines
- [x] Update the documentation

PR #357, opened 2022-04-14
Automatically get `cores_per_node` on single node

xylar:
For non-HPC machines, it's typically fine to just detect the number of available cores.

This should reduce the amount of information needed in a user config file.


xylar:
@matthewhoffman, I don't think you need to do any testing of this.  It is only useful on a Mac or Linux machine.  I just wanted to flag you because I pulled this out of #353 for clarity.

PR #360, opened 2022-04-18
Copy MPAS model executable to work directory

### LINKED ISSUE #345: Consider copying rather than linking load script and executable
This would allow you to save a work directory and reuse it without having to keep the original compass branch.

xylar:
A directory (either `mpas-ocean` or `mpas-albany-landice`) is made within the base of the work directory and the MPAS model executable (`ocean_model` or `landice_model`) is copied there.  Symlinks within the work directory are to that copy.

This is useful if you want to delete the E3SM branch used to set up test cases or suites but you still want the ability to run the code after the fact.

closes #345

### DIFF
--- a/compass/default.cfg
+++ b/compass/default.cfg
@@ -1,3 +1,9 @@
+# The setup section defines options related to setting pu test cases or suites
+[setup]
+
+# whether to copy the executable to the work directory
+copy_executable = False
+
 # Options related to downloading files
 [download]
 
--- a/compass/setup.py
+++ b/compass/setup.py
@@ -15,7 +15,7 @@ from compass.job import write_job_script
 
 def setup_cases(tests=None, numbers=None, config_file=None, machine=None,
                 work_dir=None, baseline_dir=None, mpas_model_path=None,
-                suite_name='custom', cached=None):
+                suite_name='custom', cached=None, copy_executable=False):
     """
     Set up one or more test cases
 
@@ -29,7 +29,7 @@ def setup_cases(tests=None, numbers=None, config_file=None, machine=None,
         a suffix ``c`` to indicate that all steps in that test case should be
         cached
 
-    config_file : str, optional
+    config_file : {str, None}, optional
         Configuration file with custom options for setting up and running test
         cases
 
@@ -55,6 +55,9 @@ def setup_cases(tests=None, numbers=None, config_file=None, machine=None,
         For each test in ``tests``, which steps (if any) should be cached,
         or a list with "_all" as the first entry if all steps in the test case
         should be cached
+
+    copy_executable : bool, optional
+        Whether to copy the MPAS executable to the work directory
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


xylar:
## Testing

I ran the ocean `nightly` test case on my Linux laptop with this change.  I verified that the `ocean_model` executable was copied as desired and that the steps that use this executable are symlinked the copy, not to the original.

xylar:
@sbrus89, please verify that this meets you needs for #345.

xylar:
@matthewhoffman, I suspect this is not a functionality that you and the MALI team necessarily need.  Even if not, could you run a test to make sure it does no harm?

matthewhoffman:
@xylar , thanks for tagging me.  Does this mean that if you recompile the executable, you would now need to rerun `compass setup / suite` before re-running any tests?  I can understand the problem this PR is wanting to solve (and "freezing" the exe that was used for the tests could help with reproducibility), but if the answer to my question is yes, then it disrupts a different part of a common workflow.  For a debugging workflow where you are repeatedly editing code to get a test to stop failing, you would need an extra `compass setup` command after every time you recompile, which could be easily forgotten and lead to confusion.  What do you think?

xylar:
@matthewhoffman, that's right.  I'm glad I flagged you on this.  We should make this off by default and only on if a certain flag (say `--copy_executable`) is used.

matthewhoffman:
I'm not sure what the best solution is.  I could be convinced that this new way is a better overall design choice.  I think it's a tradeoff between convenience and reproducibility.  My preference is certainly the existing functionality for the reason I mentioned, but there might be other users for who the proposed functionality makes more sense.

xylar:
@matthewhoffman, I agree with you.  The more frequent need is to rebuild frequently to debug.  The ability to preserve the executable is not needed very often.  It was a request from @sbrus89 so it clearly could be useful, but need not be the default.

I believe I have set it up to be off by default, but on with `--copy_executable`.  @sbrus89, again, please try this out for yourself.

sbrus89:
I totally agree that copying the executable should not be the default behavior. The majority of the time, it's much better to have things linked for debugging as you've both mentioned. 

What I had in mind when I brought this up was using compass  set up a series of runs for a paper, for example. In that case it would probably be better to not have the potential for the executable to change behind a sim link. 

I think another use case would be to keep around a long-lived baseline test. In this case, I think it's nice to have the executable to produced the baseline results preserved.

matthewhoffman (approved):
I agree this is a useful option, and I like the solution of using `--copy_executable` as an optional flag.  I tested this branch without the flag (old behavior) and see no problems.  I also tested with the new `--copy_executable` flag.  Tests still pass and now I see the exe is copied into the working dir as expected.  Symlinks in the individual tests point to the copied exe in the root of the workdir.  So everything seems to work as expected.

I'm not sure if there is somewhere in the docs this should be mentioned, but the help messages in the `--help` output for compass setup and compass suite look good.

xylar:
Great. @sbrus89, if you can test this out, I'll merge it.

xylar:
> I'm not sure if there is somewhere in the docs this should be mentioned

I added a paragraph to the User's Guide (the Quick Start) that explains this flag and the motivation, using @sbrus' examples from above.

sbrus89 (approved):
@xylar. I have tested this and it works as expected. Thanks!

PR #363, opened 2022-04-18
Run steps as subprocesses by default

xylar:
This makes it easier to save content to log files for each step even when calling other subprocesses within the step that output to stdout/stderr (e.g. jigsaw).

### DIFF
--- a/compass/default.cfg
+++ b/compass/default.cfg
@@ -22,6 +22,9 @@ verify = True
 # the program to use for graph partitioning
 partition_executable = gpmetis
 
+# run steps as subprocesses (e.g with srun)
+run_steps_as_subprocesses = True
+
 
 # The io section describes options related to file i/o
 [io]
--- a/compass/run/serial.py
+++ b/compass/run/serial.py
@@ -223,10 +223,15 @@ def run_tests(suite_name, quiet=False, is_test_case=False, steps_to_run=None,
             sys.exit(1)
 
 
-def run_step():
+def run_step(step_is_subprocess=False):
     """
     Used by the framework to run a step when ``compass run`` gets called in the
     step's work directory
+
+    Parameters
+    ----------
+    step_is_subprocess : bool, optional
+        Whether the step is being run as a subprocess of a test case or suite
     """
     with open('step.pickle', 'rb') as handle:
         test_case, step = pickle.load(handle)
@@ -237,6 +242,10 @@ def run_step():
     config.add_from_file(step.config_filename)
 
     check_parallel_system(config)
+
+    # when we're running a single step, we definitely don't want to run it
+    # again as a subprocess
+    config.set('parallel', 'run_steps_as_subprocesses', 'False')
 
     test_case.config = config
     set_cores_per_node(test_case.config)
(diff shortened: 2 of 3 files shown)


matthewhoffman:
I tested this branch, and I find it sets up a test suite fine, but when I execute `compass run`, it seems to hang on the running MPAS.  It does the mesh set up step fine, but then gets stuck on the model run step:
```
$ compass run
landice/dome/2000m/sia_restart_test
  * step: setup_mesh
  * step: full_run
```

The case output has a suspicious line in it saying `Skipping validation because full_run and restart_run weren't  run` but before those step actually occur:

```
compass calling: compass.landice.tests.dome.restart_test.RestartTest.run()
  in /turquoise/usr/projects/climate/mhoffman/compass/compass/compass/testcase.py

Running steps: 

compass calling: compass.landice.tests.dome.restart_test.RestartTest.validate()
  in /turquoise/usr/projects/climate/mhoffman/compass/compass/compass/landice/tests/dome/restart_test/__init__.py

Skipping validation because full_run and restart_run weren't  run
compass calling: compass.landice.tests.dome.restart_test.RestartTest.run()
  in /turquoise/usr/projects/climate/mhoffman/compass/compass/compass/testcase.py

Running steps: setup_mesh, full_run, restart_run
  * step: setup_mesh
Running: srun -n 1 compass run
compass calling: compass.landice.tests.dome.restart_test.RestartTest.run()
  in /turquoise/usr/projects/climate/mhoffman/compass/compass/compass/testcase.py


compass calling: compass.landice.tests.dome.setup_mesh.SetupMesh.run()
  in /turquoise/usr/projects/climate/mhoffman/compass/compass/compass/landice/tests/dome/setup_mesh.py

[bunch of mesh gen output removed...]

compass calling: compass.landice.tests.dome.restart_test.RestartTest.validate()
  in /turquoise/usr/projects/climate/mhoffman/compass/compass/compass/landice/tests/dome/restart_test/__init__.py

Skipping validation because full_run and restart_run weren't  run

  * step: full_run
Running: srun -n 1 compass run

xylar:
Thanks @matthewhoffman.  What machine was that on?

It seems this needs some more testing.  I'm also seeing unexpected non-bit-for-bit results on Anvil.

matthewhoffman:
@xylar , I was testing on Badger.

xylar:
@matthewhoffman, thanks.  I'll make sure both ocean and land-ice test suites work for me on Badger before I ping you to try again.

xylar:
@matthewhoffman, I was able to reproduce the problem on Badger and I think I fixed it.  It seems you can't call `srun` from a process already launched by `srun` on Badger, whereas that's fine on Anvil.  (That may cause us trouble with task parallelism so it's good to know about.)

I have done some more testing on Badger and the ocean `nightly` suite is running file now.  In need to re-test on Anvil and other systems.  I'm also hoping this fixes the non-bit-for-bit results I saw on Anvil with Intel.

So I'm leaving this in draft mode for now.  Just wanted to keep you posted.

matthewhoffman:
Sounds good.  If you want me to re-test on Badger in the meantime, let me know.

xylar:
@matthewhoffman, I ran the MALI `full_integration` test suite, using the current master as a baseline on Badger and it worked so I don't think there's any need for you to retest just yet.  I need to clean up that confusing validation output for on thing.  It would be nice to understand the non-bit-for-bit behavior on Anvil, too, but I can't think of a good way to investigate yet.

xylar:
I did some more testing of this approach today.  It's just too slow to be used for every step -- on Anvil it sometimes doubles the runtime.  I'll a different approach where certain steps are run this way to capture logging but most aren't.

PR #364, opened 2022-04-19
Set `OPENMP=true` in load script by default

### LINKED ISSUE #347: Threaded tests should use `OPENMP=true` builds
The `OPENMP=true` make option is required in order to test threading, otherwise I believe the `!$omp` directives are treated as comments and the `OMP_NUM_THREADS` environment variable is ignored.

xylar:
A developer can call `./conda/configure_compass_env.py` with `--without_openmp` to prevent this variable from being set.

closes #347

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -366,7 +366,7 @@ def build_spack_env(config, update_spack, machine, compiler, mpi, env_name,
 
 def write_load_compass(template_path, activ_path, conda_base, env_type,
                        activ_suffix, prefix, env_name, spack_script, machine,
-                       env_vars, env_only, source_path):
+                       env_vars, env_only, source_path, without_openmp):
 
     try:
         os.makedirs(activ_path)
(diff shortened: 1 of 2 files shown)


xylar:
@matthewhoffman, could you let me know if it's okay for MALI to have `OPENMP=true` even if you don't actually have OpenMP directives in your code?  Is there overhead involved in doing that that you don't want to have?

xylar:
@mark-petersen, are you okay with this solution?  Or do you feel more comfortable with the current behavior in which users have to specify `OPENMP=true`?

xylar:
@sbrus89, please take a look if you can.

xylar:
## Testing

I made the build script once without `--no_openmp` and verified that:
```
export OPENMP=true
```
was in the load script.  I then used the load script to build MPAS-Ocean and verified that it also showed `OPENMP = true` at the beginning of the build.  Then, I rebuilt the load script, this time with `--no_openmp` and verified that `export OPENMP=true` was no longer present.  This time, when I build MPAS-Ocean, I see `OPENMP = ` at the beginning of the build.

xylar:
@mark-petersen, could you see if the modified (and rebased) version works for you and is more intuitive?

mark-petersen (approved):
Yes, I tested this and it works perfectly now. Thanks!

matthewhoffman (approved):
I tested the MALI integration suite and this appears to have no effect.  In my first test the integration suite took 3 minutes longer than usual after compiling with the OPENMP=true set by the compass env, but I think that was just machine variability, because I tried it a second time and it ran 2 minutes faster than usual.  There is no reason to think including this flag would affect anything, because MALI has no OpenMP directives in the MPAS code at present.

sbrus89 (approved):
@xylar, this works as expected for me too.

xylar:
Thanks everyone!  I'll resolve the conflict and then merge.

xylar:
Rebased and "tested" in then sense of creating a load script with and without OPENMP.

PR #365, opened 2022-04-20
Switch `CompassConfigParser` to descend from `MpasConfigParser`

xylar:
This way, config parsing functionality can be shared between MPAS-Analysis and copass.  See:
https://github.com/MPAS-Dev/MPAS-Tools/pull/460
https://github.com/MPAS-Dev/MPAS-Analysis/pull/873

The only functionality added to `CompassConfigParser` beyond what is in `MpasConfigParser` is to automatically convert config options in certain sections (`paths`, `namelists`, `streams`, and `executables`) from relative to absolute paths.

The documentation has been updated accordingly.  It points to MPAS-Tools' documentation but also explains most of the functionality in relation to compass.

The new functionality requires `mpas_tools >=0.14.0`.


xylar:
## Testing

I ran the ocean `pr` and MALI `full_integration` suites on Anvil, using the baseline I made for SCORPIO testing as the baseline here as well (so before the move to SCORPIO 1.3.2).  Both test suites passed and are bit-for-bit.

xylar:
@matthewhoffman, could you give this a quick look when you have time?  And maybe test it on Badger?

matthewhoffman (approved):
@xylar , this looks great to move most of this module to a shared MPAS-Tools version.  I did a MALI test and it works fine.  My only question is if the the one remaining additional piece of functionality in the CompassConfigParser couldn't also be moved to the MpasConfigParser - even if MPAS-Analysis would never use it.  It's not a big deal to have CompassConfigParser remain, but if it were possible to completely drop it, that might be a cleaner outcome to moving it to MPAS-Tools.  I'm leaving that to your judgment if that makes sense or not and approving now.

xylar:
>  My only question is if the the one remaining additional piece of functionality in the CompassConfigParser couldn't also be moved to the MpasConfigParser - even if MPAS-Analysis would never use it. It's not a big deal to have CompassConfigParser remain, but if it were possible to completely drop it, that might be a cleaner outcome to moving it to MPAS-Tools. I'm leaving that to your judgment if that makes sense or not and approving now.

I thought about that but I'm concerned that it would be an annoying hidden feature if some other code used `MpasConfigParser` in the future.  "Why is my config options that has nothing to do with a file path suddenly converted into a path?!?!"  I had this when I tried to add a True/False config option to `[executable]` myself.  I think it's much safer to leave it as a peculiarity specific to compass.  Maybe the solution would be to deal with it in a function called explicitly rather than in this relatively hidden way.  But I'm hesitant to add new, untested functionality at the moment...

PR #371, opened 2022-04-22
Skip permissions update if not needed

xylar:
When creating (or updating) the compass environment, it is not necessary to check or change permissions unless we are modifying the shared conda or spack environments.  This merge removes unnecessary attempts to change permissions.


xylar:
## Testing

I updated the compass environment.  First, I ran without `--update_spack` and saw no permission updates.  Then, I ran with `--update_spack` and saw the expected permission updates.

PR #372, opened 2022-04-22
Don't remove `esmf` from conda environment

xylar:
This update to the conda environment causes problems if it happens in the middle of running with the conda environment elsewhere.  (A likely time this might happen is when creating a build matrix with many compilers and/or MPI implementations, where you want to start running before all builds have completed.)  It was always intended as a "safety" measure to ensure that ESMF comes from spack when we want that.  There's no indication that isn't happening anyway.


xylar:
## Testing

I updated my compass environment on Anvil.  I saw the `esmf` was installed and not subsequently removed, as expected.

PR #373, opened 2022-04-22
Add utility for matrix builds

xylar:
Often, developers want to test a given set of test cases or a test suite with multiple compilers and configurations.  Doing this process manually can be  challenging and prone to mistakes.  The `setup_matrix.py` scrip is designed to automate this process with the help of a config file similar to  `example.cfg`.  See the `README.md` added as part of this PR.


xylar:
I've tested this some but it seems prohibitively slow to do each build one after the other.  Building each configuration in parallel with each other, on the other hand, is going to be tricky, requiring a different conda environment and worktree for each.

Also, it seems like we need to generate a job script for each setup.  Otherwise, it's too hard to run everything in interactive jobs.  I think I might want to introduce that capability before the build matrix.

xylar:
@sbrus89, I will continue to test this but it seems to be working.  I am going to use it to test a different branch where I will be adding PETSC and LAPACK.  If you have time to review and give feedback, this PR is the highest priority for me among the ones where I have asked for your review.

I know you're still recovering so no pressure, just wanted to let you know which is a priority for me.

utils/matrix/setup_matrix.py
```diff
@@ -0,0 +1,211 @@
...
+                else:
+                    suffix = '{}_noopenmp'.format(suffix)
+                    make_command = '{} OPENMP=false'.format(make_command)
+                if use_debug:
+                    suffix = '{}_debug'.format(suffix)
+                    make_command = '{} DEBUG=true'.format(make_command)
+                else:
+                    make_command = '{} DEBUG=false'.format(make_command)
+                mpas_model = build_mpas(
+                    script_name, mpas_path, make_command, suffix)
```
sbrus89, line 122: It seems like the `CORE=` in the make command isn't necessary anymore, right? I've always used it, but I bet that's just a habit I've carried over from before the cores got transplanted to E3SM.

utils/matrix/setup_matrix.py
```diff
@@ -0,0 +1,211 @@
...
+    'cori-haswell': {
+        ('intel', 'mpt'): 'intel-nersc',
+        ('gnu', 'mpt'): 'gnu-nersc'},
+    'conda-linux': {
+        ('gfortran', 'mpich'): 'gfortran',
+        ('gfortran', 'openmpi'): 'gfortran'},
+    'conda-osx': {
+        ('gfortran-clang', 'mpich'): 'gfortran-clang',
+        ('gfortran-clang', 'openmpi'): 'gfortran-clang'}
+}
```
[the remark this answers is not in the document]
xylar: No, this information is very specific to MPAS so doesn't belong in `mache`.  It could live in `compass` but then it wouldn't be available to this utility, which lives firmly outside of the compass package.  For now, this info only lives in the documentation besides this dictionary.

utils/matrix/setup_matrix.py
xylar: It isn't needed since the code moved over to E3SM.  The `CORE` is automatically determined based on the directory you're building from as far as I understand it.

sbrus89 (approved):
@xylar, this worked wonderfully for me. It's going to be particularly awesome when combined with #376. 

The only other thing I'll mention is that it would be nice to have a `--submit` to kick off each job in the matrix, once this has been merged and the job script capability is available.

utils/matrix/setup_matrix.py
```diff
@@ -0,0 +1,211 @@
...
+
+# build targets from
+# https://mpas-dev.github.io/compass/latest/developers_guide/machines/index.html#supported-machines
+all_build_targets = {
+    'anvil': {
+        ('intel', 'impi'): 'intel-mpi',
+        ('intel', 'openmpi'): 'ifort',
+        ('intel', 'mvapich'): 'ifort',
+        ('gnu', 'openmpi'): 'gfortran',
+        ('gnu', 'mvapich'): 'gfortran'},
```
sbrus89, line 30: Is there some way we could choose a subset of these in the config file?

sbrus89:
Also, just thinking out loud here: could the build for each member be included in the job script?

utils/matrix/setup_matrix.py
xylar: You don't need to choose in the config file.  You choose by how you call `./conda/configure_compass_env.py`.  You can do:
```
./conda/configure_compass_env.py --compiler all --mpi all ...
```
to get all of these.  Or you can do:
```
./conda/configure_compass_env.py --compiler intel --mpi all ...
```
to get just the intel ones or:
```
./conda/configure_compass_env.py --compiler all --mpi openmpi ...
```
to get just the openmpi ones or:
```
./conda/configure_compass_env.py --compiler intel ...
```
(omitting the `--mpi`) to get the default MPI library for intel or:
```
./conda/configure_compass_env.py --compiler intel intel intel gnu --mpi impi openmpi mvapich mvapich ...
```
to get a set with 3 intel variants and a gnu variant.

Do you want to make a suggestion for how to add this to the README here?  Or the docs somewhere else?  I wasn't sure how to handle it because it's a feature I expect most developers not necessarily to need (so I didn't want to overwhelm the quick start for developers with this) but it isn't directly related to the matrix utility in that it should work fine even if you build, setup and run manually.

xylar:
> Also, just thinking out loud here: could the build for each matrix member be included in the job script?

It could be but if you submitted all the jobs at once, they would mess each other up.  The build can't happen in parallel in the same directory.

xylar:
> The only other thing I'll mention is that it would be nice to have a --submit to kick off each job in the matrix, once this has been merged and the job script capability is available.

I like that idea very much.  I don't see any downside since you should be able to override the default wall-clock time, account, etc. if the deafults don't make sense.

utils/matrix/setup_matrix.py
xylar: I will add something similar to my comment above to the README for now.  It can make its way to the main docs later...

xylar:
@sbrus89, I added more documentation (including docstrings, oops!).  I also added the `--submit` flag but I'm not 100% sure the logic for detecting the suite name will be correct.  I'll test as I have time (but not at 10 pm).  If you have time to test, feel free.  But be warned that bugs await you if past experience is any guide...

utils/matrix/setup_matrix.py
sbrus89: This is definitely helpful. My thought was that just because you have these environments created, doesn't mean you want to necessarily include them in the matrix. Can you edit the `matrix.log` file to exclude certain ones?

xylar: You could do that but maybe just call the config script again? It won't delete anything, just make sure everything is up to date and write the matrix.log you want.

xylar: I will think about this some more...

sbrus89: That's a good idea.

sbrus89:
@xylar, I tried out the `--submit` option at the very end of yesterday and checked this morning to see that all the matrix jobs ran. This is super cool!

PR #376, opened 2022-04-25
Write out a job script for suite and test cases during setup

xylar:
On supported machines, write out a job script during setup.  The job script uses `mache` to figure out what account, partition, QOS and constraint to use.  It defaults to 1 hour wall-clock time (typically, more than enough) and determines the number of nodes based on the geometric mean (often a practical compromise) of the minimum and target number of cores required for a test case or suite.

Users (and developers) can alter the defaults with a config file, as detailed in the updated User's Guide.

### DIFF
--- a/compass/job/__init__.py
+++ b/compass/job/__init__.py
@@ -4,7 +4,8 @@ import os
 import numpy as np
 
 
-def write_job_script(config, machine, target_cores, min_cores, work_dir):
+def write_job_script(config, machine, target_cores, min_cores, work_dir,
+                     suite=''):
     """
 
     Parameters
@@ -24,6 +25,9 @@ def write_job_script(config, machine, target_cores, min_cores, work_dir):
 
     work_dir : str
         The work directory where the job script should be written
+
+    suite : str, optional
+        The name of the suite
     """
 
     if config.has_option('parallel', 'account'):
@@ -70,6 +74,11 @@ def write_job_script(config, machine, target_cores, min_cores, work_dir):
             constraint = ''
 
     job_name = config.get('job', 'job_name')
+    if job_name == '<<<default>>>':
+        if suite == '':
+            job_name = 'compass'
+        else:
+            job_name = f'compass_{suite}'
     wall_time = config.get('job', 'wall_time')
 
     template = Template(resources.read_text(
@@ -77,7 +86,12 @@ def write_job_script(config, machine, target_cores, min_cores, work_dir):
 
     text = template.render(job_name=job_name, account=account,
                            nodes=f'{nodes}', wall_time=wall_time, qos=qos,
-                           partition=partition, constraint=constraint)
-    script_filename = os.path.join(work_dir, 'job_script.sh')
+                           partition=partition, constraint=constraint,
+                           suite=suite)
+    if suite == '':
+        script_filename = 'compass_job_script.sh'
+    else:
+        script_filename = f'compass_job_script.{suite}.sh'
+    script_filename = os.path.join(work_dir, script_filename)
     with open(script_filename, 'w') as handle:
         handle.write(text)
--- a/compass/job/__init__.py
+++ b/compass/job/__init__.py
@@ -1,0 +1,83 @@
+from jinja2 import Template
+from importlib import resources
+import os
+import numpy as np
+
+
+def write_job_script(config, machine, target_cores, min_cores, work_dir):
+    """
+
+    Parameters
+    ----------
+    config : compass.config.CompassConfigParser
+        Configuration options for this test case, a combination of user configs
+        and the defaults for the machine and MPAS core
+
+    machine : {str, None}
+        The name of the machine
+
+    target_cores : int
+        The target number of cores for the job to use
+
+    min_cores : int
+        The minimum number of cores for the job to use
+
+    work_dir : str
+        The work directory where the job script should be written
+    """
+
+    if config.has_option('parallel', 'account'):
+        account = config.get('parallel', 'account')
+    else:
+        account = ''
+
+    cores_per_node = config.getint('parallel', 'cores_per_node')
+
+    # as a rule of thumb, let's do the geometric mean between min and target
+    cores = np.sqrt(target_cores*min_cores)
+    nodes = int(np.ceil(cores/cores_per_node))
+
+    partition = config.get('job', 'partition')
+    if partition == '<<<default>>>':
+        if machine == 'anvil':
+            # choose the partition based on the number of nodes
+            if nodes <= 5:
+                partition = 'acme-small'
+            elif nodes <= 60:
+                partition = 'acme-medium'
+            else:
+                partition = 'acme-large'
+        elif config.has_option('parallel', 'partitions'):
+            # get the first, which is the default
+            partition = config.getlist('parallel', 'partitions')[0]
+        else:
+            partition = ''
+
+    qos = config.get('job', 'qos')
+    if qos == '<<<default>>>':
+        if config.has_option('parallel', 'qos'):
+            # get the first, which is the default
+            qos = config.getlist('parallel', 'qos')[0]
+        else:
+            qos = ''
+
+    constraint = config.get('job', 'constraint')
+    if constraint == '<<<default>>>':
+        if config.has_option('parallel', 'constraints'):
+            # get the first, which is the default
+            constraint = config.getlist('parallel', 'constraints')[0]
+        else:
+            constraint = ''
+
+    job_name = config.get('job', 'job_name')
+    wall_time = config.get('job', 'wall_time')
+
+    template = Template(resources.read_text(
+        'compass.job', 'template.sh'))
+
+    text = template.render(job_name=job_name, account=account,
+                           nodes=f'{nodes}', wall_time=wall_time, qos=qos,
+                           partition=partition, constraint=constraint)
+    script_filename = os.path.join(work_dir, 'job_script.sh')
+    with open(script_filename, 'w') as handle:
+        handle.write(text)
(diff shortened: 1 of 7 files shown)


xylar:
## Testing

I am in the process of running a couple of tests on each machine.  

I will check off here as I successfully run the ocean `pr` test suite on each machine:
- [ ] Anvil
- [ ] Badger
- [ ] Chrysalis
- [ ] Compy
- [ ] Cori-Haswell

I will check off here when I have successfully run the MALI `full_integration` suite on each machine:
- [ ] Anvil
- [ ] Badger
- [ ] Chrysalis
- [ ] Compy
- [ ] Cori-Haswell

compass/job/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    if config.has_option('parallel', 'account'):
+        account = config.get('parallel', 'account')
+    else:
+        account = ''
+
+    cores_per_node = config.getint('parallel', 'cores_per_node')
+
+    # as a rule of thumb, let's do the geometric mean between min and target
+    cores = np.sqrt(target_cores*min_cores)
+    nodes = int(np.ceil(cores/cores_per_node))
```
xylar, line 38: The geometric mean seems to be a good guess for what folks might want but I'm up for other suggestions for how to handle the node count.

xylar:
The account on Badger (from `mache`) is `e3sm`.  That account no longer exists but there isn't an obvious default.  I think that's a case where a little config file is the best option:
```
$ cat t22_ocean_time_step.cfg 
[parallel]

account = t22_ocean_time_step
```

```
$ compass suite -s -f t22_ocean_time_step.cfg ...
```

xylar:
@sbrus89, @mark-petersen and @matthewhoffman, I'm still finishing my testing on Cori and Compy but this is ready for you to test and/or review when you want.

If any of you feel like this isn't a feature you would use anyway, no need to review it.

mark-petersen (approved):
This works great, I love it! It will really simplify my process, as I usually want to run many tests simultaneously and can only run one interactive job. Thanks also for the flag for the job account. I tested on Badger with and without the flag, as follows: 
```
./conda/configure_compass_env.py --conda /usr/projects/climate/mpeterse/miconda3 \
--compiler gnu --env_name job_script
source load_job_script_badger_gnu_mvapich.sh
compass suite -s -c ocean -t nightly -p $EXEDIR -w $WORKDIR -f t22_ocean_time_step.cfg
cd $WORKDIR
sbatch job_script.sh
# wait...
cat compass.o*
```

On LANL IC machines, let's use the `t22_ocean_time_step` as the default. I own that, and I'm happy to add anyone for COMPASS testing, which is small change (I think I've already added most ocean users). That is better than `e3sm` which does not exist for anyone. When I apply next year, I'll choose a more inclusive account name.

compass/provenance.py
```diff
@@ -110,27 +101,9 @@ def write(work_dir, test_cases, mpas_core=None, config_filename=None,
     provenance_file.close()
 
 
-def _get_mpas_git_version(mpas_core, config_filename, mpas_model_path):
```
matthewhoffman, line 113: It looks like this section was moved to a different file.  This seems somewhat unrelated to the PR description, so just checking this was intentional to include in this PR.  If it's just a convenience thing for planned future changes, that's fine with me.

xylar: I did pull this out into its own PR: https://github.com/MPAS-Dev/compass/pull/378

matthewhoffman: I was fine with it being included in this PR, but pulling it out is fine too.

compass/machines/badger.cfg
```diff
@@ -37,6 +37,13 @@ spack = /usr/projects/climate/SHARED_CLIMATE/compass/badger/spack
 use_e3sm_hdf5_netcdf = False
 
 
+# The parallel section describes options related to running jobs in parallel
+[parallel]
+
+# account for running diagnostics jobs
+account = t22_ocean_time_step
```
matthewhoffman, line 44: It makes sense to have the default be the account that the most people have access to.  Trevor, Holly, and I won't be able to use this for MALI tests (unless Mark wanted to add us, but I wouldn't expect that to happen just for this reason).  It would be convenient if for us we could default to `t22_iceocean`, but I expect making a core-specific default is not worth the trouble.  A different option is that if the account is *not* included at all in the job script, SLURM will use the user's default account (which for me is set to `t22_iceocean`).  The only downside there is if a new user has not set a default account, but I would suggest it makes sense for users to do that on LANL IC.  So I guess my preference would be for account to not be included in the Badger job script.

  @xylar , @mark-petersen , what do you think about this issue?

compass/job/template.sh
```diff
@@ -0,0 +1,22 @@
...
+{%- endif %}
+{% if partition != '' -%}
+#SBATCH  --partition={{ partition }}
+{%- endif %}
+{% if constraint != '' -%}
+#SBATCH  --constraint={{ constraint }}
+{%- endif %}
+
+source load_compass_env.sh
+compass run
```
matthewhoffman, line 21: We could add the name of the pickle file to run here.  If you have multiple suites set up in one directory, then as written this will fail.  That said, it's reasonable enough to consider this submission script as a template and expect that a user can/will adjust it for the specific suite/pickle they want to run.  If we do insert the pickle name here, then we probably want to put that in the script name as well so that setting up multiple suites doesn't clobber existing scripts and so it's clear that you are submitting a script specific to a particular suite/pickle.

compass/job/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    job_name = config.get('job', 'job_name')
+    wall_time = config.get('job', 'wall_time')
+
+    template = Template(resources.read_text(
+        'compass.job', 'template.sh'))
+
+    text = template.render(job_name=job_name, account=account,
+                           nodes=f'{nodes}', wall_time=wall_time, qos=qos,
+                           partition=partition, constraint=constraint)
+    script_filename = os.path.join(work_dir, 'job_script.sh')
```
matthewhoffman, line 81: How about adding 'compass' in the name?  And then possibly the suite name or test case number (see my other comment)?  E.g. `compass_job_script.suite_full_integration.sh` or `compass_job_script.test_61.sh`. If that's too complicated or requires too many changes, it's fine as it.

matthewhoffman (changes requested):
@xylar , this is a fantastic quality of life addition, and I was actually thinking about this in the last few days!  I made a few comments about some of the details of the implementation.  I'm not committed to any of them, but wanted to have a little bit more discussion about expected behavior and conventions before this PR gets finalized.

compass/machines/badger.cfg
xylar: I agree, no default account is best.  I have taken it out (left it blank).

compass/job/template.sh
xylar: Oh, I can't believe I didn't think of that. I have added it now.  It's pretty obvious that we want that feature.  It isn't included for test case job scripts but it part of suite ones.

compass/job/__init__.py
xylar: I agree that having the suite name makes total sense.  I think the suite name will just be `custom` when you set up test cases with numbers.  I don't think `test_61_62_63` or anything like that is very helpful.  I'm not sure if that information is easily available by the time the job script is being created either.

xylar: > How about adding 'compass' in the name?

I did this but I'm not sure why it's important.  I would think the compass work directory would implicitly only be for doing stuff with compass.

xylar:
@matthewhoffman, I believe I've addressed your requested changes.  Please give it another look.

I also realized that I had made an error in generating the job scripts for individual tests -- those job scripts were ending up in the base work directory instead of the tests themselves.  I fixed this and have now tested that you can run a test case on its own as well.

xylar:
Taking @sbrus89 off as a reviewer since he's been out sick.

compass/machines/badger.cfg
xylar: > It would be convenient if for us we could default to `t22_iceocean`, but I expect making a core-specific default is not worth the trouble.

This would be possible but hard to maintain.  There isn't currently a good way to have config options for an MPAS core and machine.  We could add `account_ocean` and `account_landice` config options that get used if `account` is blank.  But the `t22_*` accounts are obviously a moving target whereas the default account seems easy.  If folks haven't set a default, they will just have to weight the relative pain of providing a config file or setting a default.

sbrus89 (approved):
@xylar, this worked great for me running the nightly suite on Compy.

sbrus89:
This also worked for me on a single test case: `compass setup -t ocean/soma/32km/default`

xylar:
Thanks, @sbrus89! @matthewhoffman, could you take another look and update your review if you're happy?

matthewhoffman (approved):

matthewhoffman:
@xylar , I took another look and the new behavior addresses all the issues I brought up.  I tried a couple suites and one single test case.  

One side question - is it normal to have `compass_job_script.custom.sh` and `custom.pickle` appear in the workdir root when setting up a single test?  I remember the concept of custom suites, but I've never explored it myself.

Also, if you want to drop `compass_` from the job script name, that's fine with me.  With the SLURM log files also starting with `compass`, tab completion is less efficient, but that is certainly a minor consideration.

sbrus89:
I was wondering about the `compass_job_script.custom.sh` and `custom.pickle` as well. Will those always be specific to the most recent single test case setup?

xylar:
> One side question - is it normal to have compass_job_script.custom.sh and custom.pickle appear in the workdir root when setting up a single test? I remember the concept of custom suites, but I've never explored it myself.

Yes, that's the normal behavior.  A test case or several of them set up with `compass setup` is considered a "custom" suite.

xylar:
> I was wondering about the compass_job_script.custom.sh and custom.pickle as well. Will those always be specific to the most recent single test case setup?

Yes.  In general, you should use different work directories for setting up different sets of test cases (so that the custom suite doesn't get overwritten).

@sbrus89 and @matthewhoffman, are you both using the same work directory for setting up several test cases or suites, one after the other?

xylar:
Thanks @mark-petersen, @matthewhoffman and @sbrus89 for the reviews!  I hope this is a useful new feature.

PR #378, opened 2022-04-26
Create basic config for an MPAS core

xylar:
This is needed for both provenance and test cases, and will also be needed for job scripts.

For provenance, it is necessary to have config options that come from most sources (defaults, mache, machine-specific, MPAS-core and user) but where a test group and test case has not yet been determined so those are not included.

Previously, the framework redundantly made new "basic" config options as a starting point for each test case, adding more config options for the test group and test case, and to call the test case's `configure()` method.

That code was duplicated because it initially seemed trivial but slowly got more complex. For #376, I needed this "basic" set of config options in a 3rd place (for making job scrips) and it was clearly time to fix the unnecessary duplicate code.

### DIFF
--- a/compass/provenance.py
+++ b/compass/provenance.py
@@ -5,8 +5,7 @@ import subprocess
 from compass.config import CompassConfigParser
 
 
-def write(work_dir, test_cases, mpas_core=None, config_filename=None,
-          mpas_model_path=None):
+def write(work_dir, test_cases, config=None):
     """
     Write a file with provenance, such as the git version, conda packages,
     command, and test cases, to the work directory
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I am in the process of running the ocean `nightly` test suite on my laptop.  All is bit-for-bit so far.  I will update when it has finished.

xylar:
@matthewhoffman, please give this a quick look when you have a chance.

xylar:
@matthewhoffman, I'm going to go ahead with this one so as not to distract from #376 and to not overwhelm you (any more than I have to at least) with reviews.

matthewhoffman:
Ok, thanks! I glanced at it last night and had no concerns to the extent I followed the details.  I went hunting for it now and found it already merged, which was nice!

PR #379, opened 2022-04-27
Optionally build load scripts for multiple compilers/MPIs at the same time

xylar:
When you configure the compass environment, you can pass a list of compilers and a corresponding list of MPI libraries, or you
can pass "all" for one or both of these.  If you leave either blank (or both), the default for the machine is used.  If you pass "all" for either or both, only supported compiler/MPI pairs are used.

This merge also adds some missing combination of machines, compiler and MPI libraries to the list of unsupported configurations.

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -14,7 +14,7 @@ from jinja2 import Template
 from importlib.resources import path
 from configparser import ConfigParser
 
-from mache import discover_machine
+from mache import discover_machine, MachineInfo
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I have tested the `--compiler all` and `--mpi all` options on all 5 supported machines and my Linux laptop.  I also tried various other combinations of flags (`all` for only one of `--compiler` or `--mpi` and lists for each as well as omitting one or both of these flags).  Things seem to work as expected.

PR #382, opened 2022-05-04
Fix missing output for daily stats test

xylar:
The missing output causes validation to fail.


xylar:
## Testing

I ran the `daily_output_test` on my Linux laptop.  It failed validation before this fix but passes with it.

PR #385, opened 2022-05-05
Add Kangerlussuaq test group and mesh generation test case

trhille:
Add a test group and mesh generation test case for Kangerlussuaq, Greenland. This uses mostly the same workflow, datasets, and config options as the Humboldt mesh generation test case.


trhille:
The mesh generation only takes about 30 seconds, compared with ~3.5 minutes for Humboldt. The default 1–10km Kangerlussuaq mesh contains 8715 cells. The resulting mesh gives high resolution where desired and preserves the fjord geometry. An older version of the Kangerlussuaq.geojson file in 
image
Note: An older version of the Kangerlussuaq.geojson file in https://github.com/MPAS-Dev/compass/pull/385/commits/b1f7a64862af0e129d2e90010a786c45b175562c creates a smaller mesh, but risks degrading quality at the fjord walls.

trhille:
The 250m–10km mesh (93k cells) only took 5 minutes to run, and looks good:
image

matthewhoffman:
@trhille , this looks good.  I didn't notice any issues requiring tweaking in looking over the code, and I was able to easily run this on Badger in 35 seconds.  

I'm going to note for posterity two issues we discussed earlier today:
1. We eventually may be able to generalize all of these mesh gen cases into a single test case that just takes some filenames and x,y coordinates as cfg options to create any mesh in the same workflow.  We decided we don't want to hold up this round of mesh generation setups for that but we will tackle that the next time we need to put one together.
2. We noted that the default thickness interpolation transitions from zero thickness to correct thickness at the terminus over about 3 cells.  However, currently we are using a separate interpolation processing workflow that deals with this much better using masks.  That will eventually be brought into COMPASS once we generalize the process more, but until then this (and other) mesh creation is really for reproducibly creating the mesh itself, not the entire initial condition.
image

matthewhoffman (approved):

xylar:
@trhille and @matthewhoffman, I think you've got the hang of compass and I don't need to review this.  But if there's anything specific you want my feedback on, let me know.

PR #387, opened 2022-05-06
Update default thwaites high_res_mesh config options

trhille:
Update default mesh config options to use ice speed and distance to grounding line, but not distance to ice ege in defining cell spacing. Testing indicates this gives a slightly cleaner mesh. Also rename mesh file from Thwaites_1to8km.nc to Thwaites.nc to prevent confusion when using non-default mesh resolution config options.


xylar (approved):
Yes, the change in the mesh name definitely makes sense. I'll trust your judgment on the config option change.

Would you like me to do any testing?

trhille:
> Would you like me to do any testing?

No need, unless you want to!

PR #392, opened 2022-05-17
Add PETSc and Netlib-LAPACK to Spack

xylar:
Similarly to Albany, these are optional and can be added with `--with_netlib_lapack` and `--with_petsc` flags to `./conda/configure_compass_env.py`.

A new version of `mache`, 1.4.1, is required to support this. Separate spack environments are created depending on whether Albany, Netlib-LAPACK and/or PETSc are included.  This is because:
1. The default environment uses optimized LAPACK (usually from MLK), which is incompatible with Netlib-LAPACK
2. PETSc and Trilinos are monsters, which make loading the Spack environment pretty slow, so we don't want to include them if they aren't needed
This makes for a bit of extra deployment work for me (offset by the recent work to support a matrix of compilers and MPI libraries).

### DIFF
--- a/compass/version.py
+++ b/compass/version.py
@@ -1,1 +1,1 @@
-__version__ = '1.1.0-alpha.3'
+__version__ = '1.1.0-alpha.4'
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -335,6 +335,7 @@ def build_spack_env(config, update_spack, machine, compiler, mpi, spack_env,
 
     albany = config.get('deploy', 'albany')
     esmf = config.get('deploy', 'esmf')
+    lapack = config.get('deploy', 'lapack')
     petsc = config.get('deploy', 'petsc')
     scorpio = config.get('deploy', 'scorpio')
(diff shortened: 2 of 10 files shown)


xylar:
## Testing

I am in the process of deploying all 13 configurations:
- [ ]  Anvil intel impi
- [ ]  Anvil intel openmpi
- [ ]  Anvil intel mvapich
- [ ]  Anvil gnu openmpi
- [ ]  Anvil gnu mvapich
- [ ]  Badger intel impi
- [ ]  Badger gnu mvapich
- [x]  Chrysalis intel impi
- [x]  Chrysalis intel openmpi
- [x]  Chrysalis gnu openmpi
- [ ]  Compy intel impi
- [x]  Cori-Haswell intel mpt
- [x]  Cori-Haswell gnu mpt

xylar:
Since, I bumped the build number and the `mache` version, I'm also updating all Albany builds and all standard builds without Albany, PETSc or Netlib-LAPACK.

PR #393, opened 2022-05-18
Add cfchecker and udunits2 to conda environment

xylar:
These are handy for checking CF compliance of MPAS output.


xylar:
## Testing

I verified that `cfchecks` and `udunits2` could both be run from an environment I created using this branch.

PR #395, opened 2022-05-19
Ignore fill values during variable validation

xylar:
The consequence is that validation will not notice situations where one array has fill values and the other does not.

### DIFF
--- a/compass/validate.py
+++ b/compass/validate.py
@@ -299,6 +299,8 @@ def _compute_norms(da1, da2, quiet, max_l1_norm, max_l2_norm, max_linf_norm,
 
     result = True
     diff = numpy.abs(da1 - da2).values.ravel()
+    # skip entries where one field or both are a fill value
+    diff = diff[numpy.isfinite(diff)]
 
     l1_norm = numpy.linalg.norm(diff, ord=1)
     l2_norm = numpy.linalg.norm(diff, ord=2)


xylar:
## Testing

I ran the QU240 performance and daily stats test cases with this compass branch and https://github.com/mark-petersen/E3SM/tree/5efe7476b7244f7ce047e315d321aa5acef17cbf (the head of https://github.com/mark-petersen/E3SM/commits/mark-petersen/ocn/cf_compliant_activeTracer as of this testing).  The results showed fill values in the output of both test cases.  A comparison to a baseline from compass master and the current E3SM-Project submodule on compass master passed for both test cases, suggesting that fill values are being ignored as desired.

xylar:
@matthewhoffman and @mark-petersen, this branch would be very useful for the hackathon.  If you are willing to approve by inspection and based on my testing, that would be great!  If you prefer to do testing of your own, could you do so in the next 3 or 4 days so we can merge this well before the Hackathon is over?

mark-petersen:
Using the code base at https://github.com/E3SM-Ocean-Discussion/E3SM/pull/20 with this PR for compass, I am able to correctly compare two runs with the tracer fill values implemented, i.e. the nightly suite with baseline comparison passes. Internal comparisons fail for
```
00:55 FAIL ocean_global_ocean_QU240_PHC_decomp_test
00:55 FAIL ocean_global_ocean_QU240_PHC_threads_test
00:55 FAIL ocean_global_ocean_QU240_PHC_RK4_decomp_test
```
and I am not sure if that is a true mismatch with the output, or a problem with the comparison script, but I suspect it is the first. Results of that comparison are here on LANL IC if that is useful: 
```
/lustre/scratch5/turquoise/mpeterse/runs/n/ocean_model_220519_275abdf5_ba_intel-mpi_19_openmp_cf_compliant_activeTracer_PR1_compare
```

mark-petersen:
@xylar, testing this on https://github.com/E3SM-Ocean-Discussion/E3SM/pull/20, it is able to correctly compare two runs. It still fails the internal comparison for intel optimized for decomp and thread tests in stand-alone, as noted here https://github.com/E3SM-Ocean-Discussion/E3SM/pull/20/#issuecomment-1137824581.

compass/validate.py
```diff
@@ -299,6 +299,8 @@ def _compute_norms(da1, da2, quiet, max_l1_norm, max_l2_norm, max_linf_norm,
 
     result = True
     diff = numpy.abs(da1 - da2).values.ravel()
+    # skip entries where one field or both are a fill value
+    diff = diff[numpy.isfinite(diff)]
```
mark-petersen, line 303: @xylar the fill value is not a NAN in the implementation I tested, it is 
```
temperature:_FillValue = 9.96920996838687e+36 ;
```
so I think the `isfinite` command will still operate on those values, since they are not Nans.

xylar: @mark-petersen, `xarray` converts fill values to `NaN` when it opens the data set.  I will verify this using https://github.com/E3SM-Ocean-Discussion/E3SM/pull/20 but it has *always* been the case for me in the past.  So I still believe this to be the correct implementation.

xylar:
> It still fails the internal comparison for intel optimized for decomp and thread tests in stand-alone

@mark-petersen, I do not see how changing compass validation (which does not alter forward runs in any way) could be the cause of failing decomp and thread tests.  I still think that must come from somewhere else (probably https://github.com/E3SM-Ocean-Discussion/E3SM/pull/20).

compass/validate.py
mark-petersen: Thanks for the explanation. I understand now.

mark-petersen (approved):
I agree with this change, so that nan values in the land cells are ignored.

xylar:
## Testing

I did some more testing using https://github.com/E3SM-Ocean-Discussion/E3SM/pull/20.

If I don't use this compass branch, I get erroneous passes like this:
```
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
       L_Infinity: 0.00000000000000e+00
temperature          Time index: 0, 1, 2
0:  l1:              nan  l2:              nan  linf:              nan 
1:  l1:              nan  l2:              nan  linf:              nan 
2:  l1:              nan  l2:              nan  linf:              nan 
  PASS
/home/xylar/data/mpas/test_20220525/tracer_fill3/ocean/global_ocean/QU240/PHC/performance_test/forward/output.nc
/home/xylar/data/mpas/test_20220525/qu240_baseline/ocean/global_ocean/QU240/PHC/performance_test/forward/output.nc
```
(Usually, you don't see the `nan` values because it runs quietly.)

These erroneous passes happen because the threshold is always less than `NaN` for any threshold.

With this branch, I see:
```
    Pass thresholds are:
       L1: 0.00000000000000e+00
       L2: 0.00000000000000e+00
       L_Infinity: 0.00000000000000e+00
temperature          Time index: 0, 1, 2
0:  l1: 8.25744384216112e-11  l2: 1.54943056359418e-12  linf: 2.70006239588838e-13 
1:  l1: 2.89371364787768e-10  l2: 7.35238263124532e-12  linf: 3.15836246045365e-12 
2:  l1: 5.30992973874454e-10  l2: 1.33675073558272e-11  linf: 7.47490958019625e-12 
  FAIL
/home/xylar/data/mpas/test_20220525/tracer_fill3/ocean/global_ocean/QU240/PHC/performance_test/forward/output.nc       /home/xylar/data/mpas/test_20220525/qu240_baseline/ocean/global_ocean/QU240/PHC/performance_test/forward/output.nc
```
In this particular test, these differences are expected because I ran one test in optimized mode and one in debug mode.

xylar:
@matthewhoffman, I'm going to go ahead and merge this.  I'm pretty convinced it works as expected and I think @mark-petersen's review is enough.

Thanks @mark-petersen!

PR #396, opened 2022-05-23
Add a flag to prevent the compass package from being reinstalled with the load script

xylar:
This merge adds a check for an environment variable `$NO_COMPASS_REINSTALL` before reinstalling compass in edit mode from the current branch.

This is used in calls to `compass setup` within a matrix build.  Without this change, some runs might get disrupted because the `compass` executable or package is not available for a short time during reinstallation.

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -502,7 +502,8 @@ def write_load_compass(template_path, activ_path, conda_base, env_type,
     if env_type == 'dev':
         update_compass = \
             """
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I ran a matrix on Anvil.  I ran into some issues but none appear to be related to compass not being available.

I also tested setting the environment variable manually, then sourced the load script, and verified that the package was not reinstalled, as desired.

PR #397, opened 2022-05-23
Add ismip6 run testgroup

hollyhan:
This PR adds a test group on running Antarctic simulations for the ISMIP6 2300 experimental protocol [https://www.climate-cryosphere.org/wiki/index.php?title=ISMIP6-Projections2300-Antarctica](url).  

Each experiment is a step that is meant to be run manually.  A user must specify input files and experiments to set up in the config file.

Checklist


compass/landice/tests/ismip6_run_ais/run_model.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+
+        self.mesh_file = None
+        self.mesh_type = mesh_type
+
+        super().__init__(test_case=test_case, name=name)
+
+    def setup(self, velo_solver="FO"):
+        config = self.config
+        section = config['ismip6_run_ais']
```
matthewhoffman, line 39: Can you add a `.cfg` file to the branch?

compass/landice/tests/ismip6_run_ais/run_model.py
```diff
@@ -0,0 +1,148 @@
...
+                               'specified_calving_velocity',
+                               'von_Mises_stress',
+                               'damagecalving', 'ismip6_retreat']:
+            raise ValueError("Value of calving_law must be one of {'none', "
+                             "'floating', "
+                             "'eigencalving', 'specified_calving_velocity', "
+                             "'von_Mises_stress', 'damagecalving', "
+                             "'ismip6_retreat'}")
+
+        self.add_input_file(filename='albany_input.yaml',
```
matthewhoffman, line 55: This file needs to be added to the branch.

compass/landice/tests/ismip6_run_ais/run_model.py
```diff
@@ -0,0 +1,148 @@
...
+        self.add_input_file(filename=self.mesh_file,
+                            target=os.path.join(base_path_mali,
+                                                self.mesh_file))
+
+        # Todo: upload the AIS meshes to the database
+#        self.add_input_file(filename=self.mesh_file, target=self.mesh_file,
+#                            database='')
+
+        self.add_namelist_file(
+            'compass.landice.tests.ismip6_run_ais', 'namelist.landice',
```
matthewhoffman, line 88: The namelist file needs to be added to the branch.

compass/landice/tests/ismip6_run_ais/projection/streams.landice
```diff
@@ -0,0 +1,132 @@
...
+<immutable_stream name="input"
+                  type="input"
+                  filename_template="{{ input_file_ais }}"
+                  input_interval="initial_only"/>
+
+<immutable_stream name="restart"
+                  type="input;output"
+                  filename_template="rst.$Y.nc"
+                  filename_interval="output_interval"
+                  output_interval="0100-00-00_00:00:00"
```
matthewhoffman, line 12: We probably want to make this somewhere between 1 and 5 years by default.

compass/landice/tests/ismip6_run_ais/run_model.py
```diff
@@ -0,0 +1,148 @@
...
+                        'cores': 400,
+                        'min_cores': 80}
+                }
+
+        res_param = res_param[self.mesh_type]
+        self.mesh_file = res_param['mesh_file']
+        self.cores = res_param['cores']
+        self.min_cores = res_param['min_cores']
+
+        self.add_input_file(filename=self.mesh_file,
```
matthewhoffman, line 79: The region mask file should also be added here.

compass/landice/tests/ismip6_run_ais/projection/streams.landice
```diff
@@ -0,0 +1,132 @@
...
+                  reference_time="1995-01-01_00:00:00"
+                  clobber_mode="truncate"
+                  precision="double"
+                  input_interval="initial_only"/>
+
+<stream name="regionsInput"
+        type="input"
+        mode="forward;analysis"
+        filename_template="regionMasks.nc"
```
matthewhoffman, line 21: This file should also be a template and dependent on the input file resolution.

compass/landice/tests/ismip6_run_ais/projection/streams.landice
```diff
@@ -0,0 +1,132 @@
...
+        mode="forward;analysis"
+        filename_template="{{ file_ais_SMB_forcing }}"
+    input_interval="0001-00-00_00:00:00"
+    record_interval="0001-00-00_00:00:00"
+                  reference_time="1950-01-01_00:00:00"
+        runtime_format="single_file">
+        <var name="sfcMassBal"/>
+</stream>
+
+<stream name="output"
```
matthewhoffman, line 71: We'll need to look through the fields required to be submitted to ISMIP6 and figure out what fields and frequency we will need.  We will likely need one file output annually for state fields and then a second file being output every timestep for flux fields that we can time average annually in post-processing.

compass/landice/tests/ismip6_run_ais/projection/streams.landice
```diff
@@ -0,0 +1,132 @@
...
+        output_interval="0000-00-00_01:00:00"
+        reference_time="2007-01-01_00:00:00"
+        packages="globalStatsAMPKG"
+        runtime_format="single_file">
+    <stream name="globalStatsOutput"/>
+    <var name="xtime"/>
+    <var name="daysSinceStart"/>
+    <var name="deltat"/>
+    <var name="allowableDtACFL"/>
+    <var name="allowableDtDCFL"/>
```
matthewhoffman, line 108: Also add the new `calvingCFLdt` variable.

compass/landice/tests/ismip6_run_ais/projection/streams.landice
```diff
@@ -0,0 +1,132 @@
...
+
+</stream>
+
+<stream name="globalStatsOutput"
+        type="output"
+        filename_template="globalStats.nc"
+        filename_interval="none"
+        clobber_mode="overwrite"
+        output_interval="0000-00-00_01:00:00"
+        reference_time="2007-01-01_00:00:00"
```
matthewhoffman, line 100: Use the same reference_time as the other streams.

matthewhoffman (changes requested):
@hollyhan , thanks for kicking this PR off.  I did a quick preliminary review and identified a handful of things that need attention.  I can perform a more careful review and testing after that.

compass/landice/tests/ismip6_run_ais/run_model.py
hollyhan: I've added the lines that are commented out for now until we create a region mask file for this mesh.

compass/landice/tests/ismip6_run_ais/projection/streams.landice
hollyhan: Done.

hollyhan: Done.

compass/landice/tests/ismip6_run_ais/projection/streams.landice
hollyhan: I've edited the `streams.landice.template` file in [https://github.com/hollyhan/compass/commit/57f85d80c8ebf3c1b6551608d071e54e41576b58](url) by going through the required ismip6 outputs [https://www.climate-cryosphere.org/wiki/index.php?title=ISMIP6-Projections2300-Antarctica#A2.3.3_Table_A1:_Variable_request_for_ISMIP6](url). There are a number of variables on the table that I need to check/confirm with you which variables on registry they correspond to, and which variables are calculated in post-processing steps rather than outputted in streams.

compass/landice/tests/ismip6_run_ais/streams.landice.template
```diff
@@ -0,0 +1,159 @@
...
+    <var name="sfcMassBalApplied"/>
+    <var name="groundedBasalMassBalApplied"/>
+    <var name="floatingBasalMassBalApplied"/>
+    <var name="dHdt"/>
+    <var name="calvingThickness"/>
+    <var name="fluxAcrossGroundingLine"/>
+    <var name="edgeMask"/>
+    <var name="cellMask"/>
+</stream>
+
```
matthewhoffman, line 119: We might also want the 2 streams we have been using in our testing on Badger to appear.  We could then comment them out if we don't want them.  Let's discuss.

matthewhoffman (changes requested):
A handful of more little things :)

compass/landice/tests/ismip6_run_ais/run_model.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Todo: confirm the number of cores needed for the high mesh
+        # We estimate that 200-1000 cells should be allocated for one core
+        res_param = {
+                'high': {'mesh_file': 'Antarctic_1to10km.nc',
+                         'mask_file': 'regionMask_AIS_1to10km.nc',
+                         'cores': 25000,
+                         'min_cores': 5000},
+                'mid': {'mesh_file': 'Antarctic_8to80km_20220407.nc',
+                        'mask_file': 'regionMask_AIS_8to10km.nc',
```
matthewhoffman, line 71: As we discussed in our meeting, let's choose a naming convention for regions masks that is  meshname.regionMask.nc

xylar: It might be worth giving the masks a date stamp, too.  We have decided to add new regions in the past in MPAS-Analysis and it became rather awkward to keep things backwards compatible without a date stamp.

compass/landice/tests/ismip6_run_ais/namelist.landice
```diff
@@ -0,0 +1,193 @@
+&velocity_solver
```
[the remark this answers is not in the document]
xylar: Also, the `&` lines and `/` lines shouldn't be included.  The parser won't understand these lines.

compass/landice/tests/ismip6_run_ais/run_model.py
```diff
@@ -0,0 +1,156 @@
...
+        mesh_type : {'high', 'mid'}
+            The resolution or mesh type of the test case
+        """
+
+        self.mesh_file = None
+        self.mesh_type = mesh_type
+
+        super().__init__(test_case=test_case, name=name)
+
+    def setup(self, velo_solver="FO"):
```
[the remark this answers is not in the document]
xylar: Also, this gets called by the framework without using any arguments:
https://github.com/MPAS-Dev/compass/blob/ff4e402533bf762d06a7b847eec290df232a44d9/compass/setup.py#L289
so any parameters you add here would always take the default value.  If you wanted to have various `velo_solver` values, this would need to happen in the constructor (`__init__()`) rather than here.

xylar:
@hollyhan, this is looking like it's in good shape.  It's some really great work!

Could you start adding the documentation when you have time?  In particular, there would need to be instructions about how the output files of the associated forcing test group can be used as inputs to this test group.  Like in the forcing test group, it would be great to have an example user config file for a specific model and scenario that you use on one of our machines.

I would like to use the example to run both the forcing tests and the projection run myself on a machine and for a model/scenario you haven't tested yet.

compass/landice/tests/ismip6_run_ais/__init__.py
```diff
@@ -0,0 +1,20 @@
+from compass.testgroup import TestGroup
+from compass.landice.tests.ismip6_run_ais.projection import Projection
+
+
+class Ismip6RunAIS(TestGroup):
```
xylar, line 5: Should ideally be:
```suggestion
class Ismip6RunAis(TestGroup):
```
(CamelCase rather than all caps).

compass/landice/tests/ismip6_run_ais/projection/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        name = 'ismip6AE'
+
+        super().__init__(test_group=test_group, name=name,
+                         subdir=name)
+
+        experiments = ['hist', 'ctrlAE', 'expAE01', 'expAE02', 'expAE03',
+                       'expAE04', 'expAE05', 'expAE06']
+        for i in range(7, 10+1):
+        #for i in range(7, 14+1):
+            experiments.append(f'expAE{i:02}')
```
xylar, line 31:
```suggestion
        experiments = ['hist', 'ctrlAE'] + [f'expAE{i:02}' for i in range(1, 11)]
```

compass/landice/tests/ismip6_run_ais/projection/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+        for exp in experiments:
+            name = f'{exp}'
+            self.add_step(
+               SetUpExperiment(test_case=self, name=name,
+                              subdir=f"{exp}", exp=exp))
```
xylar, line 37: Just use `exp` for all of these. No need for f-strings or local variables.
```suggestion
            self.add_step(
               SetUpExperiment(test_case=self, name=exp,
                               subdir=exp, exp=exp))
```

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
```diff
@@ -0,0 +1,215 @@
+import os
+from compass.model import run_model, make_graph_file
+from compass.step import Step
+import shutil, glob, sys
+from jinja2 import Template
+from importlib import resources
+
+
+class SetUpExperiment(Step):
```
xylar, line 9: It seems like this will be a step for setting up and then running an experiment eventually, not just setting it up.

compass/landice/tests/ismip6_run_ais/projection/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        super().__init__(test_group=test_group, name=name,
+                         subdir=name)
+
+        experiments = ['hist', 'ctrlAE', 'expAE01', 'expAE02', 'expAE03',
+                       'expAE04', 'expAE05', 'expAE06']
+        for i in range(7, 10+1):
+        #for i in range(7, 14+1):
+            experiments.append(f'expAE{i:02}')
+
+        for exp in experiments:
```
xylar, line 33: It seems like it might make more sense to have one test case per experiment rather than one step per experiment.  Compass is designed to make it pretty easy to set up several test cases as a "custom" suite and then run them in the order you chose but it's a little trickier to go in and edit the `steps_to_run`  config option to run only the subset of steps you want to run in a given test case.

We can work with it either way, it's just what I would think might be easier.

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
```diff
@@ -0,0 +1,215 @@
...
+#                'compass.landice.tests.ismip6_run_ais', 'namelist.sealevel.template',
+#                out_name='namelist.sealevel')
+
+#            namelist_replacements = {
+#                                    'slm_input_ice': slm_input_ice,
+#                                    'slm_input_earth': slm_input_earth,
+#                                    'slm_input_others': slm_input_others,
+#                                    'slm_output_topo': slm_output_topo,
+#                                    'slm_output_ice': slm_output_ice,
+#                                    }
```
xylar, line 92: This is still work in progress, I guess?

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
```diff
@@ -0,0 +1,215 @@
...
+                               'input_file_TF_forcing': tf_fname
+                               }
+        if self.exp == 'hist':
+            stream_replacements['input_file_init_cond'] = init_cond_fname
+            stream_replacements['input_file_region_mask'] = region_mask_fname
+            stream_replacements['input_file_melt_params'] = melt_params_fname
+        else:
+            stream_replacements['input_file_init_cond'] = 'USE_RESTART_FILE_INSTEAD'
+            stream_replacements['input_file_region_mask'] = 'USE_RESTART_FILE_INSTEAD'
+            stream_replacements['input_file_melt_params'] = 'USE_RESTART_FILE_INSTEAD'
```
xylar, line 140: This is just to make sure you don't accidentally use the initial condition?

xylar: Or is this a meaningful value that MALI knows how to work with?

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
```diff
@@ -0,0 +1,215 @@
...
+        if self.exp == 'hist' or self.exp == 'ctrlAE':
```
xylar, line 141:
```suggestion
        if self.exp in ['hist', 'ctrlAE']:
```

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
+# config options for ismip6 antarctic ice sheet data set
+[ismip6_run_ais]
+
+# Resolution that will be set up.  Should be one of 08, 04 with leading 0.  The paths below must manually be set to be consistent
+mesh_res = 08
```
xylar, line 5: Why with a leading zero?  Probably state that it is the highest resolution in km.

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
```diff
@@ -0,0 +1,215 @@
...
+        print(f"    Setting up experiment {self.exp}")
+
+        if self.exp == 'hist':
+            exp_fcg = 'ctrlAE'
+        else:
+            exp_fcg = self.exp
+
+        config = self.config
+        section = config['ismip6_run_ais']
+        mesh_res = section.get('mesh_res')
```
xylar, line 42: I would get this as an integer and add the zero padding yourself.
```suggestion
        mesh_res = section.getint('mesh_res')
```

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
```diff
@@ -0,0 +1,215 @@
...
+            'compass.landice.tests.ismip6_run_ais', 'streams.landice.template',
+            out_name='streams.landice',
+            template_replacements=stream_replacements)
+
+        # Set up namelist and customize as needed
+        self.add_namelist_file(
+            'compass.landice.tests.ismip6_run_ais', 'namelist.landice',
+            out_name='namelist.landice')
+
+        if mesh_res == '04':
```
xylar, line 156:
```suggestion
        if mesh_res == 4:
```

xylar: I think it would be better to have an `elif mesh_res == 8:` here as well and set teh corresponding options there (even if they happen to be the defaults) and then have:
````
else:
    raise ValueError(f'Unsupported mesh resolution {mesh_res}.')

File: compass/landice/tests/ismip6_run_ais/set_up_experiment.py [hunk 25]
```diff
@@ -0,0 +1,215 @@
...
+        #make_graph_file(mesh_filename=init_cond_path,
+        #                graph_filename='graph.info')
+        graphFileList = glob.glob(os.path.join(graph_files_path, 'graph.info*'))
+        for gf in graphFileList:
+            shutil.copy(gf, self.work_dir)
+
+        # provide an example submit script
+        template = Template(resources.read_text(
+            'compass.landice.tests.ismip6_run_ais',
+            f'slurm.{mesh_res}.run'))
```
xylar, line 198:
```suggestion
            f'slurm.{mesh_res:02d}.run'))
```

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
```diff
@@ -0,0 +1,215 @@
...
+        slurm_replacements = {'EXP': f'{self.exp}_{mesh_res}',
```
xylar, line 199:
```suggestion
        slurm_replacements = {'EXP': f'{self.exp}_{mesh_res:02d}',
```

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
xylar:
```suggestion
# Finest resolution of the MALI mesh that will be used in km (currently either 4 or 8). The paths below must
# be set for files on a consistent mesh
mesh_res = 8
```

compass/landice/tests/ismip6_run_ais/slurm.08.run
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+#SBATCH  --job-name={{ EXP }}
+#SBATCH  --account=m1795
+#SBATCH  --nodes=5
+#SBATCH  --tasks-per-node=64
+#SBATCH  --exclusive
+#SBATCH  --time=15:00:00
+#SBATCH  --qos=regular
+#SBATCH  --constraint=haswell
```
xylar, line 9: This template is currently designed specifically for Cori-Haswell (and for a specific account).  It would be better if we could make use of
https://github.com/MPAS-Dev/compass/tree/master/compass/job
to create a more general job script that could still meet your needs.  I suggest we try to do this in a follow-up PR that can also hopefully reduce the amount of manual work that goes into the config file.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
...
+
+# COMPASS load script path
+# This is user-specific.
+compass_load_path = /global/cscratch1/sd/hoffman2/AIS8km/compass/load_dev_compass_1.2.0-alpha.1_cori-haswell_gnu_mpt_albany.sh
```
xylar, line 9: There should be a local symlink to this config file in the test case at `../load_compass_env.sh`.  This path should not be necessary.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
...
+
+# Base path to the processed input ismip6 ocean and smb forcing files. User has to supply.
+forcing_basepath = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/forcing/ais_mesh_8to30km_res
+
+# Path to the initial condition file. User has to supply.
+init_cond_path = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/initial_conditions/AIS_8to30km_20221027/relaxation_0TGmelt_10yr_muCap/AIS_8to30km_r01_20220906.smooth3.basinsFineTuned_carvedRonne_CIRWIP_relaxation_0TGmelt_10yr_muCap.nc
+
+# Path to the file for the basal melt parametrization coefficients.
+melt_params_path = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/initial_conditions/AIS_8to30km_20221027/basin_and_coeff_gamma0_DeltaT_quadratic_non_local.nc
```
xylar, line 18: Shouldn't this be in the `forcing_basepath` and a user should only need to supply a string like the following?
```
melt_params = quadratic_non_local
```

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
...
+
+# COMPASS load script path
+# This is user-specific.
+compass_load_path = /global/cscratch1/sd/hoffman2/AIS8km/compass/load_dev_compass_1.2.0-alpha.1_cori-haswell_gnu_mpt_albany.sh
+
+# Base path to the processed input ismip6 ocean and smb forcing files. User has to supply.
+forcing_basepath = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/forcing/ais_mesh_8to30km_res
+
+# Path to the initial condition file. User has to supply.
+init_cond_path = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/initial_conditions/AIS_8to30km_20221027/relaxation_0TGmelt_10yr_muCap/AIS_8to30km_r01_20220906.smooth3.basinsFineTuned_carvedRonne_CIRWIP_relaxation_0TGmelt_10yr_muCap.nc
```
xylar, line 15: Can we eventually put this in the database for automatic download?

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
...
+
+# Path to the file for the basal melt parametrization coefficients.
+melt_params_path = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/initial_conditions/AIS_8to30km_20221027/basin_and_coeff_gamma0_DeltaT_quadratic_non_local.nc
+
+# Path to the region mask file
+region_mask_path = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/initial_conditions/AIS_8to30km_20221027/AIS_8to30km_r01_20220607.regionMask_ismip6.nc
```
xylar, line 21: What are these regions and how should be file be created?  If this is part of the forcing, can we populate this automatically?  If it's not part of the forcing yet, can we add it?

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
...
+
+# Path to the directory containing graph files
+graph_files_path = /global/project/projectdirs/piscees/MALI_projects/ISMIP6-2300/initial_conditions/AIS_8to30km_20221027
```
xylar, line 24: Can these be generated by this test case rather than being supplied?

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
...
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
+
+# NOTE: for the directories related to the sea-level model input/outputs, the slash '/' at the end of the directory name is necessary. 
+# Path to the directory containing globally defined ice thickness field for the sea-level model
```
xylar, line 30: It would be much better to append the `/` at the end in the code if it is missing than tell the user they have to remember to add it.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
```diff
@@ -0,0 +1,44 @@
...
+slm_input_earth = /global/project/projectdirs/piscees/MALI_projects/SeaLevelModel_Inputs/earthmodel/
+
+# Path to the diectory containing other input files (present-day global topography and sea-level model grid files)
+slm_input_others = /global/project/projectdirs/piscees/MALI_projects/SeaLevelModel_Inputs/others/
+
+# mapping file name, from MALI mesh to SLM grid
+mapfile_mali_to_slm = mapfile_from_AIS_8to30km_r01_to_GL512_bilinear.nc
+
+# mapping file name, from SLM grid to MALI mesh
+mapfile_slm_to_mali = mapfile_from_GL512_to_AIS_8to30km_r01_bilinear.nc
```
xylar, line 43: Wouldn't it be better if these are determined automatically in the code, based on the mesh resolution?

xylar (commented):
@hollyhan, this test group represents a lot of excellent work!  I'm looking forward to discussing it with you further today and in the coming week(s).  Here are several comments to get things started.

matthewhoffman:
@xylar , thanks for looking at this.  I more or less hijacked this PR from Holly a few weeks ago, but I haven't updated the PR description yet.  This was designed as a special usage of COMPASS to generate reproducible run directories for the ISMIP6 experiments, and not meant as a 'real' COMPASS test case.  We're finishing a few final adjustments for the last few ISMIP6 experiments.  Once that is done I will update the description with more context and explanation.

xylar:
@matthewhoffman, okay, that's fine.  I probably should have been taken off as a reviewer but no big deal.

I do think we want the equivalent of this for purposes of extending it to include the topography extrapolation as well.  So I would like to talk about how we disentangle this one-off version from what does go into compass as a test case.

matthewhoffman:
@xylar , you are right - I should have updated the labels and removed you as a reviewer until things settled down, sorry about that.  Holly mentioned the plan to meet next week and discuss the strategy for finalizing this PR and extending it for the topography extrapolation application, which I think is a good idea.  I think there a number of paths forward of varying levels of tradeoff between complexity and effort, and the best choice will depend on our expectations for future usage scenarios.

xylar:
@matthewhoffman, yes a meeting next week would be helpful.  My original thinking with this was that it would be the framework for an additional set set of test cases that included the thermal forcing extrapolation.  As such, it needs to be able to run MALI from compass itself rather than just set up a script and it needs to handle the namelist changes associated with initial vs. restart runs (i.e. run for one interval from an initial condition, then change namelist options to continue running from a restart).  This needs to be handled from compass so that compass can also coordinate the processes for doing the TF extrapolation at then end of an appropriate interval (say 1 or 5 years).

If the current test case is now for a different purpose, we need to have a different test case that serves the purpose of being a stepping stone to the runs with dynamic TF extrapolation.

compass/landice/tests/ismip6_run_ais/projection/__init__.py
matthewhoffman: Restructured in d5e586be

compass/landice/tests/ismip6_run_ais/projection/__init__.py
matthewhoffman: Cleaned up in d5e586be

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
matthewhoffman: The intent was to only use compass for setting up the experiments, but to have users 'manually' submit/run experiments.   The experiments are expensive and complicated enough to only want expert users to be running them.

compass/landice/tests/ismip6_run_ais/projection/__init__.py
matthewhoffman: There are pros and cons to each layout, but after discussion with you and with Holly, my preference is to keep each experiment as a step.  One advantage to that is not have a bunch of experiments each with a single step in a subdirectory, but instead have each a test case with all the experiments as subdirectories.  I think after making the experiments to set up configurable in the cfg file, that provides the desired user flexibility.

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
matthewhoffman: This is to make sure the initial condition is not used accidentally.  It's meant both as a way to force MALI to abort if that were somehow the case and to let a user know where the input should be coming from if the look in the streams file.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
matthewhoffman: Because the ISMIP6 submission categorizes mesh resolution as one of 16, 08, 04, 02, 01, we've adopted that naming convention to refer to our meshes.  When the results get uploaded they require a directory name with the resolution including the leading zero, so I've adopted it everywhere here.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
matthewhoffman: In our tuning, we ended up recalibrating deltaT manually and not using the default files created by the forcing.  We want the flexibility to do that, though in some cases one might choose to use the default ocean params file.

compass/landice/tests/ismip6_run_ais/set_up_experiment.py
matthewhoffman: Change made.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
matthewhoffman: I've gone ahead and taken your suggestion to have it be an integer in the cfg file and padded with zeros as needed.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
matthewhoffman: Ah, good point.  I've removed that.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
matthewhoffman: Yes, we could/should place the final input files in the database, although I don't think it will be practical to put the forcing files there.  Perhaps we could put the forcing for the hist/ctrl experiments there but leave out the projection forcings.

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
matthewhoffman: Good question.  I had two issues when I tried to do that:
1. `make_graph_file` seemed to not do anything when called during `configure`.  Because I am not intending for `compass run` to be used to actually run the model, it might not be possible to use `make_graph_file`
2. Because we are intending for a user to use the slurm batch script outside of `compass run`, we also need the `graph.info.part.*` file for the number of processors used.  These are meant to be pre-generated for the desired number of processors the experiments should be run with.  

So we either need flexibility to use the graph file generation tools in a nonstandard way, or I need to be convinced that using compass run to run the experiments is a good idea.  (Or we leave some of the nonstandard steps to users to supply themselves, given the atypical nature of this test case.)

compass/landice/tests/ismip6_run_ais/ismip6_run_ais.cfg
matthewhoffman: These are the ISMIP6 basin regions that are used in the forcing but generated as a regionMask.  It seemed overkill to me to have this test case regenerate those, when we've already made them for the mesh when we tested the initial condition, etc.  We could move these files to the database with the input file when we do that, perhaps.

matthewhoffman:
# Testing
I set up the tier1 experiments on perlmutter using the default config options in the PR.  All experiments were generated without error.  I ran the `hist` experiment by cd'ing to its directory and submitting the batch script.  The experiment ran successfully.  Due to the large resource requirements, I will not test running the entire ISMIP6 ensemble again!  But if any minor issues come up in future attempts, they can be cleaned up then.

matthewhoffman:
Docs can be reviewed here:
https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-ismip6_run/index.html
Specifically:
* User's Guide description of test group: https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-ismip6_run/users_guide/landice/test_groups/ismip6_run.html
* Developer's Guide description of test group:https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-ismip6_run/developers_guide/landice/test_groups/ismip6_run.html
* API: https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-ismip6_run/developers_guide/landice/api.html#ismip6-run

matthewhoffman:
@hollyhan & @trhille , I have finally completing updating this ISMIP6 run PR and it is ready for final review.  I reorganized things substantially to make things a little cleaner and to leave room for adding other ISMIP6 standardized experiments to the test group in the future.  I also adopted a similar design to what had evolved in the ensemble_generator test, with the main thing being individual runs are steps of a single test, but those steps are not meany to be run by compass run at the test case level directly, but instead should be submitted as individual jobs.

@hollyhan , this is a bit awkward in that you opened the PR but I implemented most of what's in the final version and now want your review.  I don't think it's possible for you to review your own PR.  You should be able to make inline comments.  But I can also close this and open a fresh PR if that's easier - let me know.

In terms of testing and review, I think simply being able to set up a set of experiments and have one or two run for a short time would be sufficient.

@xylar , I don't want to ask your time to re-review this whole thing, but a lot of what I have implemented most recently is in response to your previous review.  I would like you to review this one compass framework commit: 77a61cfae4dcea9387b4ee1cc30fe90af7a975e2

xylar:
@matthewhoffman, 77a61cf looks great to me!

xylar (approved):
I'm approving the framework changes.  I haven't had time to review the rest and may not in the timeframe you want to merge.

trhille:
I successfully set up the tier 1 ensemble using the default .cfg options and ran the `hist` member on Perlmutter.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -0,0 +1,254 @@
...
+
+        if useCalvingMask:
+            options = {'config_calving': "'none'",
+                       'config_apply_calving_mask': ".true."}
+            self.add_namelist_options(options=options,
+                                      out_name='namelist.landice')
+
+        # For all projection runs, symlink the restart file for the
+        # historical run
+        # don't symlink restart_timestamp or you'll have a mighty mess
```
trhille, line 206: :)

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -0,0 +1,254 @@
...
+
+        if use_vM_calving:
+            vM_param_path = section.get('von_mises_parameter_path')
+            vM_stream_replacements = {'input_file_VM_params': vM_param_path}
+            self.add_streams_file(
+                resource_location, 'streams.vM_params',
+                out_name='streams.landice',
+                template_replacements=vM_stream_replacements)
+
+        # Set up namelist and customize as needed
```
trhille, line 170: I don't know if you want to deal with this here, but I used these for the 4km runs:
```
config_front_mass_bal_grounded = 'ismip6'
config_use_3d_thermal_forcing_for_face_melt = .true.
```

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/streams.landice.template
```diff
@@ -0,0 +1,159 @@
...
+    <var name="betaSolve"/>
+    <var name="daysSinceStart"/>
+    <var name="cellProcID"/>
+    <var name="cellMask"/>
+    <var name="deltat"/>
+    <var name="vonMisesStress"/>
+    <var name="damage"/>
+</stream>
+
+<stream name="output_flux"
```
trhille, line 95: Add `faceMeltingThickness` and `faceMeltSpeed` here for 4km?

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/namelist.landice
```diff
@@ -0,0 +1,78 @@
...
+    config_start_time ='file'
+    config_stop_time = '2300-01-01_00:00:00'
+    config_calendar_type = 'noleap'
+
+    config_stats_interval = 0
+    config_write_stats_on_startup = .false.
+    config_stats_cell_ID = 1
+    config_write_output_on_startup = .true.
+    config_pio_num_iotasks = 10
+    config_pio_stride = 32
```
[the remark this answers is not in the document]
trhille: I see that it's replaced for the 4km configuration, but I don't think it is changed for 8km.

trhille (changes requested):
@matthewhoffman, thanks for completing this extremely useful tool! I have a few comments that you can mostly take or leave, but I'll wait to approve until you respond to those points.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -0,0 +1,254 @@
...
+        # Copy files we'll need from local paths specified in cfg file
+        if self.exp == 'hist':
+            shutil.copy(init_cond_path, self.work_dir)
+        shutil.copy(melt_params_path, self.work_dir)
+        shutil.copy(region_mask_path, self.work_dir)
+
+        # Find and copy correct forcing files
+        smb_search_path = os.path.join(
+            forcing_basepath, exp_fcg,
+            'processed_SMB_*_smbNeg_over_bareland.nc')
```
hollyhan, line 89: Thanks for taking over the PR and making it look/work fantastic, @matthewhoffman!  During the final review step of the `ismip6_forcing` test group, a change has been made that this forcing file name is not relevant anymore. That is, the `processed_` part was replaced with `mali mesh name`, e.g. AIS_4to20km_r01. The (https://github.com/hollyhan/compass/blob/867bab78c3b8af109069c571acc9c87a02885fe3/compass/landice/tests/ismip6_forcing/atmosphere/process_smb_racmo.py#L53-L54).

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -0,0 +1,254 @@
...
+        if len(fcgFileList) == 1:
+            smb_path = fcgFileList[0]
+            smb_fname = os.path.split(smb_path)[-1]
+            shutil.copy(smb_path, self.work_dir)
+        else:
+            sys.exit("ERROR: Did not find exactly one matching SMB file at "
+                     f"{smb_search_path}: {fcgFileList}")
+
+        tf_search_path = os.path.join(forcing_basepath, exp_fcg,
+                                      'processed_TF_*.nc')
```
hollyhan, line 100: The above comment is relevant for here as well.

hollyhan:
@matthewhoffman - thanks for taking over the PR and making it look/work fantastic!
I'm testing the PR on Chicoma, and I get the following error:
```
(dev_compass_1.2.0-alpha.4) hollyhan@ch-fe1:/usr/projects/climate/hollyhan/compass/compass/landice/tests/ismip6_run/ismip6_ais_proj2300> compass setup -n 87 -p MALI-Dev/components/mpas-albany-landice/ -w /lustre/scratch4/turquoise/hollyhan/run_ismip6_projections 
Setting up test cases:
  landice/ismip6_run/ismip6_ais_proj2300
    Setting up experiment expAE05
  File "/usr/projects/climate/hollyhan/mambaforge/envs/dev_compass_1.2.0-alpha.4/bin/compass", line 33, in <module>
  File "/usr/projects/climate/hollyhan/compass/compass/__main__.py", line 63, in main
    commands[args.command]()
  File "/usr/projects/climate/hollyhan/compass/compass/setup.py", line 386, in main
    setup_cases(tests=tests, numbers=args.case_num,
  File "/usr/projects/climate/hollyhan/compass/compass/setup.py", line 149, in setup_cases
    setup_case(path, test_case, config_file, machine, work_dir,
  File "/usr/projects/climate/hollyhan/compass/compass/setup.py", line 302, in setup_case
    step.process_inputs_and_outputs()
  File "/usr/projects/climate/hollyhan/compass/compass/step.py", line 784, in process_inputs_and_outputs
    self._generate_namelists()
  File "/usr/projects/climate/hollyhan/compass/compass/step.py", line 822, in _generate_namelists
    namelist = compass.namelist.ingest(defaults_filename)
  File "/usr/projects/climate/hollyhan/compass/compass/namelist.py", line 34, in ingest
    with open(defaults_filename, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/projects/climate/hollyhan/compass/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/MALI-Dev/components/mpas-albany-landice/default_inputs/namelist.landice'
```  
It looks like the path `MALI-Dev/components/mpas-albany-landice/default_inputs/` is being joined with the `resource_location` defined here (https://github.com/MPAS-Dev/compass/pull/397/files#diff-624f5cab0fa7b3d4440e9baac406942c355c12feeaedb043a18bc1246ca10ccaR59-R60).

matthewhoffman:
@hollyhan , thanks for trying this out and flagging this issue.  I'm not sure what the cause would be.  Can you confirm that you compiled MALI within the COMPASS submodule or specified a different compile location using the compass setup `-p` option?  If so, can you send me your cfg file used to set things up?

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
matthewhoffman: I created an issue here: https://github.com/MPAS-Dev/compass/issues/589

docs/developers_guide/landice/test_groups/ismip6_run.rst
```diff
@@ -0,0 +1,58 @@
...
+There is no shared functionality for the ``ismip6_run`` test group.
+Shared functions may be added if additional test cases are added and the
+needed functionality can be generalized.
+
+ismip6_ais_proj2300
+-------------------
+
+The :py:class:`compass.landice.tests.ismip6_run.ismip6_ais_proj2300.Ismip6AisProj2300`
+sets up an ensemble of ISMIP6 Antarctica 2300
+simulations.  The constructor does nothing other than
```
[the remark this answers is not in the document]
matthewhoffman: yes, added

docs/users_guide/landice/test_groups/ismip6_run.rst
```diff
@@ -0,0 +1,118 @@
...
+    qos = regular
+
+    [job]
+    wall_time = 10:00:00 
+
+Steps for setting up and running experiments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+1. With a compass conda environment set up, run, e.g.,
+   ``compass setup -t landice/ismip6_run/ismip6_ais_proj2300 -w WORK_DIR_PATH -f USER.cfg``
```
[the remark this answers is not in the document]
matthewhoffman: I added some more explanation about those options.

docs/users_guide/landice/test_groups/ismip6_run.rst
```diff
@@ -0,0 +1,118 @@
+.. _landice_ismip6_run:
+
+ismip6_run
+==========
+
+The ``landice/ismip6_run`` test group includes a test case for setting up
+one or more experiments from the `ISMIP6 set of experiments <https://www.climate-cryosphere.org/wiki/index.php?title=ISMIP6_wiki_page>`_.
+This functionality assumes the initial condition, parameter, and forcing files
+have already been generated.  It creates a consistent set of run directories
```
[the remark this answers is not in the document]
matthewhoffman: added

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+        name = 'ismip6_ais_proj2300'
+
+        super().__init__(test_group=test_group, name=name,
+                         subdir=name)
+
+    def configure(self):
+        """
+        Set up the desired ISMIP6 AIS 2300 experiments.
+
+        Read the list from the config of which experiemnts the
```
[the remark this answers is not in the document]
matthewhoffman: added

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
matthewhoffman: I created a new cfg option for including facemelting or not.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/streams.landice.template
matthewhoffman: These now get included when facemelting is activated in the cfg.

matthewhoffman:
I've updated this PR based on the reviews.  @trhille , see if the changes adequately address the items you brought up.  @hollyhan , I confirmed that I can now set up and run the 4km configuration on Chicoma using a modified version of the cfg you sent me.  

The full cfg I used is:
```
[ismip6_run_ais_2300]
# list of experiments to set up.
# Can be "tier1", "tier2", "all", or a comma-delimited list of runs
# "tier1" is expAE01-expAE06 plus hist and ctrlAE
# "tier2" is expAE07-expAE14
exp_list = hist,ctrlAE,expAE05

# Resolution that will be set up, in km.  Should be one of 4, 8, as those are
# the two resolutions currently supported.
# mesh_res is informational only and is used for directory naming conventions
# The filepaths below must manually be set to be consistent.
mesh_res = 4

# number of tasks to use for each run
ntasks = 512

# Base path to the processed input ismip6 ocean and smb forcing files.
# User has to supply.
forcing_basepath = /usr/projects/climate/hollyhan/ismip6_files/forcing/4to20km_res

# Path to the initial condition file. User has to supply.
init_cond_path = /usr/projects/climate/hollyhan/ismip6_files/initial_conditions/AIS_4to20km_20230105/AIS_4to20km_r01_20220907_m5_drop_bed_20m_bulldoze_troughs_75_to_400m_Enderby_maxstiffness_0.8_TG_pinning_40maf_bedmap2_surface_ASE_05perc_seafloor_mu.nc

# Path to the file for the basal melt parametrization coefficients.
melt_params_path = /usr/projects/climate/hollyhan/ismip6_files/initial_conditions/AIS_4to20km_20230105/basin_and_coeff_gamma0_DeltaT_quadratic_non_local_after_relaxation.nc

# Path to the region mask file
region_mask_path = /usr/projects/climate/hollyhan/ismip6_files/initial_conditions/AIS_4to20km_20230105/AIS_4to20km_r01_20220907.regionMask_ismip6.nc

# Calving method to use.  Should be one of:
# 'restore' for restore-calving front method (fixed calving front)
# 'von_mises' for von Mises threshold stress calving law
calving_method = restore

# Path to the file containing the von Mises parameter fields
# groundedVonMisesThresholdStress and floatingVonMisesThresholdStress.
# Only required if calving_method is set to 'von_mises'
von_mises_parameter_path = UNKNOWN

use_face_melting = True
```

Note my runs on Chicoma were dying in the velocity solver with this message in a few of the `log.albany.*.out` files:
```
libfabric:125500:1680736313:cxi:core:cxip_ux_onload_cb():2259<warn> nid002374: RXC (0x3882:123:0): PtlTE 231LE resources not recovered during flow control. FI_CXI_RX_MATCH_MODE=[hybrid|software] is required.
```
I was able to make it go away by setting this environment variable (e.g. in the submit script):
```
export FI_CXI_RX_MATCH_MODE=hybrid
```
(See info about this option in the NERSC docs here: https://docs.nersc.gov/performance/network/)
Longer term, we may need a more robust option for dealing with this option on Chicoma, but that's outside the scope of this PR.  For now, if one encounters this error, setting the env variable manually works to resolve it.

trhille (approved):
Thanks, @matthewhoffman!

xylar:
@hollyhan, could you rebase onto the latest `main`, where I have a fix that should allow the automated tests to pass?

matthewhoffman (approved):
Approving to remove my review as a requirement, as all issues have been addresssed.

PR #400, opened 2022-05-28
Remove extra check_call

trhille:
Remove extra check_call that leads to interpolation or cell culling
being called twice.


xylar (approved):
Yep, a clear bug.

PR #401, opened 2022-05-28
Optionally run steps as subprocesses

xylar:
This will enable calls to external packages that may, in turn, call subprocess.  If this functionality is called directly, there is no obvious way to capture stdout/stderr in a log file.  One example of this is jigsawpy's functionality for creating icosahedral meshes.

### DIFF
--- a/docs/developers_guide/organization.rst
+++ b/docs/developers_guide/organization.rst
@@ -860,6 +860,20 @@ Some attributes are available after calling the base class' constructor
     subprocess if there is not a good way to redirect output to a log
     file (e.g. if the step calls external code that, in turn, calls
     additional subprocesses).
+    
+    The default behavior when python code calls one of the ``subprocess``
+    functions is that the output goes to ``stdout``/``stderr`` 
+    (i.e. the terminal).  When python code outside of compass 
+    (e.g. ``jigsawpy``) calls a ``subprocess`` function (e.g. to call
+    JIGSAW), that output goes to the terminal rather than a log file.  
+    For most output to ``stdout``/``stderr`` like ``print()`` statements,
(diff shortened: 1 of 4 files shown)


xylar:
@matthewhoffman, all I'm looking for in a review here is:
* making sure you're okay with this change
* a quick MALI test to make sure this does no harm.

It would be great to get this in soonish because I want to build on it (draft PR coming later this weekend).

xylar:
## Testing

I have tested this using my https://github.com/xylar/compass/tree/separate_mesh_steps branch.  Previously (without using `subprocess`) when I created icosahedral meshes, JIGSAW output was going to the terminal.  With this fix, it goes to the log file as desired.

docs/developers_guide/organization.rst
```diff
@@ -854,6 +854,13 @@ Some attributes are available after calling the base class' constructor
     Whether to get all of the outputs for the step from the database of
     cached outputs for the MPAS core that this step belongs to
 
+``self.run_as_subprocess``
```
matthewhoffman, line 857: @xylar , can you elaborate on how running as a subprocess or not changes the logging behavior?  Is subprocess stdout lost, sent to the terminal/job log, or something else?

xylar: @matthewhoffman Do you want me to elaborate on that here or in the documentation?

matthewhoffman (approved):
I ran our integration suite with this branch and observed no ill effects.  I'm approving and leaving to your judgment whether and how to adjust the docs per the thread on that.

xylar:
Thanks, @matthewhoffman!  I did indeed add quite a bit more detail to the docs.

PR #403, opened 2022-06-01
Remove redundant default argument values

altheaden:
Default value for default is already None


xylar:
## Testing

I ran the `nightly` suite with these fixes and everything ran fine.

xylar:
@altheaden, thanks for this clean up!

PR #404, opened 2022-06-02
Constrain mamba to avoid recent issue

xylar:
See https://github.com/mamba-org/mamba/issues/1708

### DIFF
--- a/conda/shared.py
+++ b/conda/shared.py
@@ -171,7 +171,7 @@ def install_miniconda(conda_base, activate_base, logger):
     commands = '{}; ' \
                'conda config --add channels conda-forge; ' \
                'conda config --set channel_priority strict; ' \
-               'conda install -y mamba boa; ' \
+               'conda install -y "mamba>=0.24.0" boa; ' \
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I needed to delete my `~/miniconda3` but once I did that I was able to use `./conda/configure_compass_env.py` as usual to reinstall Miniconda3 just fine with this new constraint and then create the conda environment for compass.

So far, I don't have a great alternative for folks to try if they want to preserve their existing `miniconda3` environments.

xylar:
@matthewhoffman, can you let me know if this works for you?  It may hang when it tries to install `mamba >=0.24.0`.  If so, I don't currently know of an alternative to deleting your `miniconda3` and starting over.  I would normally use `mamba` to investigate the problem but `mamba` *is* the problem.

matthewhoffman:
@xylar , this works for me, and I didn't even need to delete my miniconda install first. Thanks for dealing with this.

xylar:
Thanks @matthewhoffman!  Fingers crossed that this problem goes away permanently in the not-too-distant future.

PR #405, opened 2022-06-02
Remove all occurrences of "max_disk"

altheaden:
Values were intended for task parallelism but are no longer relevant.


xylar:
# Testing

I ran the `nightly` test suite with these changes and everything ran fine.  I didn't test against a baseline, since these changes are not very likely to affect the results.  If anything had gone wrong, it would have showed up in the form of failing test executaion.

xylar:
Thanks @altheaden!  Always nice to clean up things we don't need.

PR #408, opened 2022-06-03
Add data interpolation to Antarctica mesh generation case

trhille:
Interpolate data from gridded datasets to Antarctic mesh within COMPASS. This takes care of the peculiarities of the current gridded compilation dataset (antarctica_8km_2020_10_20.nc), as well as using conservative remapping directly from the high-resolution BedMachineAntarctica and MeASUReS velocity datasets. There is a fairly heavy degree of pre-processing done to get the BedMachine and MeASUReS datasets ready to be used here. The pre-processing includes renaming variables, setting reasonable _FillValue and missing_value attributes, extrapolating fields to avoid interpolation ramps at ice margins, updating mask values, and raising the bed topography at Lake Vostok to ensure a flat ice surface. Those data files and processing scripts currently live here on Badger: `/usr/projects/climate/trhille/data`. Eventually that pre-processing could be integrated into a new step in COMPASS, or the processed data files could be added to the server on Anvil and downloaded as needed.

The conservative remapping step using ESMF_RegridWeightGen requires multiple nodes in order to not run out of memory. This is a major short-coming of the current set up for this test-case, which runs the entire case on five Badger nodes, even though four of those nodes are only used for a small fraction of the total run time. This should probably be updated to be a separate step in the future in order to save on cost.

This currently requires use of a branch of MPAS-Tools to interpolate the `iceMask` variable: https://github.com/trhille/MPAS-Tools/tree/add_iceMask_to_interp_script. The steps to use this are:
        1. Create your compass environment as normal and source the load script
        2. Go to your MPAS-Tools directory and checkout https://github.com/trhille/MPAS-Tools/tree/add_iceMask_to_interp_script
        3. Still from MPAS-Tools directory  python -m pip install -e conda_package
        4. Proceed setting up test case as normal

This can take a long time to run because the conservative remapping is slow. I recommend at least a 4 hour job on 5 nodes for medium resolution (~6–30km).

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -223,7 +223,7 @@ def set_cell_width(self, section, thk, vx=None, vy=None,
     return cell_width
 
 
-def get_dist_to_edge_and_GL(self, thk, topg, x, y, window_size=1.e5):
+def get_dist_to_edge_and_GL(self, thk, topg, x, y, section, window_size=None):
     """
     Calculate distance from each point to ice edge and grounding line,
     to be used in mesh density functions in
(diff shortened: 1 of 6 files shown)


matthewhoffman:
Replaced by #750

PR #414, opened 2022-06-24
New Mesh: Add ARRM10to60E2r1 mesh

xylar:
The mesh resolution has been ported from legacy COMPASS.


compass/ocean/tests/global_ocean/mesh/arm10to60/arm10to60.cfg
```diff
@@ -0,0 +1,57 @@
+# Options related to the vertical grid
+[vertical_grid]
+
+# the type of vertical grid
+grid_type = tanh_dz
+
+# Number of vertical levels
+vert_levels = 64
```
xylar, line 8: @milenaveneziani and @darincomeau, the vertical grid for this mesh in legacy COMPASS was the 64-level version.  If you plan to use the 60-level version instead, please let me know.  We would replace this whole config section with:
```ini
# Options related to the vertical grid
[vertical_grid]

# the type of vertical grid
grid_type = 60layerPHC
```

compass/ocean/tests/global_ocean/mesh/arm10to60/arm10to60.cfg
```diff
@@ -0,0 +1,57 @@
...
+# minimum of cores, below which the step fails
+init_min_cores = 8
+# maximum memory usage allowed (in MB)
+init_max_memory = 1000
+
+## config options related to the forward steps
+# number of cores to use
+forward_cores = 720
+# minimum of cores, below which the step fails
+forward_min_cores = 144
```
xylar, line 35: These numbers were copied from the WC14 mesh and need to be updated for ARM10to60.  We want to have about 300 cells per core as the target (`forward_cores`) and 3000 cells per core for the minimum (`forward_min_cores`).

compass/ocean/tests/global_ocean/mesh/arm10to60/arm10to60.cfg
```diff
@@ -0,0 +1,57 @@
...
+# maximum memory usage allowed (in MB)
+forward_max_memory = 1000
+
+## metadata related to the mesh
+# the prefix (e.g. QU, EC, WC, SO)
+prefix = ARM
+# a description of the mesh and initial condition
+mesh_description = MPAS Arctic Regional Mesh (ARM) for E3SM version
+                   ${e3sm_version}, with ${min_res}-km resolution in the Arctic
+                    and ${levels} vertical levels
```
xylar, line 45: @darincomeau and @milenaveneziani, please suggest updated text for this description.  It goes into the metadata for every NetCDF file produced for this mesh in compass.

compass/ocean/tests/global_ocean/mesh/arm10to60/dynamic_adjustment/__init__.py
```diff
@@ -0,0 +1,240 @@
+from compass.ocean.tests.global_ocean.dynamic_adjustment import \
+    DynamicAdjustment
+from compass.ocean.tests.global_ocean.forward import ForwardStep
+
+
+class ARM10to60DynamicAdjustment(DynamicAdjustment):
```
xylar, line 6: This hope class was copied from the WC14 version and will need to be updated with an appropriate number of steps and appropriate parameter values.

compass/ocean/tests/global_ocean/mesh/arm10to60/namelist.split_explicit
```diff
@@ -0,0 +1,9 @@
+config_time_integrator = 'split_explicit'
+config_dt = '00:12:00'
+config_btr_dt = '00:00:30'
```
xylar, line 3: @milenaveneziani and @darincomeau, these time steps need to be adjusted.  They are from the WC14.

compass/ocean/tests/global_ocean/mesh/arm10to60/namelist.split_explicit
```diff
@@ -0,0 +1,9 @@
...
+config_run_duration = '0000_01:00:00'
+config_write_output_on_startup = false.
+config_mom_del2 = 400
+config_mom_del4 = 7.7e9
```
xylar, line 7: @darincomeau and @milenaveneziani, same with these.  They need to be updated from the WC14 values.

xylar:
@darincomeau and @milenaveneziani, if you can please figure out the missing pieces while I'm on vacation next week, I can finish this PR up when I get back to work on July 4th (not a holiday for me).

I'm still working on building the mesh but it should be done soon(ish).  Here is the resolution map and construction history:

cellWidthGlobal

mesh_construction

I don't plan to try to create the initial condition run any of the other tests cases until we resolve the questions above.

xylar:
It looks as if critical passages are off by default: https://github.com/MPAS-Dev/compass/blob/master/compass/ocean/tests/global_ocean/mesh/cull.py#L15
We need to turn them on, so that's something I will add to this PR.  I think we haven't noticed this default because we haven't used the new compass to produce any official E3SM meshes yet (just for testing).

xylar:
Here's culled mesh in Paraview:
arm10to60_res

compass/ocean/tests/global_ocean/mesh/arm10to60/namelist.split_explicit
```diff
@@ -0,0 +1,9 @@
+config_time_integrator = 'split_explicit'
+config_dt = '00:12:00'
```
darincomeau, line 2: `config_dt = '00:10:00'`

compass/ocean/tests/global_ocean/mesh/arm10to60/namelist.split_explicit
darincomeau, line 3: @milenaveneziani looking at our existing cases I see `config_btr_dt = '0000_00:00:24'`, is that a mistake?

compass/ocean/tests/global_ocean/mesh/arm10to60/namelist.split_explicit
darincomeau: `config_mom_del2 = 10.0`, `config_mom_del4 = 1.5e10`

compass/ocean/tests/global_ocean/mesh/arm10to60/namelist.split_explicit
darincomeau: I commented on the individual lines but it's probably easier to keep it in this comment: `config_dt = '00:10:00'`, and we currently have `config_btr_dt = '0000_00:00:24'`, which may be a mistake?

darincomeau:
@xylar *many* thanks for getting this started! @milenaveneziani  and I are also on vacation next week, but I was able to address some of the comments. Would you prefer I push to this branch, or leave it to you based on the comments?

xylar:
> @milenaveneziani and I are also on vacation next week

Ah, okay.  Then just keep the comments coming today and we'll pick up in a week.

> Would you prefer I push to this branch, or leave it to you based on the comments?

I think having you comment here and me change things is best for now.  I imagine the dynamics adjustment will take a bit of trial and error.  Probably it would make sense for whoever does that work to then push the final result to the branch at the end.

compass/ocean/tests/global_ocean/mesh/arm10to60/dynamic_adjustment/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+        step = ForwardStep(test_case=self, mesh=mesh, init=init,
+                           time_integrator=time_integrator, name=step_name,
+                           subdir=step_name)
+
+        namelist_options = {
+            'config_run_duration': "'00-00-00_06:00:00'",
+            'config_dt': "'00:00:30'",
+            'config_btr_dt': "'00:00:01.5'",
+            'config_Rayleigh_friction': '.true.',
+            'config_Rayleigh_damping_coeff': '1.0e-3'}
```
xylar, line 70: For each step (except `simulation`), these are the values you'll want to adjust.  We could try the same as the WC14 to begin with but likely the time steps will need to be reduced because of the difference in minimum resolution.

Run each `damped_adjustment` step in compass on its own (rather than trying to run this as a suite).  Monitor the global CFL number and the kinetic energy.  If the CFL gets much over 0.1, you're in trouble.  If the KE spikes, you're in trouble.  Some combination of a shorter time step and more Rayleigh damping typically helps.

compass/ocean/tests/global_ocean/mesh/arm10to60/arm10to60.cfg
milenaveneziani: @xylar : the old ARRM mesh had 80 levels, and I think we would like to keep it that way.

compass/ocean/tests/global_ocean/mesh/arm10to60/arm10to60.cfg
milenaveneziani: We have been asked to move away from the `ARM` acronym because it is confused with the ARM observational program. Maybe we could use ARRM as before (Arctic Regional Refined Mesh).

xylar:
@milenaveneziani, I switched from ARM to ARRM and to the 80-layer vertical grid, as you requested.  I'm in the car so don't have time or battery life to try this out yet.  Maybe give it a try yourself if you have time.

milenaveneziani:
Thanks @xylar. Before we generate the initial condition, I would like to check the mesh with paraview. Do you have a vtk file at nersc or other computer?
Also, do you know if the land-locked cells in the polar regions are handled by default in compass? That's another thing that needs to be checked.

Of course, we can wait until next week for all this.

compass/ocean/tests/global_ocean/mesh/arm10to60/namelist.split_explicit
milenaveneziani: @darincomeau: as I mentioned above, I don't think this is a mistake. I could double check with Gennaro, who ran the first ARRM60to10 simulation, to see if he remembers something about it.

xylar:
Hmm, I should have posted a path yesterday.  It is:
```
/lcrc/group/e3sm/ac.xylar/compass_1.1/anvil/test_20220627/arrm/ocean/global_ocean/ARRM10to60/
```
Things got stuck in `init` until I upped the number of cores so I need to update that.

milenaveneziani:
Thanks @xylar. A couple of us have been checking the mesh, especially in the Arctic. I noticed that all the key passages are open, but there are several potentially problematic areas with one-cell channels or land-locked small regions where sea-ice can either build up or be channeled at high-velocities (see screen shots below). How did we handle these situations in the past? And is there a way to instruct compass to add a land or an ocean cell at will?
Pinging @mark-petersen as well, as I know he has had to do this in the past.

Screen Shot 2022-06-29 at 4 45 31 AM
Screen Shot 2022-06-29 at 4 53 27 AM
Screen Shot 2022-06-29 at 5 24 50 AM

xylar:
@milenaveneziani, thanks for taking a careful look.  

The process for widening channels always runs for all cells poleward of +/-43 degrees latitude:
https://github.com/MPAS-Dev/compass/blob/e3056b8420ab6f638034b1bc52ff0e9f956b2a76/compass/ocean/tests/global_ocean/mesh/cull.py#L147-L150
If it is missing cells, I can only guess that there is something wrong with the algorithm.  This would not apply to Gibraltar (your last image) but I don't undersand how we can be missing so many 1-cell channels in your other images.  We should write out the land mask out right before and after the call to `add_land_locked_cells_to_mask()` and see if these areas are missing or if they are somehow added by critical passages later.  I don't currently see how the latter can be true.

I will try adding lines to write out these masks and rerun right now.

xylar:
I'm rerunning here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.1/anvil/test_20220629/arrm/ocean/global_ocean/ARRM10to60/mesh
```
There should be a new file `land_mask_with_land_locked_cells.nc` to compare with `land_mask.nc`.  The latter should include land-locked cells so should be safe for sea-ice.  Hopefully, someone can use the paraview extractor to take a look at these 2 in paraview.

milenaveneziani:
Thanks @xylar, that's very helpful. I will run the extractor and look at the 2 masks.

milenaveneziani:
@xylar: I think something went wrong with the creation of `land_mask.nc` because it does not contain the variable `landMaskDiagnostic` as `land_mask_with_land_locked_cells.nc` does. But anyways, I extracted the vtk files for the latter and compared `areaCell` in paraview, comparing it with my original layer (which should *not* have the land-locked cells), and the two plots look identical (same one-cell channels/bays). This does not sound right, so maybe there is some kind of error somewhere (in our diagnostic process, not in compass).

xylar:
@milenaveneziani, the `landMaskDiagnostic` filed is added by `add_land_locked_cells_to_mask()`:
https://github.com/MPAS-Dev/MPAS-Tools/blob/ac85e45a1c128f89fac4e4e226b065a176de3d6c/conda_package/mpas_tools/ocean/coastline_alteration.py#L113
and contains the land mask *before* the changes from that function, flattened into a single mask.  The fields to compare are `regionCellMasks`.  The second one is the Antarctic coast so it may also be worth a look, but the first will be your primary focus.

xylar:
@milenaveneziani, sorry, I'm not very expert with `add_land_locked_cells_to_mask()`. It seems the `landMaskDiagnostic` has different numbers for different purposes. Search the function and you can see various places where it gets filled with different values. @mark-petersen can surely provide more expertise.

xylar:
>  And is there a way to instruct compass to add a land or an ocean cell at will?

@milenaveneziani, no there's not a practical way to do that, in part because JIGSAW will generate a different base mesh on different machines.  Instead, we need to figure out why the algorithms we already have in place are allowing these glitches through.

xylar:
@mark-petersen, indeed, I'm seeing many places like @milenaveneziani pointed out.  Here's an example:
example_problem
Why is `add_land_locked_cells_to_mask()` okay with the one-cell-wide "neck" in the middle of this image?  Is that okay for the MPAS-Seaice B-grid? 

@akturner, maybe you have some insight to offer?

xylar:
@milenaveneziani, how we proceed will depend on what @mark-petersen says.  But I'm pretty sure all of our v2 meshes would be broken if these cells were problematic.  I think what MPAS-Seaice requires is that there is at least one active vertex on the cell with active (non-land) edges adjacent to it.  Therefore, I believe it is the case that one-cell channels like you have pointed out are fine as long as there are active vertex on either side of that cell.

My suggestion would be to proceed with making an initial condition (I've done this already) and performing dynamic adjustment (I haven't tried this yet) to get some experience with that process.  Any changes we might make to the mesh-generation process would likely not affect the parameters used in dynamic adjustment.  Ideally, we would also get mapping files and such from @jonbob and do some G-case testing to see if any MPAS-Seaice problems emerge.

milenaveneziani:
Thanks @xylar for running the adjustment steps. I think we had a different default for disabling Redi in shallow areas for the 80-level meshes. Let me look that up.

xylar:
> but we believe that problems arise because of the very shallow bathymetry (there is probably a total of a few active layers in these areas, even in the 80-level case)

It would be possible to set the minimum number of levels to a higher value so the ocean is not too shallow.  The minimum depth in the 60-layer grid is about 30 m, which is probably many layer for the 80-level grid.

> So, yes, I agree that not all the one-cell wide channels will cause trouble, but maybe we can look more closely at the cases where we know we have had problems in the past.

It should be possible to add some more (custom) critical passages or land blockages for this mesh to keep areas that are giving you trouble more room or block them off.  This won't help with the shallow areas, though.  We do have a mechanism for digging critical passages to a specified depth but it's pretty clumsy and I'd prefer not to use it if deepening the minimum depth is a feasible alternative.

xylar:
@milenaveneziani, keep me posted about the Redi parameters.

xylar:
Thanks @milenaveneziani.  I will update both of those parameters (Redi and minimum depth). 

> I bet that would help. Is the minimum depth a custom parameter in Jigsaw?

No, this is part of making the 3D mesh with bathymetry.  For our cases, JIGSAW just makes the horizontal mesh.

xylar:
@jonbob, wonderful, thank you!

xylar:
Okay, I'm rerunning everything with the updated minimum ocean depth (30 m) and minimum Redi layers (15) in:
```
/lcrc/group/e3sm/ac.xylar/compass_1.1/anvil/test_20220630/arrm
```
I'll keep you posted.

xylar:
> I don't think the mesh changed substantially in the 20220630 test, right?

I presume the horizontal mesh is unchanged and only the vertical grid and dynamic adjustment would be different.

> One last think related to the sea-ice initial condition, before I forget again: it would be good to have the initial 1-m thick ice disk at 80S as default. Maybe it already is like that in compass.

I believe that's an E3SM setting. It's nothing that compass supports. @jonbob or @darincomeau, any insight?

jonbob:
Yes, the 1-m disk seaice initial condition is an mpas-seaice setting for the coupled model -- so we can make sure that gets set appropriately in the mpassi namelist

milenaveneziani:
What about the dynamical adjustment step in compass (which I suppose generates an initial condition): is that for mpas-o only?

xylar:
Yes, it's just running standalone MPAS-Ocean. The MPAS-Seaice initial condition is a disc set up later, as @jonbob said in E3SM.

xylar:
@milenaveneziani and @darincomeau, I haven't heard anything in awhile about this.  I'm assuming that's just because you haven't had time to look at it further.  Just wanting to make sure you're not waiting on anything from me.

milenaveneziani:
@xylar: thanks for checking! We discussed the new mesh at length at our meeting yesterday. Here are the main points:
- we agreed on a minimum depth of 10 m (because there are many coastal areas in the Arctic that we would completely miss if we used 30 m, especially on the Siberian side);
- we also decided to wait for @mark-petersen to see if the presence of the land-locked cells is something that should be looked into a bit more closely. We will still look at the 'problematic' areas (there are about 10) in the new mesh, to see how they look;
- finally, I need to look at a few more channels to make sure that they are open. This is quick and easy.

So, for now this is all on us. I'll probably ping you on slack for updates after this. Enjoy the rest of this special month :)

xylar:
> we agreed on a minimum depth of 10 m (because there are many coastal areas in the Arctic that we would completely miss if we used 30 m, especially on the Siberian side)

This minimum depth doesn't affect the coastline.  The coastline comes from a separate land mask from `geometric_features`.  It just mean that ocean columns will be deepened to 30 m (or 10 m if you prefer).  I'm fine with whatever you decide but I don't see why 10 m minimum depth would obviously be preferable, given the problems you described with the previous mesh.

> we also decided to wait for @mark-petersen to see if the presence of the land-locked cells is something that should be looked into a bit more closely. We will still look at the 'problematic' areas (there are about 10) in the new mesh, to see how they look;

That's obviously fine but I think you should also think about what the algorithm is for defining these problematic areas and how they should be added to @mark-petersen's script that we currently have for detecting problems.  We won't be able to do too much with circles on a paraview plot.  But @mark-petersen can probably translate those into an algorithm in a pinch if need be.

I won't plan to be available for work on this mesh for the next 2 weeks or so.

mark-petersen:
Based on all the plots here, the culling is working as designed. Please see the confluence page [Dynamically locked sea-ice cells:Solutions 3](https://acme-climate.atlassian.net/wiki/spaces/OCNICE/pages/119211250/Dynamically+locked+sea-ice+cells+Solutions+3) for a full explanation.

In the images above there are two classes of circled cells. One set is single-cell wide channels, but they have a width of two _edges_. These are allowed because ice on a B-grid can flow through on the vertex that connects the edges:
image
The second class

milenaveneziani:
First of all, thanks to @mark-petersen for the detailed explanation about the land-locked cell's algorithm. It's very clear now and it all makes sense to me.

So, here is an update from me, as well as next steps for me to follow:
- we will go through the 3 stages of the culling since this process has been vetted for a while now. I would like to add a total of i) 5 critical passages: 3 in the Arctic and 2 that I think are useful to have, one of which is Lombok Strait in the Indonesian Throughflow region; and ii) 3 island groups in the Bering Sea (one of them is the Diomede Islands in the Bering Strait);
- we need to have the minimum depth set to 10 m (thanks @xylar for clarifying what this really means for me). We think this will be OK for an 80-level mesh, since we have 5 layers in the top 10m of the water column.

I would like to start testing all this on my local branch. So I started with adding the new geojson features to my local branch of geometric_data (properly, following the documentation). Now, I am working from this PR compass branch (making sure that I point to the local geometric data cache location) to test how the changes will look like in ARRM10to60 as well as EC30to60, just to make sure that the low resolution mesh will handle well the island blockage features.

I'll update with figures as soon as I have them.

milenaveneziani:
I was able to update my geometric_features repo with the new critical passages and land blockages, and I changed my `compass/ocean/tests/global_ocean/mesh/cull.py` to point to my local cache location of geometric_features, but I don't think that worked. In the sense that, after setting test case 150 (arrm10to60 mesh) and 140 (ec30to60 mesh) and running `compass run`, the critical_passages and critical_blockages features do not contain the new features. I haven't looked at the vtk files for the culled meshes, but I will do that next.

@mark-petersen: do you know if I should do something else in compass to make it use my local geometric_data?

xylar:
@milenaveneziani, are you are changing this line
https://github.com/MPAS-Dev/compass/blob/bf29cdc0164cea2a70858fa5e6c4af9883933d03/compass/ocean/tests/global_ocean/mesh/cull.py#L99
in `cull.py` to this?
```
gf = GeometricFeatures(cacheLocation='/path/to/your/geometric_data')
```
see
https://github.com/MPAS-Dev/geometric_features/blob/bcb98678bbdeaf6f29f452a6df62994e3c0d2668/geometric_features/geometric_features.py#L37
Are you sure that you also installed your local `geometric_features` package?  
```
python -m pip install -e .
```
Did you update the list of features and tags in your local branch (following the docs)?
http://mpas-dev.github.io/geometric_features/stable/adding_features.html

If you tried this and it didn't work, the next suggestion I have is to build the `geometric_features` conda package:
```
rm -rf ~/miniconda3/conda-bld
conda mambabuild recipe
```
Then, make your compass environment with:
```
./conda/configure_compass_env.py --use_local ...
```

milenaveneziani:
Hey @xylar, didn't expect to see you back so soon :)
I did everything you mention, aside for installing the geometric_features environment. I thought that the load compass environment would take care of that. I will try it and report back.
Thanks!

xylar:
@milenaveneziani, if you push your `geometric_features` branch to your fork, I'm happy to test it out.

xylar:
@milenaveneziani, I just rebased to pick up the latest compass changes and switched the minimum depth to be 10 m.  As soon as you point me to your `geometric_featuers` branch, I will do another test of making the mesh and spinning it up.

milenaveneziani:
@xylar: I pushed my geometric features changes.

I still would like to understand why I cannot pick up the changes in compass though..
Here is the change I made to my `cull.py` (paths are for NERSC):
`gf = GeometricFeatures(cacheLocation='/global/project/projectdirs/e3sm/milena/MPAS-git-repositories/geometric_features/geometric_data')`

Then, this is the result of running `compass run` for test case 140, using an interactive node on cori-haswell:
```
(dev_compass_1.2.0-alpha.1) nid00220:milena/global/.../mesh> compass run
ocean/global_ocean/EC30to60/mesh
/global/project/projectdirs/e3sm/milena/MPAS-git-repositories/compass/compass/parallel.py:93: UserWarning: Slurm found 64 cpus per node but config from mache was 32
compass calling: compass.ocean.tests.global_ocean.mesh.Mesh.run()
  in /global/project/projectdirs/e3sm/milena/MPAS-git-repositories/compass/compass/ocean/tests/global_ocean/mesh/__init__.py

Running steps: mesh
  * step: mesh
QStandardPaths: error creating runtime directory '/run/user/69600' (Permission denied)

compass calling: compass.ocean.tests.global_ocean.mesh.Mesh.validate()
  in /global/project/projectdirs/e3sm/milena/MPAS-git-repositories/compass/compass/ocean/tests/global_ocean/mesh/__init__.py

  test execution:      SUCCESS
  test runtime:        20:04
Test Runtimes:
20:04 PASS ocean_global_ocean_EC30to60_mesh
Total runtime 20:08
```

but when I grep for 'name' in test_case_directory/mesh/critical_passages.geojson, none of the new ones are listed.

milenaveneziani:
Granted, the getting started and developer's guides for `compass` were fantastic: I was able to follow all the steps without hiccups.

A couple of notes I took along the way:
- that `QStandardPaths` error that you see in the message above is only thrown for the 140 case, but not for the 150 (ARRM10to60 mesh);
- this documentation line: `./conda/create_compass_env.py --conda $HOME/miniconda3` in the Getting Started guide should be updated with the more recent command (`configure_compass_env.py`).

xylar:
@milenaveneziani, I've never seen that `QStandardPaths` issue before.  I have absolutely no idea what could be causing it.  But I suspect it's unrelated to your problem.

I'm going to try building the `geometric_features` conda package and using that.  I'm not sure what's going wrong with your testing method but it seems more time-consuming to debug it than to create a conda package, which I'm 99% sure will work right.

xylar:
Here's what I do to build (and then use) the `geometric_features` package.  First, in the geometric_features branch, update the version number to 0.7.0 in both the package itself and the recipe.  We will probably want to commit this to your branch so I can make a release as soon as the branch is merged.
```
 vim geometric_features/__init__.py recipe/meta.yaml
# change the version number to 0.7.0 on both files (do :b2 to switch to the second file)
source ~/miniconda3/etc/profile.d/conda.sh
# remove any packages you might have built before, just in case
rm -rf ~/miniconda3/conda-bld
conda mambabuild recipe/
```
Now, in the compass branch:
```
./conda/configure_compass_env.py --env_name compass_arrm --conda ~/miniconda3/ --recreate --use_local
```

milenaveneziani:
Great, thanks for the recipe, I'll take notes.

It's possible (now I can't remember exactly the timeline) that I created the compass environment *before* I changed my geometric_features repo. And I definitely did not update the geometric_features version number.

xylar:
@milenaveneziani, my latest tests are in:
```
/home/ac.xylar/mpas-work/compass/port_arm10to60_mesh
```
on Anvil/Chrysalis.

milenaveneziani:
perfect, thanks! I'll take a look asap.

xylar:
Sorry @milenaveneziani, I somehow posted the wrong path.  I corrected it above and you should have permission to view.

milenaveneziani:
OK, I checked both meshes created with the new critical passages: this ARRM mesh and the EC30to60. I think `compass` is behaving very nicely. See for example the way that it handles the Diomede, St. Peter, and St. George Islands in the Bering Sea:
ARRM:
Screen Shot 2022-07-28 at 4 06 02 PM

EC30to60:
Screen Shot 2022-07-28 at 4 05 23 PM

So, the Diomede Islands disappear in the EC30to60 because otherwise they would probably block Bering Strait, but the other two islands become one in the EC30to60, while being nicely defined in the ARRM mesh.
This seems acceptable to me, would you agree, @xylar and @mark-petersen?

milenaveneziani:
The other passage that seemed tricky to me was Cardigan Strait in the Canadian Northwest Passages. Here is how it appears in the two meshes:
ARRM10to60:
Screen Shot 2022-07-28 at 4 02 27 PM

EC30to60:
Screen Shot 2022-07-28 at 4 03 01 PM

This also seems OK to me.

milenaveneziani:
I would like to take a look at the initial condition and bathymetry next. Should I look at this file: `/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220727/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/dynamic_adjustment/simulation/output.nc` or is it the restart files in the `restarts` subdirectory?

xylar:
> So, the Diomede Islands disappear in the EC30to60 because otherwise they would probably block Bering Strait, but the other two islands become one in the EC30to60, while being nicely defined in the ARRM mesh.
This seems acceptable to me, would you agree, @xylar and @mark-petersen?

I think that's fine.  I think it's important to keep in mind that the bathymetry still provides a certain amount of blockage even if the land boundary isn't there.  I think there's really only so much you can do at EC30to60 resolutions because you either block too much or too little whatever choice you make and the bathymetry is overly smooth regardless of what the coastline looks like.

xylar:
> I would like to take a look at the initial condition and bathymetry next. Should I look at this file: `/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220727/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/dynamic_adjustment/simulation/output.nc` or is it the restart files in the `restarts` subdirectory?

The final `restart` file is very close to the E3SM initial condition.  However, the relevant path is this:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220727/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/files_for_e3sm/assembled_files/inputdata/ocn/mpas-o/ARRM10to60E2r1
```
This was not available yesterday because validation of the final simulation failed due to the known extremes in temperature produced by Redi (>34 C, whereas we permit only <=33 C).  I'm now running `files_for_e3sm` despite the validation failure so the files should be there now.

xylar:
@milenaveneziani, I ran things again with the new bathymetry here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220730/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60
```
In particular, the initial condition for E3SM is here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220730/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/files_for_e3sm/assembled_files/inputdata/ocn/mpas-o/ARRM10to60E2r1
```

milenaveneziani:
Thanks @xylar.

Now the bathymetry is definitely fixed: I'm glad we were able to fix this with the new GEBCO/Bedmachine versions.

Here is the **old** Chukchi bottomDepth in this mesh:
Screen Shot 2022-07-29 at 9 24 30 AM

and here it is now with the new bathymetry file:
Screen Shot 2022-07-30 at 9 48 28 PM

so that's great.

I then looked at what I think should be the initial condition for the e3sm_files:
`/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220730/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/files_for_e3sm/assembled_files/inputdata/ocn/mpas-o/ARRM10to60E2r1/mpaso.ARRM10to60E2r1.220730.nc`
and specifically at T and S for various depths.
The values look fine for the first 5 layers, but below that all T and S values are 0.. Which is of course weird and concerning. Any idea what might be going on?

xylar:
> The values look fine for the first 5 layers, but below that all T and S values are 0.. Which is of course weird and concerning. Any idea what might be going on?

That's really discouraging and I have no idea what could be causing it.  Can you try to trace it back starting at `PHC/init/initial_state/` to see if you can see these same zeros?

I won't have time to look at this until at least Tuesday and maybe later.

milenaveneziani:
I looked at `temperature` and `salinity` in `initial_state.nc` (I used matlab because I wasn't sure what paraview was showing..). As before, the first 5 layers look fine, with sensible ranges of salinity and temperature, but below layer 5 things start to look weird. My impression is that the 0 value is used instead of a NaN whenever `k>maxLevelCell`. Which I suppose is OK since those values are masked out. But I should double check that.
I will look at the initial condition for E3SM using matlab next, to see if I get similar patterns.

Here are some matlab plots (using `scatter`) from initial_state
Screen Shot 2022-07-31 at 11 55 40 PM
Screen Shot 2022-07-31 at 11 57 45 PM
Screen Shot 2022-08-01 at 12 02 38 AM
Screen Shot 2022-08-01 at 12 03 59 AM
:

milenaveneziani:
ok, I verified that both initial_state and the E3SM initial condition are fine: I was just not visualizing the land-sea mask properly at depth. So, false alarm. Phew!

I think we are ready to try a E3SM case. What would be the next step then?

xylar:
@milenaveneziani, that's great! Quite a relief!

The next step is to work with @jonbob and @darincomeau on creating mapping, domain, etc. files for E3SM. I don't (yet) have the expertise to help here

jonbob:
@milenaveneziani - please just let me know when you think you're happy with the grid and I can start making the mapping and domain files for it. All I should need in an ocean file with initial conditions and the mesh info. And I assume you want support for CORE-II, JRA and ne30pg2?

darincomeau:
@jonbob We're good to go here for the mapping/domain files.

mark-petersen:
@xylar, this mesh looks great, thanks for the work on it. When you create the new directory on the [E3SM input data server](https://web.lcrc.anl.gov/public/e3sm/inputdata/), could you please add the netcdf forcing file created by init mode, and the original `graph.info` file? Most of our cases do not have the forcing file, but it is useful for stand-alone testing, and easy to add at this point. Most have the `graph.info` file, but a few in the past only had the `*.part` files, and that was a snag for running on new partition numbers.

darincomeau:
@mark-petersen what exactly is the 'netcdf forcing file created by init mode', or rather, where do I find it in the compass output here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220730/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/
```
I'll stage it to the inputdata server (I've already done it for the original `graph.info` file as requested).

xylar:
@mark-petersen and @darincomeau, I would prefer that we wait to copy files into `inputdata` until @milenaveneziani and others have taken a look at some E3SM tests.  I could imagine that there will be another iteration of the mesh.  It wouldn't be a big deal if it gets the next revision number but our usual procedure is to only consider the mesh official once E3SM testing has happened.

darincomeau:
@xylar I actually already staged everything on `inputadata`. I realize this is earlier than we have done in the past, but I wanted to be able to run quick smoke tests at LCRC, and do a longer run on Cori. If there is another iteration of this mesh, I'll remove these files from the inputdata directory.

darincomeau:
Just to update here, I've got the integration branch ready [here](https://github.com/E3SM-Project/E3SM/tree/darincomeau/grids/add-ARRM10to60E2r1). I've successfully run smoke tests on the four configurations (2 G-cases, 2 B-cases), and have a long low-res B-case running on Cori. It's going to take some time for that run to complete, so I'm holding off on opening the E3SM PR for that.

xylar:
> and have a long low-res B-case running on Cori.

@darincomeau, this means low-res atmosphere?

darincomeau:
@xylar yes, `ne30pg2_ARRM10to60E2r1`

mark-petersen:
@darincomeau the file I was referring to is here
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220730/ec_arrm_for_e3sm/ocean/global_ocean/ARRM10to60/PHC/init/initial_state/init_mode_forcing_data.nc
```
It includes winds and T&S restoring fields that are used as a simple mock-up of surface forcing in stand-alone. Thanks.

xylar:
@mark-petersen, maybe you want to make a PR to include a symlink to `init_mode_forcing_data.nc` in `assembled_files`?  That way, it would be more likely to be included with any new mesh, rather than being an additional manual step.

xylar:
I will need to fix this branch following the merge of #413.  Hopefully, I can do that tomorrow.

darincomeau:
Just FYI the progress of the B-case is going *very* slowly on Cori, and the machine is down for maintenance tomorrow. Only completed 7 years so far.

xylar:
@darincomeau, I know resources are what they are but might it make sense to do the testing on Compy or Anvil instead?

xylar:
Rebased onto #429, so I moved it back to draft mode and "in progress" until that gets merged.

xylar:
I retested this after the rebase onto #429.  Everything ran successfully except that we're still seeing undesirable temperature extremes because of Redi:
```
ValueError: Max of temperatureMax > allowed threshold: 36.71978992754585 > 33.0
```
Results are in:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/anvil/test_20220817/arrm_for_e3sm
```

milenaveneziani:
> except that we're still seeing undesirable temperature extremes because of Redi

Do you think we should up the min # layers to 16?

milenaveneziani:
We can check if the setting `config_redi_min_layers_diag_terms=16` avoids the temperature error check. For the 80-level mesh, that would correspond to an average depth of 66m, so that's not that different from 60.

xylar:
@milenaveneziani, I don't think one more layer will make an appreciable difference based on past experience.  We just need to fix Redi so this doesn't happen.

xylar:
@milenaveneziani, @darincomeau and everyone, I am opening this back up for review after #429 just got merged.  I'm happy to run this mesh through its paces again but wanted to make any changes you all might want before doing that.

darincomeau:
Thanks @xylar. To update on the B-case, I'm *almost* at 20 years (3 months to go) on the Cori run, at which point I'm hoping we'll have enough to be able to approve the mesh, at which point I'll open the E3SM PR. I realize previously we did 50 years, but I'm hoping 20 years will be sufficient. The throughput on Cori has been incredibly slow; as to the Anvil suggestion, I was a bit wary of running a simulation there since this is primarily for a different project, especially given how busy that machine has been lately, and we'd still probably get only about 2 SYPD on a medium size layout.

xylar:
@darincomeau, thanks for the update.  I can see the challenges with Anvil, though it sounds like 2 SYPD would still be an improvement on what you're getting on Cori.

darincomeau:
[Here is the MPAS-Analysis for years 11-20 of the B-case (finally)](https://portal.nersc.gov/project/m1199/dcomeau/E3SMv2-Arctic/20220803.WCYCL1950.ne30pg2_ARRM10to60E2r1.baseline.cori-knl/mpas_analysis/ts_0011-0020_climo_0011-0020/ocean/index.html)

milenaveneziani (approved):
I think this is good to go.

darincomeau:
Linking the E3SM PR: https://github.com/E3SM-Project/E3SM/pull/5178

xylar:
Thanks @darincomeau.  I created a permanent branch (https://github.com/MPAS-Dev/compass/tree/ARRM10to60E2r1) and tag (https://github.com/MPAS-Dev/compass/tree/mesh_ARRM10to60E2r1) for the version of this mesh you used in https://github.com/E3SM-Project/E3SM/pull/5178.  I'll review the E3SM PR as soon as I can.

darincomeau (approved):
@xylar thanks for all this! (Sorry I meant to do this yesterday).

xylar:
Thanks! merging...

PR #417, opened 2022-07-05
Port legacy dam break test case

cbegeman:
This PR ports the legacy "dam break" test case. This test case is modeled after Warner et al. (2013) and includes comparison between MPAS-Ocean output, experimental data, and ROMS simulation data at a handful of locations within the domain. 
image


cbegeman:
Note: documentation to be added soon

compass/ocean/tests/dam_break/viz/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+        self.add_output_file(filename='dam_break_comparison.png')
+
+    def run(self):
+        """
+        Run this step of the test case
+        """
+        # read output.nc
+        data = xarray.open_dataset('output.nc')
+        xCell = data.xCell.values
+        yCell = data.yCell.values
```
xylar, line 65: Here and elsewhere in this module, I would ideally have variables at are all lowercase, possibly with underscores.  But since this is a port of a legacy code, I don't think it's a problem.

I do think we want to discourage folks from using this test group as a starting point for other work, though.  (Also because of how the initial condition is created.)  Just something I want us to keep in mind.

docs/developers_guide/ocean/test_groups/dam_break.rst
```diff
@@ -0,0 +1,69 @@
...
+domain in plan view.
+
+
+.. _dev_ocean_dam_break_default:
+
+default
+-------
+
+The :py:class:`compass.ocean.tests.dam_break.default.Default`
+test performs one 10-minute run on 4 cores. It doesn't contain any
```
xylar, line 65: If you decide to change the number of cores as I suggested, please update this, too.

xylar (approved):
This looks great!  I have a few small comments that you can take as suggestions but I think this is ready to go regardless of what you decide to do with them.

I ran the 120 cm test case on my laptop with no trouble.  The 40 cm case is still running and presumably will take a few hours on 4 cores (part of the reason I suggest upping the core count).  I'll let you know if that gives me any trouble.

xylar:
@sbrus89, I'm not sure if you're available at all for reviews at the moment.  If so, please have a look.  If not, maybe you can make a guess when you might have time again.

compass/ocean/tests/dam_break/default/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+            res_name = f'{int(resolution*1e3)}cm'
+        else:
+            res_name = f'{int(resolution)}m'
+        subdir = f'{res_name}/{name}'
+        super().__init__(test_group=test_group, name=name,
+                         subdir=subdir)
+
+        self.add_step(InitialState(test_case=self))
+        self.add_step(Forward(test_case=self, resolution=resolution,
+                              cores=4, threads=1))
```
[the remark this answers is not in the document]
cbegeman: Thanks for the detailed suggestion! I increased the core count for both cases as you suggested.

compass/ocean/tests/dam_break/namelist.init
```diff
@@ -0,0 +1,6 @@
+config_ocean_run_mode = 'init'
+config_init_configuration='dam_break'
+config_dam_break_vert_levels=3
+config_dam_break_R0=12.0
+config_drying_min_cell_height=1.0e-2
+# Consider: config_write_cull_cell_mask=.false.
```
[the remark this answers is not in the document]
cbegeman: Now removed

compass/ocean/tests/dam_break/default/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        name = 'default'
+
+        self.resolution = resolution
+        if resolution < 1.:
+            res_name = f'{int(resolution*1e3)}cm'
+        else:
+            res_name = f'{int(resolution)}m'
+        min_cores = int(40/(resolution/0.04)**2)
+        cores = 10*min_cores
```
xylar, line 40: Yep, this is great!

xylar:
@cbegeman, this will need some tweaks now that we have made some framework changes in #413.  `cores` and `min_cores` attributes no longer exist and will instead need to be `ntasks` and `min_tasks`.  `threads` is now `openmp_threads`.  Let me know if you'd like me to push some fixes or if you want to handle it yourself.

cbegeman:
@xylar Thanks for the heads up! Done

tuanggolt:
open tuanngo362@gmail.com

compass/ocean/tests/dam_break/viz/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+    def run(self):
+        """
+        Run this step of the test case
+        """
+        # read output.nc
+        data = xarray.open_dataset('output.nc')
+        x_cell = data.xCell.values
+        y_cell = data.yCell.values
+        ssh = data.ssh.values
+        ssh = ssh + 0.6
```
sbrus89, line 64: @cbegeman, What is the reason for the 0.6 offset here?

sbrus89 (approved):
@cbegeman, I apologize for the delayed review on this. Everything looks great to me. I tested on Anvil and this worked great. I only had one minor comment in the viz.

compass/ocean/tests/dam_break/viz/__init__.py
cbegeman: That's the value of `config_dam_break_eta0`, which `bottomDepth` is set to. I added this commit [232e370](https://github.com/MPAS-Dev/compass/pull/417/commits/232e370a43cb2cb5f4a87b729d0998e0c66d3611) to use `bottomDepth` directly. Let me know if you like that solution.

sbrus89: I like that solution very much. Thanks!

cbegeman:
@xylar Ok, I think we're all set to merge!

xylar:
Thanks @cbegeman for the careful work on porting this test case!  Thanks for the review @sbrus89.

PR #423, opened 2022-07-13
Give log path for failed tests only in suite

### LINKED ISSUE #398: Incorrect log file message when running individual testcase
```
  test execution:      ERROR
  see: case_outputs/landice_ismip6_run_ais_Antarctica_8to80km_projection.log
```

xylar:
When test cases are run on their own, they don't put a log file in `case_outputs` so we shouldn't incorrectly point to one.

This merge also clean things up by converting calls to the `.format()` method to f-strings.

closes #398

### DIFF
--- a/compass/run/serial.py
+++ b/compass/run/serial.py
@@ -183,7 +183,8 @@ def run_tests(suite_name, quiet=False, is_test_case=False, steps_to_run=None,
                     success[test_name] = pass_str
                 else:
                     logger.error(status)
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the `nightly` test suite on Linux with Gnu compilers and saw the normal output (all tests passed).

I ran a single test in its work directory (deliberately causing it to fail) and saw the expected stack trace and error message but not the message about a log file in `case_output`.

PR #424, opened 2022-07-26
Fix namelist errors in internal_wave

altheaden:
Omitting `mode='init'` seems to have lead to any namelist replacements given at runtime being ignored. This fix lead me to a secondary bug, where we were attempting to add floats to the namelist file. These issues may lead to non-bit-for-bit changes in `internal_wave` runs.

### DIFF
--- a/compass/ocean/tests/internal_wave/initial_state.py
+++ b/compass/ocean/tests/internal_wave/initial_state.py
@@ -27,10 +27,10 @@ class InitialState(Step):
                          min_cores=1, threads=1)
 
         self.add_namelist_file('compass.ocean.tests.internal_wave',
-                               'namelist.init')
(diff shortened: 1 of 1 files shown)


xylar:
@altheaden, thanks for making this PR.  I'm testing it right now with Intel and Intel-MPI on Anvil to see if it's bit-for-bit or not.

xylar (approved):
Yep, this passes and results are bit-for-bit with the `ocean/internal_waves/default` test case.  As long as @cbegeman agrees with this fix, I think it's right.

cbegeman (approved):
This looks right to me. @altheaden, thanks for catching these bugs! Let me know if you'd like me to do additional testing; I should be able to get to that tomorrow.

PR #425, opened 2022-07-27
Add function for logging function calls

xylar:
The function `log_function_call()` is similar to `log_method_call()` but for functions instead of methods.

This will be useful in future work where we are changing some methods to be functions instead.

### DIFF
--- a/compass/logging.py
+++ b/compass/logging.py
@@ -78,7 +78,7 @@ def log_function_call(function, logger):
         The function that will be run immediately following this call
 
     logger: logging.Logger
-        The logger to log the method path and file path to
+        The logger to log the function path and file path to
     """
     if not inspect.isfunction(function):
         raise ValueError('The "function" argument must be a function')
--- a/compass/logging.py
+++ b/compass/logging.py
@@ -1,5 +1,4 @@
 import inspect
-import functools
 
 
 def log_method_call(method, logger):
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the ocean `nightly` test suite on my Ubuntu laptop.  All tests ran successfully and I see:
```
compass calling: compass.parallel.set_cores_per_node()
  in /home/xylar/code/compass/generalize_log_method_call/compass/parallel.py
```
as expected in the test-case log file when I added a test call to `log_function_call()`.

compass/logging.py
```diff
@@ -64,3 +63,29 @@ def log_method_call(method, logger):
...
+      compass calling: compass.parallel.set_cores_per_node()
+        in /home/xylar/code/compass/compass/compass/parallel.py
+
+    Parameters
+    ----------
+    function : function
+        The function that will be run immediately following this call
+
+    logger: logging.Logger
+        The logger to log the method path and file path to
```
altheaden, line 81: This should be `The logger to log the function path and file path to`

altheaden (approved):
Other than the one typo, this looks good to me.

compass/logging.py
xylar: Done!

xylar:
Thanks @altheaden.  Please give it another look and try testing this branch as well if you get a chance.

xylar:
@altheaden, I'm going to merge this based on my testing.  We can always make further fixes if needed.

PR #426, opened 2022-07-27
Update E3SM-Project submodule to current master

xylar:
This is commit  [6b81271](https://github.com/E3SM-Project/E3SM/tree/6b8127137728101ff719e3726a2a7f636921be40).

This update includes the following MPAS-Ocean PR:
* https://github.com/E3SM-Project/E3SM/pull/5026


xylar:
## Testing

I ran the ocean `pr` test suite using the current submodule as a baseline.  I just tested on Anvil with Intel and Intel-MPI since there has only been one PR of interest since the last submodule update and because that is expected to leave all tests bit-for-bit.  Indeed, all tests in the suite are bit-for-bit.

PR #422, opened 2022-07-08
Add jigsaw spherical mesh steps

xylar:
There is a base class and classes for creating either icosahedral subdivision meshes or meshes from a map of cell width.

The `QuasiUniformSphericalMeshStep` class creates a quasi-uniform base mesh, and allows the `build_cell_width_lat_lon()` an `make_jigsaw_mesh()` methods to be overridden to change the functionality (e.g. to build a map of cell width that varies with space).

The `IcosahedralMeshStep` class creates a subdivided icosahedral mesh defined either by its approximate cell width or the number of subdivisions.

This merge also updates the `global_convergence` test group to make use of these new classes.  There are now `qu` and `icos` versions of the `cosine_bell` test case.

The documentation has been updated based on these changes.

PR #429, opened 2022-08-03
Switch `global_ocean` and `hurricane` to framework base and culled mesh

xylar:
This allows us to use the more generalized spherical mesh step class with better JIGSAW support from #422 and breaks the `mesh` test into `base_mesh` and `cull_mesh` steps.


xylar:
This still needs documentation updates.

xylar:
## Testing

The ocean `pr` test suite passes on Anvil with Intel and Intel-MPI.  It is bit-for-bit except for the mesh test cases (because of the reorganization and the new cache files).

sbrus89:
@xylar, yes. I’ll take a look.

xylar:
I will rebase and deal with conflicts tomorrow.

xylar:
@sbrus89, I rebased this branch after #413 and (after a few missteps) all the tests pass as before (see https://github.com/MPAS-Dev/compass/pull/429#issuecomment-1214404282 above).

sbrus89 (approved):
@xylar, I have tested this and approve the qualitative differences in the hurricane case solution resulting from a non BFB mesh. Thanks for making this change!

xylar:
Thanks @sbrus89!  I'll go ahead and merge this.

PR #433, opened 2022-08-31
Remove `dThreshMLD` validation from mixedLayerDepth AM

xylar:
As of https://github.com/E3SM-Project/E3SM/pull/5099, this variable is no longer part of the AM and is instead part of the main MPAS-Ocean code.


xylar:
@mark-petersen, I think a review by inspection is sufficient in this case.  I have been seeing failures in the `analysis_test` test case in the `pr` test suite when I compare with the latest E3SM `master` because this variable is no longer available in either the baseline or the branch I'm comparing it against.

mark-petersen (approved):
Yes, this makes sense because that variable was removed from the output. Thanks!

xylar:
Thanks @mark-petersen!

PR #436, opened 2022-10-12
Remove `--exclusive` flag from slurm steps

xylar:
This is causing compass to hang in interactive jobs


xylar:
@sbrus89, you have already been testing with this change, so you can approve by inspection if you like.

sbrus89 (approved):
Approved, this is the exact change I've been testing with and it fixes the hanging issue.

PR #437, opened 2022-10-12
Add a step for making mapping files for `e3sm_to_cmip`

xylar:
To make things easier for the team publishing CMIP6 data from E3SM simulations, we want a set of mapping files from each MPAS-Ocean (and -Seaice) mesh to a standard CMIP6 grid.  This merge adds a step to the `files_for_e3sm` test case for making these mapping files.

It also adds this capability to the `make_diagnostic_files` to test case so these mapping files can be created (along with other diagnostics files) for existing meshes.


xylar:
## Testing

I ran the `qu240_for_e3sm` suite on Chrysalis and verified that the CMIP6 mapping files were generated and symlinked in the expected place in `assembled_files`.

I ran the `make_diagnostic_files` test case pointing to the EC30to60E2r2 and ECwISC30to60E2r1 meshes, and verified that it produced the expected mapping files (and other output files)

PR #438, opened 2022-10-20
Drop compass_ prefix from job script names

xylar:
This prefix makes things harder to find and tab complete, and doesn't add any clarity since it is always obvious that each job script is a compass job script.


xylar:
@matthewhoffman, @mark-petersen, @sbrus89, @cbegeman and @trhille, as frequent compass users, I'm just making you aware of this small change.  It's been annoying me for awhile.

PR #442, opened 2022-10-23
Add a sigma vertical coordinate to ocean framework

xylar:
This merge also adds ISOMIP+ test cases that demonstrate the usage of the coordinate.

In the process, a small bug was found in the ISOMIP+ visualization in which the colormap bounds were incorrectly including zero even for fields (like salinity) where this is not appropriate.  This has also been fixed here for convenience.

### DIFF
--- a/compass/ocean/tests/isomip_plus/__init__.py
+++ b/compass/ocean/tests/isomip_plus/__init__.py
@@ -21,10 +21,13 @@ class IsomipPlus(TestGroup):
                         OceanTest(test_group=self, resolution=resolution,
                                   experiment=experiment,
                                   vertical_coordinate=vertical_coordinate))
-            for experiment in ['Ocean0']:
-                for vertical_coordinate in ['z-star']:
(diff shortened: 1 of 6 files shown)


xylar:
## Testing

I ran the `performance` and then 1 month of the `simulation` step in the `2km/sigma/Ocean0` test case.

The results look as expected, e.g. here is a cross section of potential temperature:
sectionTemp_0001

And here is the overturning streamfunction:
osf_0001

xylar:
@cbegeman, this is ready for you to review.  I think all you need to do is describe the testing you've already done (since I believe you are already using this coordinate and the sigma versions of the ISOMIP+ test cases).

cbegeman (approved):
Approving based on testing on chrys with intel-mpi. I ran both the included 10-layer sigma `Ocean0` case and a 3-layer sigma `time_varying_Ocean0` case.

xylar:
Thanks for the review, @cbegeman!

PR #443, opened 2022-10-24
Switch from 2- to 4-digit year in global ocean

xylar:
In the date stamp in output files and metadata, it is significantly less ambiguous to have a 4-digit year.


xylar:
## Testing

I tested both `files_for_e3sm` and `make_diagnostics_files` on the QU240 mesh on my Linux laptop.  Both produced 4-digit years in all date stamps.

PR #447, opened 2022-10-31
Add more projection grids to diagnostics_files

xylar:
MPAS-Analysis now supports 5 more projection comparison grids, and we want to generate the associated mapping files for each new MPAS-Ocean and -Seaice mesh.

For now, 2 functions have largely been copied from MPAS-Analysis. In the near future, it would be best to handle these in a better way.


xylar:
## Testing

I uses this to generate mapping files for 5 new projection grid for 4 E3SM v2 meshes.  These have been added to the public `diagnostics` directory on LCRC and will be synced to other machines along with other diagnostics data as part of installing the next test version of E3SM-Unified.

PR #451, opened 2022-11-06
Bugfix for dam break tests

cbegeman:
In order to avoid crashes of the dam break test case, tracers need to be initialized  (currently handled in init mode and fixed here https://github.com/E3SM-Ocean-Discussion/E3SM/pull/33), and those tracers also need to be stored in the initial state stream (fixed here). 

Related changes also included here are (1) adding the validation step to the test case and (2) disabling tracer tendencies, which are not needed in this test case.

### DIFF
--- a/compass/ocean/tests/dam_break/default/__init__.py
+++ b/compass/ocean/tests/dam_break/default/__init__.py
@@ -3,6 +3,7 @@ from compass.testcase import TestCase
 from compass.ocean.tests.dam_break.initial_state import InitialState
 from compass.ocean.tests.dam_break.forward import Forward
 from compass.ocean.tests.dam_break.viz import Viz
+from compass.validate import compare_variables
 
 
 class Default(TestCase):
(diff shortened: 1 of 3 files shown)


cbegeman:
Note: The results of this test case are not BFB with master prior to https://github.com/E3SM-Project/E3SM/pull/5154, but the general behavior with respect the the SSH field is unchanged.

sbrus89 (approved):
Approved based on testing this on Compy in conjunction with https://github.com/E3SM-Project/E3SM/pull/5279. Thanks @cbegeman. Sorry again for the delay on this.

xylar (approved):
This looks great to me! I tested along with https://github.com/E3SM-Project/E3SM/pull/5279 on Chrysalis with Intel and Intel-MPI.  I tested both 40 and 120 cm cases.

I tested the validation as well (by rerunning the test and comparing with the previous run) and that also worked as expected.

xylar:
I will merge this as soon as https://github.com/E3SM-Project/E3SM/pull/5279 has been merged and we update the `E3SM-Project` submodule to include that update.

xylar:
Thanks @cbegeman!!!

PR #452, opened 2022-11-08
Port merry-go-round test group from legacy

cbegeman:
This PR ports the merry-go-round test group from legacy. The legacy test group was developed by @scalandr https://github.com/MPAS-Dev/compass/pull/108. The original test case is here named `convergence_test` and I added a `default` test that is just the lowest resolution forward run from the `convergence_test`. 

The initial state was found to have only machine-precision level differences from the legacy version, and it also shows the same convergence behavior when compared with legacy using the same MPAS-Ocean code base. 

The main differences between this implementation and that in legacy are
- using `init_vertical_coord` rather than hard-coding the vertical coordinate
- using matrix operations in `initial_state` to speed up computations
- minor tweaks to visualizations in `viz` and `analysis` steps


cbegeman:
Old analysis plot:
image
New analysis plot: 
image

cbegeman:
Old viz plot:
image
New viz plot: 
image

cbegeman:
@scalandr The porting process went really smoothly! 

Could you give [initial_state](https://github.com/MPAS-Dev/compass/blob/f739bb548ca42f87a66d8227b90c71210733db0a/compass/ocean/tests/merry_go_round/initial_state.py) a look? I removed your density initialization because it did not appear to be necessary. 

Other than that, feel free to review or test with whatever rigor you prefer and have time for.

cbegeman:
@mark-petersen Pinging you on this, since you reviewed the legacy version, if you want to take a look.

cbegeman:
@scalandr Thanks for catching that! I renamed the config options and missed that one.

mark-petersen (approved):
Beautiful! Thank you @cbegeman. I tested successfully on badger:
<details>
<summary> report: </summary> <p>


```
compass list |grep merry
 213: ocean/merry_go_round/default
 214: ocean/merry_go_round/convergence_test
(dev_compass_1.2.0-alpha.1) compass setup -p /usr/projects/climate/mpeterse/repos/E3SM/master/components/mpas-ocean -w $n/221115_merry -n 213 214
Setting up test cases:
  ocean/merry_go_round/default
  ocean/merry_go_round/convergence_test
target cores: 4
minimum cores: 4

salloc -N 1 -t 2:0:0 --qos=debug --reservation=debug --account=t22_ocean_time_step

cr
ocean/merry_go_round/default
compass calling: compass.ocean.tests.merry_go_round.default.Default.run()
  in /turquoise/usr/projects/climate/mpeterse/repos/compass/master/compass/testcase.py

compass calling: compass.run.serial._run_test()
  in /turquoise/usr/projects/climate/mpeterse/repos/compass/master/compass/run/serial.py

Running steps: initial_state_5m, forward_5m, viz_5m
  * step: initial_state_5m
  * step: forward_5m
  * step: viz_5m

compass calling: compass.ocean.tests.merry_go_round.default.Default.validate()
  in /turquoise/usr/projects/climate/mpeterse/repos/compass/master/compass/ocean/tests/merry_go_round/default/__init__.py

  test execution:      SUCCESS
  test runtime:        00:04
Test Runtimes:
00:04 PASS ocean_merry_go_round_default
Total runtime 00:05
ba440:g:default$ cd ../
convergence_test/ default/
ba440:g:default$ cd ../convergence_test/
ba440:g:convergence_test$ cr
ocean/merry_go_round/convergence_test
compass calling: compass.ocean.tests.merry_go_round.convergence_test.ConvergenceTest.run()
  in /turquoise/usr/projects/climate/mpeterse/repos/compass/master/compass/testcase.py

compass calling: compass.run.serial._run_test()
  in /turquoise/usr/projects/climate/mpeterse/repos/compass/master/compass/run/serial.py

Running steps: initial_state_1.25m, forward_1.25m, viz_1.25m, initial_state_2.5m, forward_2.5m, viz_2.5m, initial_state_5m, forward_5m, viz_5m, analysis
  * step: initial_state_1.25m
  * step: forward_1.25m
  * step: viz_1.25m
  * step: initial_state_2.5m
  * step: forward_2.5m
  * step: viz_2.5m
  * step: initial_state_5m
  * step: forward_5m
  * step: viz_5m
  * step: analysis

compass calling: compass.ocean.tests.merry_go_round.convergence_test.ConvergenceTest.validate()
  inherited from: compass.testcase.TestCase.validate()
  in /turquoise/usr/projects/climate/mpeterse/repos/compass/master/compass/testcase.py

  test execution:      SUCCESS
  test runtime:        01:07
Test Runtimes:
01:07 PASS ocean_merry_go_round_convergence_test
Total runtime 01:07
```
</details>
and the images were produced successfully. Thanks for your attention to detail on those images.

scalandr:
Hi @cbegeman, I checked initial_state and everything looks fine to me. I tested successfully on Anvil.
As you saw, the default advection types for this test case are
```
config_vert_tracer_adv_order=2
config_horiz_tracer_adv_order=2
config_monotonic=.false.
```
I was wondering if we have to test what happens with a third-order advection scheme or with monotonic on. This goes beyond the test case itself, but I wanted to ask for your opinion.

cbegeman:
@scalandr I'm glad you asked that question. It led me to realize that the namelist files do not have the updated namelist options so I fixed that here: https://github.com/MPAS-Dev/compass/pull/452/commits/72ac942c0c9b6c86378290070d451709c181003b. The results I showed earlier are with the monotonic scheme on.

Interestingly, when the monotonic scheme is on, the results are the same for 2nd and 3rd order. When the monotonic scheme is off, we get lower error overall. Changing the order of accuracy doesn't change the slope but does reduce the error:

**2nd order, monotonic**
image

**2nd order, non-monotonic**
image

**3rd order, non-monotonic**
image

scalandr:
Hi @cbegeman, from your plots is clear that monotonic has, as expected, an impact on the convergence rate. I re-did your plots and added an order 1 line, so that is even more clear that the global order of the model goes down to 1 if we move from a standard finite volume advection discretization to a monotonic one. This result is expected, but is nice to see it confirmed in this test. 

**monotonic advection:**
convergence_plot_monotonic

**non-monotonic advection:**
convergence_plot_standardFV

scalandr (approved):
Hi @cbegeman, I think it all looks great. 
I believe we should keep `config_flux_limiter = 'none'` as default exactly as it is right now. I am happy that this test case will be in the official compass list!

cbegeman:
@scalandr Thank you so much for your thorough review! In the spirit of code coverage, we might consider adding a version that is monotonic in addition to the non-monotonic test later on. But I'm happy to merge this PR with only the non-monotonic tests.

compass/ocean/tests/merry_go_round/forward.py
```diff
@@ -0,0 +1,79 @@
...
+    # no setup() is needed
+
+    def run(self):
+        """
+        Run this step of the test case
+        """
+        run_params = {'5m':    {'config_dt': '00:12:00'},
+                      '2.5m':  {'config_dt': '00:06:00'},
+                      '1.25m': {'config_dt': '00:03:00'}}
+        self.update_namelist_at_runtime(run_params[self.resolution])
```
xylar, line 78: Is there a reason to do this at runtime?  If so, it isn't immediately clear to me.  I would do this in the constructor with `self.add_namelist_options()`.  That would mean a user could modify `config_dt` if they had some reason to and it wouldn't get overridden when they call `compass run`.  If a user could change the resolution at runtime, it would make sense to compute the corresponding `dt` at runtime (and that is the case for some of our test cases like `cosine_bell`).  But I think the resolution is chosen based on which test case you set up here.

compass/ocean/tests/merry_go_round/initial_state.py
```diff
@@ -0,0 +1,192 @@
...
+        cell2 = ds.cellsOnEdge[:, 1].values - 1
+        for iEdge in range(0, nEdges):
+            if ((cell1[iEdge] < 0.5*nCells and
+                 cell1[iEdge] > 0.25*nCells - 1) and
+                (cell2[iEdge] < 0.5*nCells and
+                 cell2[iEdge] > 0.25*nCells - 1)):
+                mask[iEdge] = True
+            elif (cell1[iEdge] > 0.75*nCells-1 and
+                  cell2[iEdge] > 0.75*nCells-1):
+                mask[iEdge] = True
```
xylar, line 162: This would be substantially more efficient if it used `numpy.logical_and` and `numpy.logical_or`.  It would also seems strange to use the cell index rather than a physical coordinate _(I think `xCell` in this case) to do this masking.  As it is, I don't entirely understand what this code is doing.

```suggestion
x1 = xCell[cell1]
x2 = xCell[cell2]
xQuarter = 0.75*xMin + 0.25*xMax
xThreeQuarters = 0.25*xMin + 0.75*xMax
mask = numpy.logical_or(
    numpy.logical_and(
        numpy.logical_and(x1 < xMid, x1 >= xQuarter),
        numpy.logical_and(x2 < xMid, x2 >= xQuarter)),
    numpy.logical_and(x1 > xThreeQuarters, 
                      x2 > xThreeQuarters)))
```

xylar (changes requested):
@cbegeman, wonderful work as always!  @scalandr, again I appreciate your work on the original test case!

I'm testing this now but based on looking through the code, there are a few changes it would be nice to have.

The only one I feel strongly about is the mask, which I don't think is very easy to follow with the current logic based on cell index.

xylar:
My actual runs of the code on Chrysalis with Intel and Intel-MPI went very smoothly.  The viz and analysis plots look great!

compass/ocean/tests/merry_go_round/forward.py
cbegeman: Addressed by https://github.com/MPAS-Dev/compass/pull/452/commits/5fda332b0f4ce20466de02d541d3c353d8a462fd

compass/ocean/tests/merry_go_round/convergence_test/analysis.py
```diff
@@ -0,0 +1,103 @@
...
+        The output file name
+
+    resolutions : list of str
+        The resolutions of the test case
+    """
+    plt.switch_backend('Agg')
+    fig = plt.gcf()
+    dt = [3, 6, 12]
+    order2 = [0.01, 0.04, 0.16]
+    operators = ['tracer1']
```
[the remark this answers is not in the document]
cbegeman: Addressed by https://github.com/MPAS-Dev/compass/pull/452/commits/fb1176e49f5a71e1ecde80c551d21e77100f9d6d

cbegeman:
@xylar Thanks for the suggestions! Ready for your re-review when you have a chance.

xylar (approved):
The updated code looks perfect to me!

PR #453, opened 2022-11-08
Switch to mambaforge and add mamba init

xylar:
This should make a fresh deployment a little faster.


xylar:
I successfully did a fresh deployment on Chrysalis.

PR #457, opened 2022-11-15
Add error message for namelist replacements

cbegeman:
Prior to this commit, if namelist options specified in test cases are out of date with the registry, the user is not warned. This PR adds a warning which is printed at the setup stage.

### DIFF
--- a/compass/namelist.py
+++ b/compass/namelist.py
@@ -51,10 +51,16 @@ def ingest(defaults_filename):
 def replace(namelist, replacements):
     """ Replace entries in the namelist using the replacements dict """
     new = dict(namelist)
+    is_not_replaced = [True for key in replacements.keys()]
     for record in new:
-        for key in replacements:
+        for idx, key in enumerate(replacements):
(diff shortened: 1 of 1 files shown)


sbrus89:
In my opinion, I think this is a great idea.

cbegeman:
@sbrus89 Thanks for the suggestion. I haven't implemented something like that before so I don't know whether it's worth the effort. I'm hoping that this warning will mostly come up when someone is testing a PR that introduces the namelist change, and then they will immediately update the compass namelists. In those cases, the tester would be familiar with the namelist change.

sbrus89:
@cbegeman, very true. My suggestion probably wouldn't have much added value in that case.

xylar (approved):
Looks great!  As long as this has been tested and works, I'll merge!

xylar:
As I discussed with @sbrus89, I think his fuzzy approach sounds interesting.  But I think it might be too much work at least for now.

PR #459, opened 2022-11-16
Update ocean test case namelists

cbegeman:
This PR resolves discrepancies between ocean test case namelists and the current ocean Registry. These discrepancies were identified using the new error message from https://github.com/MPAS-Dev/compass/pull/457.


cbegeman:
These changes are BFB with master using the PR test suite (chrys, intel, impi).

PR #460, opened 2022-11-18
Add SO12to60E2r5 (without ice-shelf cavities)

xylar:
This merge also cleans up a doc string and switches the CMIP mesh for both SO12to60 and WC14 to the 1 degree, consistent with conversations with colleagues.

### DIFF
--- a/compass/mesh/mesh.cfg
+++ b/compass/mesh/mesh.cfg
@@ -17,7 +17,7 @@ mpas_mesh_filename = base_mesh.nc
 plot_cell_width = True
 cell_width_filename = cellWidthVsLatLon.nc
 cell_width_image_filename = cellWidthGlobal.png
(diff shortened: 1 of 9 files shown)


xylar:
## E3SM initial condition

Results from generating an E3SM initial condition on Chrysalis are here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20221118/so12to60/ocean/global_ocean/SO12to60
```
The test cases are still running and I'll update here when they finish.

xylar:
## Plots of mesh resolution

cellWidthGlobal

so12to60_north

so12to60_south

xylar:
@maltrud, please let me know how this goes.  I will merge the PR as soon as I hear from you that this mesh is doing what you need.

xylar:
Thanks @maltrud!

PR #449, opened 2022-11-01
Add thin film cases to ISOMIP+ Ocean0

cbegeman:
This PR adds a set of test cases building on ISOMIP+ Ocean0. The domain is extended to include a thin-film region for the purposes of testing wetting-and-drying algorithms. These test cases include:
- Thin film cases in all supported vertical coordinate types with no external forcing
- Thin film cases in all supported vertical coordinate types with time-varying forcing `thin_film_time_varying_Ocean0`
- Thin film case in a single-layer configuration with tidal forcing `single_layer/thin_film_tidal_forcing_Ocean0`
- Thin film case in single-layer or sigma coordinates with increasing or decreasing land ice pressure (performance step extended to 1 day) `thin_film_drying_Ocean0`, `thin_film_wetting_Ocean0`

There are also some additions to visualizations for both `initial_state`, `performance` and `simulation` output.

PR #467, opened 2022-12-03
Clean up ISOMIP+

xylar:
This is just a clean-up PR to take care of a few things that have been missed over time in the ISOMIP+ test group.
* Rename `OceanTest` to `IsomipPlusTest` and move it from its own package directory to a module
* Some PEP8 issues
* Some parameters and attributes missing from docstrings
* Some plots that look weird for "normal" (z-star, no thin film) test cases.

### DIFF
--- a/compass/ocean/tests/isomip_plus/initial_state.py
+++ b/compass/ocean/tests/isomip_plus/initial_state.py
@@ -168,7 +168,8 @@ class InitialState(Step):
 
         section = config['isomip_plus']
 
-        # Deepen the bottom depth to maintain the minimum water-column thickness
+        # Deepen the bottom depth to maintain the minimum water-column
+        # thickness
         min_column_thickness = section.getfloat('min_column_thickness')
         min_layer_thickness = section.getfloat('min_layer_thickness')
         min_levels = section.getint('minimum_levels')
@@ -232,10 +233,12 @@ class InitialState(Step):
                                sectionY=section_y, dsMesh=ds, ds=ds,
                                showProgress=show_progress)
 
-        ds['landIcePressure'] = ds['landIcePressure'].expand_dims(dim='Time', axis=0)
+        ds['landIcePressure'] = \
+            ds['landIcePressure'].expand_dims(dim='Time', axis=0)
         ds['bottomDepth'] = ds['bottomDepth'].expand_dims(dim='Time', axis=0)
         ds['totalColThickness'] = ds['ssh']
-        ds['totalColThickness'].values = numpy.sum(ds['layerThickness'].values, axis=2)
+        ds['totalColThickness'].values = \
+            ds['layerThickness'].sum(dim='nVertLevels')
         plotter.plot_horiz_series(ds.landIcePressure,
                                   'landIcePressure', 'landIcePressure',
                                   True)
@@ -247,9 +250,9 @@ class InitialState(Step):
                                   vmin=min_column_thickness+1e-10, vmax=700,
                                   cmap_set_under='r', cmap_scale='log')
         plotter.plot_horiz_series(ds.totalColThickness,
-                                  'totalColThickness', 'totalColThickness', True,
-                                  vmin=min_column_thickness+1e-10, vmax=700,
-                                  cmap_set_under='r')
+                                  'totalColThickness', 'totalColThickness',
+                                  True, vmin=min_column_thickness+1e-10,
+                                  vmax=700, cmap_set_under='r')
         plotter.plot_layer_interfaces()
 
         plotter.plot_3d_field_top_bot_section(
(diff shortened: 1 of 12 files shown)


xylar:
## Testing

I ran all ISOMIP+ test cases on Chrysalis with Intel and Intel-MPI using the E3SM branch https://github.com/cbegeman/E3SM/commits/ocn/add-wetting-drying-ramp-feature.

I'm seeing failures in both the baseline using `master` and this branch for:
* `ocean/isomip_plus/2km/single_layer/thin_film_drying_Ocean0`
* `ocean/isomip_plus/2km/single_layer/thin_film_wetting_Ocean0`
* `ocean/isomip_plus/5km/single_layer/thin_film_drying_Ocean0`
* `ocean/isomip_plus/5km/single_layer/thin_film_wetting_Ocean0`
I didn't look at all these runs but the first is failing (with both `master` and this branch) with state validation errors during `performance`.

Other than that, all test cases pass and are bit-for-bit between this branch and `master`.

xylar:
@cbegeman, I was testing `isomip_plus` tests as you can see.  When I tried setting things up using the current `E3SM-Project` submodule, i got a bunch of warnings about config options that were not found.  I think that might mean I didn't do proper testing of #449 (before these warnings were added).  It seems like some of these test cases depend on features that aren't on E3SM `master` yet.  That's something we try to avoid but it's not a huge problem.

It there a branch that should work for all of the `isomip_plus` test cases?

Regardless of that, could you have a quick look at this clean up and let me know if you're good with it?

compass/ocean/tests/isomip_plus/initial_state.py
```diff
@@ -234,10 +233,12 @@ def run(self):
...
                                showProgress=show_progress)
 
-        ds['landIcePressure'] = ds['landIcePressure'].expand_dims(dim='Time', axis=0)
+        ds['landIcePressure'] = \
+            ds['landIcePressure'].expand_dims(dim='Time', axis=0)
         ds['bottomDepth'] = ds['bottomDepth'].expand_dims(dim='Time', axis=0)
         ds['totalColThickness'] = ds['ssh']
-        ds['totalColThickness'].values = numpy.sum(ds['layerThickness'].values, axis=2)
+        ds['totalColThickness'].values = \
+            ds['layerThickness'].sum(dim='nVertLevels')
```
xylar, line 241: The `numpy.sum()` function doesn't handle NaNs well, whereas the xarray sum does (and it has the added advantage of summing over a named dimension rather than a numbered axis).  My runs of the "generic" Ocean0 test case were showing masked-out column thickness wherever there were NaNs in `layerThickness` (when bathymetry was above the top of the bottom layer).

compass/ocean/tests/isomip_plus/viz/plot.py
```diff
@@ -677,13 +686,15 @@ def plot_layer_interfaces(self, figsize=(9, 5)):
             plt.plot(1e-3 * X[0, :], self.zBotSection, 'g')
 
             ax.autoscale(tight=True)
+            x1, x2, y1, y2 = 420, 470, -650, -520
+            xlim = [min(x1, 1e-3*numpy.amin(X)), 1e-3*numpy.amax(X)]
+            plt.xlim(xlim)
```
xylar, line 691: The inset was weird if the full x axis didn't include x = 420 km (i.e. for non-thin-film cases).  So I'm making sure it includes the full inset box.

cbegeman:
@xylar Thanks for this cleanup! 

We should remove the line that turns the ramp feature on from the namelist, as in this commit cbdab2589b54ac1d5b29ced6bcf161c8df5eb987. 

I ran these test cases:

- ocean/isomip_plus/5km/single_layer/thin_film_drying_Ocean0
- ocean/isomip_plus/5km/single_layer/thin_film_wetting_Ocean0

Using the last master commit on the `ocn/add-wetting-drying-ramp-feature` branch, 4dcc8dba15, and a more recent commit, 19711c38bb, I don't reproduce the failures. I also used chrys, intel-mpi so I'm not sure what the difference between our configurations is.

I am reproducing the failures of the `ocn/add-wetting-drying-ramp-feature` branch when the ramp is on but not when it is off. This supports turning the ramp feature off until we figure out what is going on there. I haven't been using those tests lately, only the sigma versions of those tests.

xylar:
@cbegeman, thanks for looking at this.  I agree, cbdab25 is an easy way around the warning I was seeing.  Let's include that here unless you have it already included somewhere else.

I was testing with a merge of `E3SM/master` into `ocn/add-wetting-drying-ramp-feature`, so maybe there's some change there compared with the head of `ocn/add-wetting-drying-ramp-feature` that explains the error I ran into.  Let's not worry about it if there's an easy way to make the current tests compatible with the current `E3SM/master`.

cbegeman:
@xylar Sounds good. You can go ahead and cherry-pick that commit.

I successfully ran the `wetdry` suite with this compass branch, so it's ready to merge after including that commit unless you want me to do further testing.

cbegeman (approved):

xylar:
Thanks, @cbegeman!

PR #461, opened 2022-11-19
Update E3SM-Project submodule

xylar:
This merge updates the E3SM-Project submodule from [6b81271377](https://github.com/E3SM-Project/E3SM/tree/6b81271377) to [569ed6b730](https://github.com/E3SM-Project/E3SM/tree/569ed6b730).

This update includes the following MPAS-Ocean and MPAS-Frameworks PRs (check mark indicates bit-for-bit with previous PR in the list):
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5027
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5128
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5142
- [x]  (fwk) https://github.com/E3SM-Project/E3SM/pull/5120
- [ ]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5099
  - comparison with 5120 fails for 17 test cases:
    - all forward runs except with the z-level ice-shelf test case and those with RK4
    - This was documented in https://github.com/E3SM-Project/E3SM/pull/5099#issuecomment-1226972097
- [ ]  (ocn) https://github.com/E3SM-Project/E3SM/pull/4945
  - comparison with 5099 fails for 4 tests with RK4, as expected
- [x]  (fwk) https://github.com/E3SM-Project/E3SM/pull/5152
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5162
- [ ]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5167
  - comparison with 5162 fails for 17 test cases:
    - all forward runs except with the z-level ice-shelf test case and those with RK4
    - This was documented in https://github.com/E3SM-Project/E3SM/pull/5167#issuecomment-1234445151
- [x]  (fwk) https://github.com/E3SM-Project/E3SM/pull/5174
- [x]  (fwk) https://github.com/E3SM-Project/E3SM/pull/5154
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5170
- [x]  (fwk) https://github.com/E3SM-Project/E3SM/pull/5183
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5151
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5186
- [ ]  (ocn) https://github.com/E3SM-Project/E3SM/pull/5171
  - comparison with 5186 fails for 13 test cases:
[...]

PR #468, opened 2022-12-03
Switch Chrysalis Intel to use OpenMPI by default

xylar:
This is what E3SM uses and we have had a lot of bit-for-bit failures with Intel-MPI (see #461) that might be better with OpenMPI.


xylar:
## Testing

I verified that I get a load script for Intel and OpenMPI on Chrysalis now when I don't specify the compiler or MPI library, as expected.

PR #469, opened 2022-12-05
Reorganize ISOMIP+ steps

xylar:
This merge reorganizes the ISOMIP+ test group to break the `initial_state` step into:
* `process_geom`: process the BISICLES topography data for the requested resolution
* `planar_mesh`: create a planar mesh of the requested resolution and size
* `cull_mesh`: cull out the grounded ice and land from the mesh
* `initial_state`: put the initial condition on the culled mesh

Some functionality has been taken out of the `geom` module and put into `process_geom` for simplicity. 

The `process_geom` step has been updated to use PEP8 conventions and to use `xarray` instead of `netcdf4`.  The code has been updated in other ways as well to shorten the code, increase code reuse and improve readability.

The remaining content in `geom` is shared between `cull_mesh` and `initial_state`, and has been cleaned up to comply with PEP8 (e.g. lowercase variables with underscores).

I also removed interface plotting (from the central section) on top of the overturning streamfunction, which we don't want because these interfaces don't apply.

Finally, I added a small fix to make sure horizontal plots of potential density and salinity respect the given `vmin` for the colormap.  It seems we need to set `clim` after any other alterations to the patches object or matplotlib will (seemingly at random) make `vmin` include 0.

This reorganization is in anticipation of supporting a `spherical_mesh` as an alternative to `planar_mesh`.

PR #470, opened 2022-12-05
Fix `landIceFraction` in ISOMIP+

xylar:
We need `landIceFraciton` to be zero when `landIceMask` is zero.

Following up on #469, but this is a bug and is non-BFB.

### DIFF
--- a/compass/ocean/tests/isomip_plus/initial_state.py
+++ b/compass/ocean/tests/isomip_plus/initial_state.py
@@ -102,11 +102,17 @@ class InitialState(Step):
 
         ds = interpolate_geom(ds_mesh, ds_geom, min_ocean_fraction, thin_film_present)
 
-        for var in ['landIceFraction']:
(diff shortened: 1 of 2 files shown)


xylar:
I was seeing strange plots like these:
thermalDriving_0001
halineDriving_0001
I strongly believe the artifacts at the edges are places where `landIceMask` is zero but `landIceFraction > 0`.  I don't think we want this to happen.

cbegeman (approved):
@xylar I successfully ran up through the performance step but am waiting in the queue to get through simulation, streamfunction and viz. Did these code changes end up removing the artifacts you mentioned?

xylar:
@cbegeman, yes, it does.  I forgot to post results.

Thanks for testing!

I'll merge but please post if your remaining resting produces unexpected results.

xylar:
I rebased to take out the commits unrelated to this PR (for posterity).

PR #478, opened 2022-12-14
Fix provenance when E3SM submodule not needed

xylar:
Some test cases don't need an MPAS component to be built in order to run.  This merge makes it so provenance does not attempt to get a git has for the MPAS component if the path to the component doesn't exist (presumably because the submodule has not been initialized).

Checklist
- [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

### DIFF
--- a/compass/provenance.py
+++ b/compass/provenance.py
@@ -105,6 +105,9 @@ def _get_mpas_git_version(config):
 
     mpas_model_path = config.get('paths', 'mpas_model')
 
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I verified that I can now set up a test case (`files_for_e3sm`) that doesn't need MPAS-Ocean without specifying a valid `-p` path.  Before this fix, I got error messages from provenance when I tried to do this.

PR #479, opened 2022-12-14
Add `database_component` parameter for input files

xylar:
This may be useful when accessing files for a different core (e.g. seaice files within a test case in the ocean core).

This merge also changes the way that subdirectories for the local database are specified.  Instead of specifying a `database_root` for each core, a single root is used and the `database_subdirs` for each core are hard-coded in `database_subdirs.json`

This merge also handles the deprecation of `importlib.resources.path()` in python 3.11 by making a `compass.io.package_path()` function with the same functionality.  All calls to `importlib.resources.path()` have been replaced with the new function.  This was included here because the changes introduce to database would have otherwise added yet another call to the deprecated function.

Checklist


xylar:
## Testing

I was able to set up a new ocean test case that includes input files from the `seaice` database using:
``` python
self.add_input_file(filename=filename,
                    target=filename,
                    database='partition',
                    database_component='seaice')
```

xylar:
@hollyhan, @matthewhoffman, @mark-petersen, @trhille, @cbegeman, @sbrus89, @hollyhan pinging you as regular compass users.  Please ping anyone else I missed.

If you run on your laptop/desktop, you will need to modify the config file you use locally after this merge goes in.  You will need `database_root` instead of `ocean_database_root` and/or `landice_database_root`.  This is because I need to add a sea-ice database, too, and I just want to have one config option to handle all 3 databases instead of a proliferation.

xylar:
Given that I really need to build off of this functionality and no one is around to review at the moment, I'm going to chance a disruptive merge.  Sorry in advance, everyone!

PR #481, opened 2022-12-18
Override `constrain_resources()` in individual steps

xylar:
This merge drops calls to `Testcase.run()` and instead override `step.constrain_resources()` in individual steps.

We have deprecated `Testcase.run()` because its use is not compatible with the task parallelism approach we (@altheaden and I) are developing.  Instead, the individual steps in a test case that need to set the number of tasks or cpus per task should override `constrain_resources()`, set `self.ntasks`, `self.min_tasks`, etc., and then call `super().constrian_resources()`.

In the near future (again related to task parallelism), it will also be useful to override `step.runtime_setup()` for tasks like creating graph partitions before running an MPAS component.  In anticipation of this change and to be consistent with the deprecation warning about `Testcase.run()`, this merge adds a call to `runtime_setup()` when a step gets run.  Currently, no steps override this function.

Checklist

### DIFF
--- a/compass/ocean/tests/global_convergence/cosine_bell/__init__.py
+++ b/compass/ocean/tests/global_convergence/cosine_bell/__init__.py
@@ -72,25 +72,6 @@ class CosineBell(TestCase):
 
         self.update_cores()
 
-    def run(self):
-        """
-        Run each step of the testcase
-        """
-        config = self.config
-        for resolution in self.resolutions:
-            if self.icosahedral:
-                mesh_name = f'Icos{resolution}'
-            else:
-                mesh_name = f'QU{resolution}'
-            ntasks = config.getint('cosine_bell', f'{mesh_name}_ntasks')
-            min_tasks = config.getint('cosine_bell', f'{mesh_name}_min_tasks')
-            step = self.steps[f'{mesh_name}_forward']
-            step.ntasks = ntasks
-            step.min_tasks = min_tasks
-
-        # run the step
-        super().run()
-
     def update_cores(self):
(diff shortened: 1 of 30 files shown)


xylar:
## Testing

I ran the `pr` test suite on Chrysalis with Intel and OpenMPI in optimized mode, using the current `master` as a baseline.

I will also test the remaining test groups that aren't part of the `pr` suite or ask others to do this as part of their reviews.

Test groups:
- [ ] GlobalConvergence
- [ ] GlobalOcean
- [ ] Hurricane
- [ ] IsomipPlus
- [ ] PlanarConvergence
- [ ] SphereTransport
- [ ] Tides

xylar:
@vanroekel and @sbrus89, this should address the issue with the Tides test group discussed in https://github.com/MPAS-Dev/compass/pull/456#issuecomment-1349848668

xylar:
@sbrus89 and @vanroekel, I'm not quite done testing Hurricane and Tides because they haven't run to completion but I did verify that I could change the resources for both init and forward at runtime.

I would appreciate having you each test as well, though, just to make sure the issue if fixed.

vanroekel (approved):
I tested the hurricane and tides test case on chrysalis and confirmed a change to ntasks in the cfg file changed the number of cores used in the run.  Thanks @xylar!

xylar:
Seeing #482 but otherwise everything is working.

xylar:
Thanks @vanroekel.

xylar:
@sbrus89, I think @vanroekel's testing is sufficient.  I'm going to merge.

PR #488, opened 2022-12-22
Add Thwaites UQ study framework

matthewhoffman:
This merge creates a framework for setting up and running a basic Thwaites UQ study.  It sets up a user-defined number of simulations with parameter values selected from a Sobol sequence.  It is set up to consider the basal friction exponent and the von Mises stress threshold as the two parameters varied.  There also is plumbing to consider the gamma0, deltaT, and calving speed limit as additional or alternative parameters.

Each ensemble member is a step of the test case.  A slurm job script is created for each job, so that they can be submitted and run in parallel.  There is also an additional step called the ensemble manager that submits each run with sbatch.  Due to the requirement that compass run only occurs on a compute node, this step has to be submitted via a batch script or interactive job.  (This may be addressed in the future.)  There is a visualization script that performs basic analysis on the results of the ensemble.  It will be further developed in the future, but provides a basic assessment of results.  

A config file specifies the location of the input file, the basal friction exponent used in the input file, the name of the parameter vector file to use, and the start and end run numbers to set up.  The ensemble manager can set up and run additional runs with a different range of run numbers after being run initially.  This allows one to perform a small ensemble (e.g. 2-10 runs) to make sure results look as expected before spending time on a larger ensemble.  This also allows one to add more ensemble members from the Sobol sequence later if UQ analysis indicates the original sample size was insufficient.

The design philosophy is to balance generalizability with simplicity.  The test group is set up to fairly easily add other ensemble configurations with minimal effort.  On the other hand, in the specific thwaites test included now, there are assumptions about running on Cori, etc.  These could be generalized in the future if desired.  Because this will not be included in suites, it is intended for expert users and it is expected that it will typically be run with a customized cfg file.

A number of issues remain that may be addressed in future branches:
* numTasks is hard coded for Cori
* input files comes from paths on NERSC specified in the cfg file.  Once a final initial condition is established, this can be moved to the anvil data server.  Similarly, forcing files are hardcoded to paths on NERSC in the streams file.
* Run manager should manage restarting runs on a subsequent call to `compass run` by identifying runs that had not reached the stop time and only resubmitting those jobs (as restarts).  This has not been an issue yet, because the configuration currently specified is short enough to not require restarts.
* ideally the run manager would provide the user with an estimate of how many node-hours they are launching to avoid burning through large amounts of allocation time.

There is a small compass framework change to setup.py that stores the machine name in the config so a test case has access to it.  This was necessary to make use of the framework job script generation script, which requires the machine name as an input.

Checklist


compass/machines/cori-haswell.cfg
```diff
@@ -20,6 +20,9 @@ compass_envs = /global/common/software/e3sm/compass/cori-haswell/base
 # machines
 [deploy]
 
+# the name of the machine
+machine = cori-haswell
```
matthewhoffman, line 24: @xylar , based on our discussion last week, I made changes to this PR to use the standard compass job generation function instead of having a special copy in this test case.  I was able to do that with only one change to compass framework, which is this line letting the config know the machine name being used.  This is because the `write_job_script` function requires `machine` as an input argument, and I could not see anywhere that the machine is saved and made accessible other than through the `compass setup` call stack (which was not accessible from my step's `setup` method).  

I could imagine there are other ways this problem could be resolved, so I want to get your feedback on my adding this line to the machine .cfg file before I add the corresponding lines to the other machine cfg files.  You certainly would not want users to modify this cfg entry, but there are a number of other entries that have the same situation (e.g. compiler).  On the other hand, I could imagine other uses for saving the machine name (e.g. a test case wanting to do things differently depending on which machine it is being run on).

Commit f894a044e4e012e73c98f444c5095c4d3bdf4dea explains why this was needed and shows how it was used to solve my problem.

compass/landice/tests/thwaites/uq_ensemble/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+        """
+        name = 'thwaites_uq'
+        super().__init__(test_group=test_group, name=name)
+
+        # We need the list of runs to set up steps.
+        # Steps need to be defined in TestCase.__init__
+        # But we want to keep operations to a minimum
+
+        self.nRuns = 20 # number of runs in ensemble
```
matthewhoffman, line 33: @xylar , in our discussion, I believe you suggested the number of runs could come from the config file.  But my understanding is the config is not parsed until the `configure`/`setup` step.  Can you explain or point me to an example of how best to get this value from the config here in the constructor?

compass/landice/tests/thwaites/uq_ensemble/ensemble_member.py
```diff
@@ -0,0 +1,119 @@
...
+        self.add_namelist_options(options=options,
+                                  out_name='namelist.landice')
+
+
+        # set job name to run number so it will get set in batch script
+        self.config.set('job', 'job_name', f'uq_{self.name}')
+        machine = self.config.get('deploy', 'machine')
+        write_job_script(self.config, machine,
+                         target_cores=self.ntasks, min_cores=self.min_tasks,
+                         work_dir=self.work_dir)
```
xylar, line 101: Setting `[job]/job_name` might have unintended consequences (like setting the name of the job for the whole suite to the last uq name).  One better option would be to copy the config options to a new config object just to pass to `write_job_script()`, see http://mpas-dev.github.io/MPAS-Tools/stable/generated/mpas_tools.config.MpasConfigParser.copy.html#mpas_tools.config.MpasConfigParser.copy.

But I think a better option would just be to add another keyword argument to `write_job_script(..., job_name=None)`.  Then, we would use the argument if it's not `None` and the current behavior if it's `None`.

compass/machines/cori-haswell.cfg
xylar: @matthewhoffman, I think it might be better to just add the machine name in code during setup.  That way, it doesn't have to be included by hand in each machine's config file, especially because it's redundant.

I would add the following here in the code:
https://github.com/matthewhoffman/compass/blob/7b02c0c924ee77babe975afea4bf48e846360733/compass/setup.py#L434-L437
```python
    # add the compass machine config file
    if machine is None:
        machine = 'default'
    config.add_from_package('compass.machines', f'{machine}.cfg')

    if machine is not None:
        config.set('deploy', 'machine', machine)
```

compass/landice/tests/thwaites/uq_ensemble/__init__.py
xylar: You can (and I think should in this circumstance) pass a user config file with `-f` as part of `compass setup`.  That would be the config file that sets a config option.  This config option could default to 20 (or 1 or whatever makes sense as a default).  Ideally, users would also use the config file to set other properties of the ensemble *before* setting it up.  This would be qualitatively similar to the config file required for the `ismip6_forcing` test cases.

compass/landice/tests/thwaites/uq_ensemble/ensemble_member.py
matthewhoffman: I'll take a look at including a new optional argument to `write_job_script`.  I had started doing that yesterday, but chose this route for some reason I'm not recalling.

compass/machines/cori-haswell.cfg
matthewhoffman: @xylar , yes I agree that sound cleaner.  I'll add that.  Thanks for the feedback.

compass/landice/tests/thwaites/uq_ensemble/__init__.py
matthewhoffman: @xylar , I agree with the idea of having a cfg file where some user adjustments are expected prior to setting up the test case for this. I took a look at the ismip6_forcing test group again, but I can't find an example of where the config is parsed in the constructor.  In order to init the steps for the, say, 20 runs, we need to know the number of runs in the test case constructor, but standard practice is that the config object is not parsed and available until `setup`.  From what I can tell, the ismip6_forcing group follows that standard practice, but I might be missing something.

(BTW, you may have noticed this comment thread only appears in the conversation tab of the PR, because I have subsequently changed this line of code for other reasons.)

xylar: The config file is never parsed in the constructor.  That's not possible because that happens even for compass list.  It must happen in the `setup()` method.  You have to move this stuff from the constructor to `configure()`.

xylar: Take a look at the cosine bell test case for an example of this:
https://github.com/MPAS-Dev/compass/blob/main/compass/ocean/tests/global_convergence/cosine_bell/__init__.py

We start with a default set of steps based on the default value of the config option `resolutions`, but then we throw away those steps at `configure()` and make a new set based on a modified set of resolutions (if necessary).

xylar: In your case, I wouldn't bother with the stuff in `__init__()` because you should be fine with there being zero steps in the test case by default (e.g. in `compss list --verbose`).  You can just add the steps within `configure()`.

matthewhoffman: Ah, I see.  I was so focused on trying to get the steps finalized in the constructor.  I know you keep referring me to the cosine bell test, and now I understand that aspect of it.  I had understood that `steps_to_run` could be changed at setup time, but I hadn't made the connection that the actual set of steps that are part of the test case could be entirely adjusted during setup.  I'll take a closer look and implement that when I have time to get back to this.  Thanks for the explanation.

compass/machines/cori-haswell.cfg
matthewhoffman: Changed in c1069fe9b14fcc5fc4d892c08687a0eaa8d39bf7

compass/landice/tests/thwaites/uq_ensemble/__init__.py
matthewhoffman: Changed in 974353cb81252add490a5ede2c939eb878cf7567

compass/landice/tests/thwaites/uq_ensemble/ensemble_member.py
matthewhoffman: Changed in f894a044e4e012e73c98f444c5095c4d3bdf4dea

matthewhoffman:
# Testing

Tested with `compass setup -n 34 -w /global/cscratch1/sd/hoffman2/COMPASS/ens_reorg`.  By default, it will set up 4 ensemble members.  I confirmed they all get submitted properly with `compass run`, using the following hack to avoid needing to execute from a compute node:
```
diff --git a/compass/parallel.py b/compass/parallel.py
index 0414a4ce..86ce638a 100644
--- a/compass/parallel.py
+++ b/compass/parallel.py
@@ -25,6 +25,7 @@ def get_available_cores_and_nodes(config):
     """

     parallel_system = config.get('parallel', 'system')
+    parallel_system = 'single_node'
     if parallel_system == 'slurm':
         job_id = os.environ['SLURM_JOB_ID']
         node = os.environ['SLURMD_NODENAME']
@@ -76,8 +77,9 @@ def check_parallel_system(config):
     parallel_system = config.get('parallel', 'system')
     if parallel_system == 'slurm':
         if 'SLURM_JOB_ID' not in os.environ:
-            raise ValueError('SLURM_JOB_ID not defined.  You are likely not '
-                             'on a compute node.')
+            pass
+            #raise ValueError('SLURM_JOB_ID not defined.  You are likely not '
+            #                 'on a compute node.')
     elif parallel_system == 'single_node':
         pass
     else:
```

Testing need to be done on Cori to have access to the input file specified in the cfg file and the forcing files hardcoded in the streams files.

This PR is meant to merge the basic ensemble capability, and more PRs are expected in the future to build out details.  Thus, I consider simple ensuring a user can set up and submit an ensemble is sufficient testing for this branch.  This test group/test case will not be added to an integration suite.

I will be writing up documentation next before requesting review.

matthewhoffman:
@xylar and @trhille , this is ready for review.  As I've mentioned to both of you, the goal for this PR is to merge a basic ensemble functionality that can be checkpointed and then built out further as the project progresses.  

@xylar , I think a review by inspection, plus more critically looking at the (small) framework and docs changes should suffice from you, but you are welcome to review and/or run at whatever level of detail you wish.  Your initial review helped a lot with the design and structure.

@trhille , for you, I suggest reading the docs and then trying to set up and submit a small ensemble.  (By default it only sets up 4 runs, but you can also kill them after they're submitted if you don't actually want to run them.)  You can look at the output path for https://acme-climate.atlassian.net/wiki/spaces/FAN/pages/3656220835/Thwaites+UQ+test+4 if you want to browse/plot a larger ensemble.  You may need to touch base with me on how to customize a cfg for a test, but I think the default will more or less work fine for testing purposes.

Note that the history is fairly convoluted, so it's probably easier to look at the final file changes than follow along the history.  I originally created this as a test case within the thwaites test group, but at the end decided it should be its own test group, with the thwaites application as its first test case.  It would be extensible to other applications and is in general not unique to the thwaites configuration.

You can view the docs changes here, where I've moved temporarily for review:
https://portal.nersc.gov/cfs/fanssie/hoffman2/html/index.html, specifically:
* https://portal.nersc.gov/cfs/fanssie/hoffman2/html/users_guide/landice/test_groups/ensemble_generator.html
* https://portal.nersc.gov/cfs/fanssie/hoffman2/html/developers_guide/landice/test_groups/ensemble_generator.html
* https://portal.nersc.gov/cfs/fanssie/hoffman2/html/developers_guide/landice/api.html
(Note: bullets are not rendering for some reason, but I cannot figure out why.  The issue occurs on a number of pages besides the ones I added, so I don't think it's specifically related to the changes I've made.)

compass/setup.py
```diff
@@ -436,6 +436,10 @@ def _get_basic_config(config_file, machine, mpas_model_path, mpas_core):
         machine = 'default'
     config.add_from_package('compass.machines', f'{machine}.cfg')
 
+    # store name of machine so tests have access to it if needed
+    if machine is not None:
+        config.set('deploy', 'machine', machine)
+
```
xylar, line 442: I think this is great!  There is also a `$COMPASS_MACHINE` environment variable set in the load script, so that would have been another way that maybe I didn't think of at the time.

docs/developers_guide/building_docs.rst
```diff
@@ -17,4 +17,8 @@ Then, run the following script to build the docs:
     rm -rf developers_guide/generated/ developers_guide/*/generated/ _build/
     make html
 
+You may need to rerun your compass load script for the api docs to build
+successfully if you have added new modules since the load script was
+last run.
+
```
xylar, line 23: An important addition.  Let me offer a couple of edits:
```suggestion
You may need to re-source your compass load script in the root of the compass branch
for the api docs to build successfully if you have added new modules since the load
script was last sourced.  The load script will reinstall compass into the conda
environment when it is sourced in the root of the compass branch.

```

compass/landice/tests/thwaites/__init__.py
```diff
@@ -3,7 +3,6 @@
 from compass.landice.tests.thwaites.restart_test import RestartTest
 from compass.landice.tests.thwaites.high_res_mesh import HighResMesh
 
-
```
xylar, line 6: This extra newline is standard in PEP8.  Please put it back when you get a chance.

compass/landice/tests/ensemble_generator/ensemble_generator.cfg
```diff
@@ -0,0 +1,21 @@
...
+# Run numbers should be zero-based
+# These values do not affect viz/analysis, which will include any
+# runs it finds.
+start_run = 0
+end_run = 3
+
+param_vector_filename = Sobol_Initializations_seed_4_samples_100.csv
+
+input_file_path = /global/cfs/cdirs/fanssie/MALI_projects/Thwaites_UQ/Thwaites_4to20km_r02_20230126/Thwaites_4to20km_r02_20230126.nc
+# Base path to the input files. User has to supply.  This path needs to include the files listed below.  Eventually this could be hard-coded to use files on the input data server, but initially we want flexibility to experiment with different inputs and forcings
```
xylar, line 17: This comment should be above the associated config option for it to translate correctly into the config file in the setup directory.  This comment will get attached to `orig_fric_exp` where it is.

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -0,0 +1,217 @@
...
+rhoi = 910.0
+rhosw = 1028.0
+
+runs = sorted(glob.glob("run*"))
+nRuns = len(runs)
+
+
+# obs from Rignot et al 2019 PNAS
+obs_discharge_thwaites = np.array([82.42,	83.25,	84.08,	84.90,	85.73,	86.56,	87.39,	88.22,	89.04,	89.87,	90.70,	91.40,	92.10,	92.80,	93.50,	94.20,	94.90,	95.60,	95.90,	96.20,	96.50,	96.80,	97.10,	97.40,	97.75,	98.10,	101.60,	101.40,	102.80,	104.30,	109.70,	114.10,	115.80,	118.70,	121.60,	123.80,	117.20,	115.10,	119.68])
+obs_discharge_haynes = np.array([10.10, 10.25, 10.40, 10.55, 10.70, 10.85, 11.00, 11.15, 11.30, 11.45, 11.60, 11.49, 11.37, 11.26, 11.14, 11.03, 10.91, 10.80, 10.93, 11.07, 11.20, 11.33, 11.47, 11.60, 11.35, 11.10, 12.20, 11.90, 12.50, 12.11, 12.80, 12.80, 12.90, 12.80, 13.90, 14.70, 12.30, 11.60, 12.53])
```
xylar, line 24: Please reformat these to stick with the 79 character line length and with just one space between entries. Maybe consider having them in a data file and loading them instead of having them hard coded here.

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -0,0 +1,217 @@
...
+
+#axfltArea = fig.add_subplot(nrow, ncol, 4)
+#plt.title(f'Floating area change at year {targetYear} (km$^2$)')
+#plt.xlabel('von Mises stress threshold (kPa)')
+#plt.ylabel('basal friction law exponent')
+#plt.grid()
+#plt.scatter(vmThresh, fricExp, s=markerSize, c=fltAreaChangeAll, plotnonfinite=False)
+#badIdx = np.nonzero(np.isnan(fltAreaChangeAll))[0]
+#plt.plot(vmThresh[badIdx], fricExp[badIdx], 'kx')
+#plt.colorbar()
```
xylar, line 203: I will leave this up to you all but I prefer not to have commented-out code unless there is a high likelihood that it will be added back soon.  It would be better to have a config option to plot this, and have it set to `False` by default.

xylar (approved):
A few very minor comments but this looks good to me as long as it works in @matthewhoffman and @trhille's testing.

compass/landice/tests/thwaites/__init__.py
matthewhoffman: I've added it back in.

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: I've deleted it.

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: I've fixed for formatting, but left it hardcoded for now.  This script is run as a symlink from various places, so having this small amount of data hardcoded avoids issues of pointing to a data file.

xylar: Hmm, if you can symlink the script, can't you symlink a data file in the same way? But I'm okay with you doing this your way.

xylar:
@matthewhoffman, as you'll notice here, I added a code linter `pre-commit` that lints the code before you commit changes and also here on GitHub.  I can walk you through it if you have time at some point.

If you rerun `./conda/configure...` after your rebase, and then run:
```
pre-commit run --show-diff-on-failure --color=always --files compass/landice/__init__.py compass/landice/extrapolate.py compass/landice/tests/ensemble_generator/Sobol_Initializations_seed_4_samples_100.csv compass/landice/tests/ensemble_generator/__init__.py compass/landice/tests/ensemble_generator/ensemble_generator.cfg compass/landice/tests/ensemble_generator/ensemble_manager.py compass/landice/tests/ensemble_generator/ensemble_member.py compass/landice/tests/ensemble_generator/plot_ensemble.py compass/landice/tests/ensemble_generator/thwaites/__init__.py compass/landice/tests/ensemble_generator/thwaites/albany_input.yaml compass/landice/tests/ensemble_generator/thwaites/namelist.landice compass/landice/tests/ensemble_generator/thwaites/streams.landice compass/setup.py docs/developers_guide/building_docs.rst docs/developers_guide/landice/api.rst docs/developers_guide/landice/framework.rst docs/developers_guide/landice/test_groups/ensemble_generator.rst docs/developers_guide/landice/test_groups/index.rst docs/users_guide/landice/test_groups/ensemble_generator.rst docs/users_guide/landice/test_groups/index.rst
```
(See https://github.com/MPAS-Dev/compass/actions/runs/4245474450/jobs/7381031296#step:6:83)
this should tell you what it's either fixing automatically for you (e.g. sorting imports) or what it would like you to fix.  Run that command once, then do:
```
git commit compass docs
```
It'll presumably fix a few more things, so run it again.  When it stops fixing things and start complaining about code not following PEP8 or having incorrect typing, that's when you have to actually start fixing the code.

It's going to be nasty and tedious this time because a lot of code is affected at once.  Once we have this going all the time, it'll get easier since only a small amount of code will be involved.  Polaris has this going on all the code committed there so far and everything that moves there will need to have this clean up so it might as well start in compass now.

xylar:
@matthewhoffman, before I forget about it, please add `csv` to this list:
https://github.com/MPAS-Dev/compass/blob/main/setup.py#L50-L51
That way, the data file you added here will be included in the `compass` conda package whenever that gets built and installed.  It doesn't matter if you're using the GitHub repo (so for everyone right now) but it can be an unpleasant surprise to have missing data files when someone is using the conda package.

matthewhoffman:
Updated docs have been generated for perusal here:
https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-thwaites-uq/html/index.html
The main things to review are:
* User's guide test case description: https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-thwaites-uq/html/users_guide/landice/test_groups/ensemble_generator.html
* Developer's guide test case description: https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-thwaites-uq/html/developers_guide/landice/test_groups/ensemble_generator.html

matthewhoffman:
@trhille , I've completed cleanup on this PR, including adding to the docs and doing all the PEP8 delinting.  See links above for relevant parts of compiled documentation.    It's ready for you to test and review whenever NERSC is sufficiently operational again.

@xylar , I took care of the items you identified yesterday.  You are welcome to review again if you'd like, but I don't think that's necessary.

docs/developers_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,127 @@
...
+
+* calving speed limit
+
+* gamma0 melt sensitivity parameter in ISMIP6-AIS ice-shelf basal melting
+  parameterization
+
+* deltaT thermal forcing bias adjustment  parameter in ISMIP6-AIS ice-shel
+  basal melting parameterization
+
+These parameter values are always set; if it desired they not be modified
```
trhille, line 49:
```suggestion
These parameter values are always set; if it is desired they not be modified
```

docs/developers_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,127 @@
...
+from the baseline configuration, a default value should be passed to the
+constructor.
```
trhille, line 51: Which file does the user modify to pass a default value to the constructor?

docs/developers_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,127 @@
...
+in the input file to maintain an unchanged basal shear stress.  Similarly,
+there is a function ``_adjust_basal_melt_params`` that modifes gamma0 and
+deltaT in a basal melt parameter file.
+
+ensemble_manager
+~~~~~~~~~~~~~~~~
+The class :py:class:`compass.landice.tests.ensemble_generator.EnsembleManager`
+defines a step for managing the entire ensemble.  The constructor and setup
+methods perform minimal operations.  The ``run`` method submits each run in
+the ensemble as a slurm job.  Eventually the ensemble_manager will be able
```
trhille, line 87:
```suggestion
the ensemble as a slurm job.  Eventually the ``ensemble_manager`` will be able
```

docs/users_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,220 @@
...
+simulations with different parameter values.  The ensemble framework
+sets up a user-defined number of simulations with parameter values selected
+from a space-filling Sobol sequence.
+
+A test case in this test group consists of a number of ensemble members,
+and one ensemble manager.
+Each ensemble member is a step of the test case, and can be run separately
+or as part of the complete ensemble.  Ensemble members are identified by a
+three digit run number, starting with 000.
+A config file specifies the the run numbers to set up, as well as some common
```
trhille, line 16:
```suggestion
A config file specifies the run numbers to set up, as well as some common
```

docs/users_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,220 @@
...
+information about the run configuration.
+
+The test case can be generated multiple times to set up and run additional
+runs with a different range of run numbers after being run initially. This
+allows one to perform a small ensemble (e.g. 2-10 runs) to make sure results
+look as expected before spending time on a larger ensemble. This also allows
+one to add more ensemble members from the Sobol sequence later if UQ analysis
+indicates the original sample size was insufficient.
+
+Individual test cases will define which parameter are being sampled and
```
trhille, line 26:
```suggestion
Individual test cases will define which parameters are being sampled and
```

docs/users_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,220 @@
...
+over what range.  Currently these parameters are supported:
```
trhille, line 27:
```suggestion
over what ranges.  Currently these parameters are supported:
```

docs/users_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,220 @@
...
+* basal friction power law exponent
+
+* von Mises threshold stress for calving
+
+* calving rate speed limit
+
+* gamma0 melt sensitivity parameter in ISMIP6-AIS ice-shelf basal melting
+  parameterization
+
+* deltaT thermal forcing bias adjustment parameter in ISMIP6-AIS ice-shel
```
trhille, line 38:
```suggestion
* deltaT thermal forcing bias adjustment parameter in ISMIP6-AIS ice-shelf
```

docs/users_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,220 @@
...
+  basal melting parameterization
+
+Additional parameters can be easily added in the future.
+The test group currently includes a file of unit parameter values for two
+parameters with 100 samples using a Sobol sequence.  The parameter
+dimensionality or sample size can be increased by modifying this file and
+its usage.  It also would be possible to modify the sampling strategy to
+perform uniform parameter sensitivity tests.
+
+``compass setup`` will set up the set up simulations and the ensemble manager.
```
trhille, line 48:
```suggestion
``compass setup`` will set up the simulations and the ensemble manager.
```

docs/users_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,220 @@
...
+* von Mises calving
+
+* ISMIP6 surface mass balance and sub-ice-shelf melting using climatological
+  mean forcing
+
+Steps for setting up and running a Thwaites ensmble
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+1. With a compass conda environment set up, run
+   ``compass setup -n X -w WORK_DIR_PATH -f USER.cfg``
```
trhille, line 184:
```suggestion
   ``compass setup -t landice/ensemble_generator/thwaites_ensemble -w WORK_DIR_PATH -f USER.cfg``
```

trhille: Or specify that X is the number shown by `compass list -t landice`

docs/users_guide/landice/test_groups/ensemble_generator.rst
```diff
@@ -0,0 +1,220 @@
...
+   subdirectory and execute ``compass run``.  Note, as stated above, this
+   currently will fail on a login node and has to be performed from a
+   interactive job or batch script.  This will be addressed in the future.
+
+4. Each run will have its own batch job that can be monitored with ``squeue``
+   or similar commands.
+
+5. When the ensemble has completed, you can assess the result through the
+   basic visualization script ``plot_ensemble.py``.  The script will skip runs
+   that are incomplete or failed, so you can run it while an emsemble is
```
trhille, line 209:
```suggestion
   that are incomplete or failed, so you can run it while an ensemble is
```

trhille (approved):
@matthewhoffman , this is excellent! Thanks for creating this useful new tool. I tested on Perlmutter, and was able to easily create and submit a Thwaites ensemble.

I added some very minor comments on the documentation (mostly fixing typos), but I'm approving so this can be merged as soon as possible.

I also visually checked the diagnostic solve for the first four runs to make sure the recalculated basal friction law exponent is producing a good velocity field:
image

and on a log scale
image

image

matthewhoffman:
@trhille , thanks for testing this out, checking the results, and the feedback on the docs.  I'm excited to have this merged and ready to build off of.

PR #490, opened 2022-12-30
Update shapely, geometric_features and mpas_tools

xylar:
New versions are needed to support shapely >=2.0

The only code in compass itself that uses shapely is for generating MISOMIP output from the `isomip_plus` test cases.  As this is rarely used, a considerable amount of work to test, and the calls did not look likely to be affected by the change in shapely, I will not test that part of the code here.

Checklist

### DIFF
--- a/conda/compass_env/spec-file.template
+++ b/conda/compass_env/spec-file.template
@@ -8,7 +8,7 @@ cartopy_offlinedata
 cmocean
 esmf=*={{ mpi_prefix }}_*
 ffmpeg
-geometric_features=0.7.0
+geometric_features=1.0.1
 git
 ipython
 jigsaw=0.9.14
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I ran the `pr` test suite on Chrysalis with Intel and OpenMPI, using the current `main` branch as a baseline.  I used local builds of the new `geometric_features 1.0.0` and `mpas_tools 0.16.0` as those have not yet been released.  To my pleasant surprise, all results are bit-for-bit.  I thought some of the mask creation in MPAS-Tools might have changed with the new shapely.

Results are in:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20221230/pr_shapely_2.0
```

PR #508, opened 2023-01-18
Fix handling of ntasks = 1 for single node

### LINKED ISSUE #507: Issue running mpirun on one proc for a single_node system
When `parallel_system == single_node` and when `ntasks == 1`, the jobs are launched as, e.g.
`mpirun   ./landice_model`
instead of 
`mpirun -n 1 ./landice_model`

Unfortunately the former line executes the job using all the available slots, instead of just one rank  (at least on my machine).

@matthewhoffman pointed out that this is due to the following conditional
https://github.com/MPAS-Dev/compass/blob/main/compass/parallel.py#L154

@xylar Would it be possible to remove that line? That fixed the issue on my machine.

xylar:
Omitting `-n 1` (as was done previously) means the step runs on all cores, not just 1.

Checklist

closes #507

### DIFF
--- a/compass/parallel.py
+++ b/compass/parallel.py
@@ -151,8 +151,7 @@ def run_command(args, cpus_per_task, ntasks, openmp_threads, config, logger):
     if parallel_system == 'slurm':
         command_line_args.extend(['-c', f'{cpus_per_task}', '-n', f'{ntasks}'])
     elif parallel_system == 'single_node':
-        if ntasks > 1:
-            command_line_args.extend(['-n', f'{ntasks}'])
+        command_line_args.extend(['-n', f'{ntasks}'])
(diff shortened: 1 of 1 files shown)


mperego (approved):
Looks good to me. Thanks!

xylar:
@mperego, I haven't been able to test this myself because I'm not able to get MPAS-Ocean to run on linux right now at all (unrelated to this issue).  If you are able to test and this works for you, please report back and I'll merge.  I don't want to merge without any testing at all.

mperego:
@xylar I tested it and it works for me.

xylar:
Okay, great! I'll merge and test it myself when I can run on linux again.

PR #511, opened 2023-01-24
Fix deployment using python 2

xylar:
Checklist

### DIFF
--- a/conda/shared.py
+++ b/conda/shared.py
@@ -187,7 +187,7 @@ def backup_bashrc():
     files = ['.bashrc', '.bash_profile']
     for filename in files:
         src = os.path.join(home_dir, filename)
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

Without this fix, @vankova and I got:
```
ile “./conda/configure_compass_env.py”, line 19, in <module>
  from shared import parse_args, get_conda_base, check_call, install_miniconda, \
 File “/gpfs/fs1/home/ac.vankova/compass/compass/conda/shared.py”, line 190
  dst = os.path.join(home_dir, f’{filename}.conda_bak’)
                            ^
SyntaxError: invalid syntax
```
on machines where the default python is still python 2.  With this fix (not using f-strings), the error is gone on Anvil.

PR #501, opened 2023-01-12
Port overflow test group

cbegeman:
Here we port the overflow test group from legacy compass. It features a density current flowing down a continental shelf. We include a default test at low resolution and a RPE test at higher resolution. The computation time for the RPE test was reduced from legacy by changing the resolution from 1km to 2km, the number of vertical levels from 100 to 50, and reducing the domain length from 240km to 200km. Only the sigma vertical coordinate is correctly initialized in MPAS-Ocean's init mode, so only that coordinate is tested here.

Checklist
* [X] User's Guide has been updated
* [X] Developer's Guide has been updated
* [X] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [ ] New tests have been added to a test suite

PR #516, opened 2023-01-30
Cleanup ocean RPE tests

cbegeman:
This PR includes clean-up of the RPE test cases in the internal wave and baroclinic test groups. It fixes formatted strings and moves the list of viscosities to be tested to the config file.

Checklist

### DIFF
--- a/compass/ocean/tests/baroclinic_channel/baroclinic_channel.cfg
+++ b/compass/ocean/tests/baroclinic_channel/baroclinic_channel.cfg
@@ -26,6 +26,9 @@ min_pc_fraction = 0.1
 # or fractions. False means fractions.
 use_distances = False
 
+# Viscosity values to test for rpe test case
+viscosities = 1, 5, 10, 20, 200
+
 # Temperature of the surface in the northern half of the domain.
 surface_temperature = 13.1
 
--- a/compass/ocean/rpe.py
+++ b/compass/ocean/rpe.py
@@ -9,10 +9,10 @@ def compute_rpe(initial_state_file_name='initial_state.nc',
     Computes the reference (resting) potential energy for the whole domain
     Parameters
     ----------
-    initial_state_file_name : string
+    initial_state_file_name : str
         Name of the netCDF file containing the initial state
 
-    output_file_prefix : string
+    output_file_prefix : str
         Prefix for the netCDF file containing output of forward step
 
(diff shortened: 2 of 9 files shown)


compass/ocean/tests/baroclinic_channel/baroclinic_channel.cfg
```diff
@@ -48,4 +51,4 @@ gradient_width_dist = 40e3
 salinity = 35.0
 
 # Coriolis parameter for entire domain.
-coriolis_parameter = -1.2e-4
\ No newline at end of file
+coriolis_parameter = -1.2e-4
```
cbegeman, line 54: @xylar I haven't seen this before. I'm not sure how to fix it. Does it mean there should be a line at the end of the document?

cbegeman:
This PR makes changes consistent with https://github.com/MPAS-Dev/compass/pull/501 and was prompted by @xylar's review suggestions.

compass/ocean/tests/baroclinic_channel/baroclinic_channel.cfg
xylar: The lack of a new line before your changes is undesirable so what you've done here is correct.  Thank you!

xylar:
@cbegeman, I think this PR might also be a good place to fix this docstring:
https://github.com/MPAS-Dev/compass/blob/main/compass/ocean/rpe.py#L9-L23
which should be something like:
```
    Computes the reference (resting) potential energy for the whole domain
    Parameters
    ----------
    initial_state_file_name : str
        Name of the netCDF file containing the initial state

    output_file_prefix : str
        Prefix for the netCDF file containing output of forward step

    num_files : int
        Number of files on which to compute rpe

    Returns
    -------
    rpe : numpy.ndarray
          the reference potential energy of size (num_timesteps) x (num_files)
```

cbegeman:
> @cbegeman, I think this PR might also be a good place to fix this docstring

Done! Thanks for pointing this out.

xylar:
## Testing

I verified that both `ocean/baroclinic_channel/10km/rpe_test` and `ocean/internal_wave/rpe_test` are bit-for-bit compared with `main`.

xylar (approved):
@cbegeman, thank you for taking the time to do this!

PR #519, opened 2023-01-31
Port cosine bell initial state

cbegeman:
This PR moves the generation of an initial state for the ocean global cosine bell test case from init mode to the initial state compass step. 

Checklist

### DIFF
--- a/compass/ocean/tests/global_convergence/cosine_bell/__init__.py
+++ b/compass/ocean/tests/global_convergence/cosine_bell/__init__.py
@@ -1,12 +1,15 @@
 from compass.config import CompassConfigParser
-from compass.testcase import TestCase
-
-from compass.mesh.spherical import QuasiUniformSphericalMeshStep, \
-    IcosahedralMeshStep
-from compass.ocean.tests.global_convergence.cosine_bell.init import Init
+from compass.mesh.spherical import (
+    IcosahedralMeshStep,
+    QuasiUniformSphericalMeshStep,
+)
+from compass.ocean.tests.global_convergence.cosine_bell.analysis import (
+    Analysis,
+)
 from compass.ocean.tests.global_convergence.cosine_bell.forward import Forward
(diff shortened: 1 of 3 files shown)


cbegeman:
### Testing:

I have run both `icos` and `qu` tests on chrysalis with intel compiler and found that the changes in the initial state `normalVelocity` and `debugTracers` fields are O(1e-8). These diffs are maintained at the end of the forward simulation. This difference can be due to such differences in precision as the value of pi used by numpy vs the fortran code.

xylar (approved):
This looks excellent to me and the small differences seem at the expected level for differences in order of operations and the like.

cbegeman:
Thanks, @xylar, for the review!

xylar:
@cbegeman, would you be willing to rebase this and then fix any issues that pre-commit might suggest when CI runs again here?

xylar:
@cbegeman, you may have figured this out but you can run:
```
pre-commit run --show-diff-on-failure --color=always --files compass/ocean/tests/global_convergence/cosine_bell/__init__.py compass/ocean/tests/global_convergence/cosine_bell/cosine_bell.cfg compass/ocean/tests/global_convergence/cosine_bell/init.py
```
locally (and a few times) to find out what linting is needed.  Then, just make a commit to fix those issues.

cbegeman:
@xylar Thanks! That would definitely save me some time!

xylar:
@cbegeman, since we already ported this over to polaris, I'm tempted to merge.  Would you like to wait to hear from @lconlon and @alicebarthel first?

cbegeman:
@xylar Let's go ahead and merge. Thanks for reviewing!

xylar:
@cbegeman, thanks for adding this!

alicebarthel:
Thanks @cbegeman @xylar , this looks good to me (and I appreciate the opportunity to read through the files) and I'm glad you moved ahead with the merge.

PR #524, opened 2023-02-04
Constrain docutils

xylar:
This should fix an issue with bullet points in the documentation.

Checklist


xylar:
@matthewhoffman, let's see if this fixes the issue you reported to me.

xylar:
@matthewhoffman, I'm not able to reproduce this problem on Ubuntu.  It also looks like the current documentation is building correctly.  See bullet points here:
https://mpas-dev.github.io/compass/latest/users_guide/machines/cori.html

I'm going to close this for now.  We can re-open if the problem occurs elsewhere.

matthewhoffman:
Sounds good.  Thanks for looking into it.

PR #533, opened 2023-02-17
Enable submeso. parameterization in` global_ocean`

xylar:
This merge also adds submeso output variables to `daily_output` test.

Checklist


compass/ocean/tests/global_ocean/daily_output_test/streams.forward
```diff
@@ -127,4 +129,4 @@
     <var name="icebergHeatFlux"/>
 </stream>
 
-</streams>
+</streams>
```
xylar, line 132:
```suggestion
</streams>

```

compass/ocean/tests/global_ocean/namelist.forward
```diff
@@ -17,3 +17,4 @@ config_Redi_min_layers_diag_terms = 8
 config_use_GM = .true.
 config_AM_mixedLayerDepths_enable = .true.
 config_AM_mixedLayerDepths_compute_interval = 'dt'
+config_submesoscale_enable = .true.
```
xylar, line 20:
```suggestion
config_submesoscale_enable = .true.

```

xylar:
I tried to run QU240 tests with this branch but they are all failing with state validation errors, perhaps related to the masking PR https://github.com/E3SM-Project/E3SM/pull/5325 that we're trying to debug.

OSX CI failures are unrelated to this.

cbegeman:
@xylar I had in mind that we would add an additional test with submesoscale turned on. It could be just `daily_output_submeso_test` or up a level at `PHC_submeso`

xylar:
@cbegeman, we could do that, but the default in E3SM is to run with the parameterization on:
https://github.com/E3SM-Project/E3SM/blob/982d11520aec729ef32ee6991bd85162f26b917b/components/mpas-ocean/bld/namelist_files/namelist_defaults_mpaso.xml#L156
So it feels to me like it should be on in `global_ocean`, too.  This is consistent with GM being on by default in `global_ocean`.

cbegeman:
@xylar I didn't realize that. In that case, I'm fine with it being on by default.

cbegeman:
@xylar I've tested the nightly suite plus `ocean_global_ocean_QU240_PHC_daily_output_test` with the bugfix branch https://github.com/E3SM-Project/E3SM/pull/5458 and found that all tests pass (chrys, intel, openmpi).

cbegeman (approved):

xylar:
## Testing

I tested this on Chrysalis with Intel and OpenMPI.  I ran all QU240 test cases using the branch for https://github.com/E3SM-Project/E3SM/pull/5458.  All tests passed with this branch, whereas 8 forward mode tests failed with the version of MPAS-Ocean in the `E3SM-Project` submodule.

We will need to update the submodule before this goes in.

xylar:
The submodule was update to include https://github.com/E3SM-Project/E3SM/pull/5458 in #558 so this can now be merged.

xylar:
## Re-testing

I ran the `pr` test suite, comparing with the current `main` as a baseline and (to my surprise!) all tests passed.  I verified that submesoscale is turned off for the baseline and on for this run (according to the namelist options), so it seems the tests are short enough not to detect the difference.

I ran all QU240 test cases (this time without a baseline) and they finished successfully.

PR #538, opened 2023-02-19
Allow test cases to run on login nodes

### LINKED ISSUE #495: Allow compass run to be executed on a login node in some situations
In some cases, it is desirable to run `compass run` on a login node if the run operations for a case is serial and lightweight.  There are two ideas for how this could be implemented:
* as an attribute of a test case
* as a command line argument to the `compass run` command

Perhaps we want both options.

xylar:
If you are on a slurm system but a slurm job is not detected, you are assumed to be on a login node.  In that case, it is like being on a single node system, restricted to a small number of cores (set with the config option `login_cores = 4` in the `[parallel]` config section).  In addition, you are not allowed to run MPI jobs (steps with `ntasks` > 1).

This is enforced by modifying the `Step.constrain_resources()` method, which now takes a dictionary of `available_resources` rather than a single `available_cores` argument.  (Many tasks have their overridden `constrain_resources()` method changed as a result.)

The new, stricter linter required cleaning up some code in the process.

Checklist

closes #495

### DIFF
--- a/compass/default.cfg
+++ b/compass/default.cfg
@@ -28,6 +28,9 @@ verify = True
 # the program to use for graph partitioning
 partition_executable = gpmetis
 
+# the number of cores a user can use on a login node
+login_cores = 4
+
 
 # The io section describes options related to file i/o
 [io]
--- a/compass/parallel.py
+++ b/compass/parallel.py
@@ -6,7 +6,7 @@ import warnings
 from mpas_tools.logging import check_call
 
 
-def get_available_cores_and_nodes(config):
+def get_available_parallel_resources(config):
     """
     Get the number of total cores and nodes available for running steps
 
@@ -17,14 +17,15 @@ def get_available_cores_and_nodes(config):
 
     Returns
     -------
-    cores : int
-        The number of cores available for running steps
-
-    nodes : int
-        The number of cores available for running steps
+    available_resources : dict
+        A dictionary containing available resources (cores, tasks, nodes
+        and cores_per_node)
     """
 
     parallel_system = config.get('parallel', 'system')
+    if parallel_system == 'slurm' and 'SLURM_JOB_ID' not in os.environ:
+        parallel_system = 'login'
+
     if parallel_system == 'slurm':
         job_id = os.environ['SLURM_JOB_ID']
         node = os.environ['SLURMD_NODENAME']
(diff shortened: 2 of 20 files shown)


xylar:
## Testing

So far, I've only done do-no-harm testing of this PR.  I'm able to run the `nightly` ocean test suite on my laptop using the (meaningless) `--on_login` flag.

xylar:
@matthewhoffman, this is ready for you to try out (and to use in #488 if you like).  You would need to set `on_login = True` in the config options for those tests where appropriate.

matthewhoffman:
@xylar , thanks for taking care of this.  I'll test it out next time I test or set up an ensemble.

matthewhoffman:
@xylar , do you expect it would work to set `on_login=True` for some steps in a test case but have others be set to False?  For #488 , we want to be able to run the step `ensemble_manager` from a login node, but the steps that form the ensemble members still need to know they are on a compute node.

xylar:
No that wouldn't work in the current implementation. The config options have to be the same for all steps in a test case. But one option might be to use the command line flag in that particular situation. I'll have to think about a way to include that in the step itself.

xylar:
@matthewhoffman, I think this solution will be much messier to review but much better to use.

Essentially, you don't have to do anything different now for a step (or even a test case or a suite) to run on a login node.  It just has to be made up of steps that don't use MPI and don't use more than `login_cores = 4` CPUs (e.g in python threading).

I haven't had a chance to test this yet so hold off on doing anything but responding to the concept until I have a chance.

xylar:
@matthewhoffman, okay, I tested this to my satisfaction.  Have a look and tell me what you think.  If you have something to test on, give it a whirl!

matthewhoffman:
Thanks, @xylar .  I'll try it out with an ensemble test and suggest @trhille apply it when he tests #488 .

compass/default.cfg
```diff
@@ -28,6 +28,9 @@ verify = True
 # the program to use for graph partitioning
 partition_executable = gpmetis
 
+# the number of cores a user can use on a login node
+login_cores = 4
```
matthewhoffman, line 32: Is it allowable to have this be more than 1 at most centers?

compass/parallel.py
```diff
@@ -37,65 +38,42 @@ def get_available_cores_and_nodes(config):
         else:
             args = ['sinfo', '--noheader', '--node', node, '-o', '%Z']
             threads_per_core = _get_subprocess_int(args)
-        cores_per_node = sockets_per_node*cores_per_socket*threads_per_core
+        cores_per_node = sockets_per_node * cores_per_socket * threads_per_core
         args = ['squeue', '--noheader', '-j', job_id, '-o', '%D']
         nodes = _get_subprocess_int(args)
         cores = cores_per_node * nodes
+        mpi_allowed = True
+    elif parallel_system == 'login':
```
matthewhoffman, line 46: I like this solution - defining a different system for a login node seems like it would provide the flexibility needed to handle the situation properly.

trhille:
I used this when testing https://github.com/MPAS-Dev/compass/pull/488, and I was able to use `compass run` from the login node to run the `ensemble_manager` to set up the default Thwaites ensemble.

matthewhoffman (approved):
@xylar , thanks for reworking the solution for this issue.  I tested this PR merged into my ensemble branch and it worked like a charm.  I made a cursory review of the parts of the code changes that I understood.  I'm approving now because my comment is minor.

compass/default.cfg
xylar: Yes, this is thread parallelism and we're often allowed to use a few cores (e.g. you can compile code in parallel)

xylar:
Thanks @matthewhoffman for your review and @trhille for trying it out, too!

PR #540, opened 2023-02-21
Remove init mode namelist, stream from cosine bell case

cbegeman:
Fixes https://github.com/MPAS-Dev/compass/pull/519, which replaced an `init_mode` run of MPAS-Ocean with the `initial_state` step. We remove the namelist and streams corresponding to `init_mode` from the test case.


cbegeman:
### Testing

I ran successfully ran `ocean/global_convergence/qu/cosine_bell` on chrysalis with intel, openmpi.

xylar (approved):
Thanks @cbegeman! I fixed the issues that the linter found.  This will be helpful for running with cached initial conditions.

cbegeman:
@xylar Thanks for the further fixup!

PR #547, opened 2023-02-28
Add baroclinic gyre tests

alicebarthel:
This PR aims to add the MITgcm baroclinic test case to compass. Started as a Work In Progress [WIP] PR to allow discussion and debugging before merging. 

Checklist


xylar:
Nice! All tests passing!

compass/ocean/__init__.py
```diff
@@ -43,6 +45,7 @@ def __init__(self):
         self.add_test_group(IceShelf2d(mpas_core=self))
         self.add_test_group(IsomipPlus(mpas_core=self))
         self.add_test_group(MerryGoRound(mpas_core=self))
+        self.add_test_group(MitgcmBaroclinicGyre(mpas_core=self))
```
cbegeman, line 48: I'd be inclined to call this test group `BaroclinicGyre` rather than `MitgcmBaroclinicGyre`. I think shorter is better, and I doubt we'll be implementing multiple baroclinic gyre cases.

compass/ocean/tests/mitgcm_baroclinic_gyre/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    A test group for MITgcm baroclinic gyre test cases
+    """
+    def __init__(self, mpas_core):
+        """
+        mpas_core : compass.MpasCore
+            the MPAS core that this test group belongs to
+        """
+        super().__init__(mpas_core=mpas_core, name='mitgcm_baroclinic_gyre')
+
+        for resolution in ['80km']:
```
cbegeman, line 18: Consider adding a comment that resolutions here must be added with the 'km' suffix (since later you chop off the last 2 characters to get the float)

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
```diff
@@ -0,0 +1,131 @@
...
+    init_vertical_coord(config, ds)
+
+    # setting the initial conditions
+    temperature = (-11. * np.log(0.0414 *
+                   (-1. * ds.zMid + 100.3)) + 48.8)
+    temperature = temperature.transpose('Time', 'nCells', 'nVertLevels')
+    bottT = temperature[0, 200, -1].values
+    surfT = temperature[0, 200, 0].values
+    print(f'bottom T: {bottT} and surface : {surfT}')
+    salinity = 34.0 * xr.ones_like(temperature)
```
cbegeman, line 71: Consider making certain initial temperature and salinity parameters a part of the `mitgcm_baroclinic_gyre` section in your config file

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
```diff
@@ -0,0 +1,131 @@
...
+        (tempMax - tempMin) * (latMax - lat) / (latMax - latMin) + tempMin
+    temperatureSurfaceRestoringValue = \
+        temperatureSurfaceRestoringValue.expand_dims(dim='Time', axis=0)
+
+    temperaturePistonVelocity = \
+        restoring_temp_piston_vel * xr.ones_like(
+            temperatureSurfaceRestoringValue)
+
+    salinitySurfaceRestoringValue = \
+        34.0 * xr.ones_like(temperatureSurfaceRestoringValue)
```
cbegeman, line 119:
```suggestion
        initial_salinity * xr.ones_like(temperatureSurfaceRestoringValue)
```
Where `initial_salinity` is given in the config file as suggested above

compass/ocean/tests/mitgcm_baroclinic_gyre/gyre_test_case.py
```diff
@@ -0,0 +1,61 @@
...
+
+        long : bool
+            Whether to run a long (3-year) simulation to quasi-equilibrium
+        """
+        name = 'performance'
+        self.resolution = resolution
+        self.long = long
+
+        if long:
+            name = 'long'
```
cbegeman, line 41: We've sometimes called these something like `performance_test` and `3_year_test`. You might consider changing it after deciding how long the test should be. I know the soma test case uses `long` but I think it's nice to know as a user how long it will be.

cbegeman:
@alicebarthel I think this is coming together really nicely! I left only minor comments at this point.

compass/ocean/__init__.py
xylar: I'm fine with the shorter name.  @alicebarthel, do you remember why we settled on the longer name?  Presumably because we did think we might have other baroclinic gyre tests?

compass/ocean/tests/mitgcm_baroclinic_gyre/__init__.py
xylar: An alternative would be to provide the resolution as a float and add the `km` suffix to the subdirectory.  I think that might be cleaner.  (I know we were following ziso, so that one should probably be fixed in a similar way.)

cbegeman: By the way, I have something along these lines here, where the resolution is always given in meters
https://github.com/MPAS-Dev/compass/blob/f8af494f5c277c381d1b02320c1b9e297d01f9da/compass/ocean/tests/turbulence_closure/default/__init__.py#L38-L41

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
xylar: The initial temperature is a function of depth.  @alicebarthel has constructed a function that's a pretty good approximation of the table of numbers used in the MITgcm version of the test case.  It's full of "magic" numbers so they wouldn't be super easy to add to the config file.  But maybe there would be a way to give the top (z=0) and bottom (z=bottom_depth) temperature and have them vary between those two with your prescribed functional form?  That would seem cleaner to me.

I wouldn't hardcode sampling `bottT` and `surfT` at the 200th grid cell here.  Either you should take grid cell 0 or you'll want to have a number somehow based on `nCells`.  We don't necessarily know that the mesh will have a 200th grid cell (though it's obviously highly likely) since we're flexible about the resolution.

Regarding the initial salinity, I agree for sure that that needs to be a config option rather than a magic number here.

compass/ocean/__init__.py
alicebarthel: I like the idea of a more general (and shorter) name, unless it significantly increases the scope of this new test case. 
As I remember it: as a dev step, I preferred to err on the side of too specific because I was not sure how much of the set-up was going to be hard-coded vs. generalized, or whether other baroclinic cases would be developped in the future. We hoped to have a narrowly defined scope of reproducing the MITgcm test case - so went with the specific name. 
As it currently stands, I could see this becoming a general baroclinic gyre case if we include options to set T,S in the config file. Does it matter that it is on spherical coordinates (vs. barotropic on planar)? Is there anything else that seems specific enough to either include in the name or at least in the description?

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
alicebarthel: Good point @xylar re. cell 200. @cbegeman I like the idea of it being more general (setting T in cfg) but I have not come up with a satisfying way to do that at this stage. For the original purpose of matching MITgcm, hard-coding the structure was easiest. It can extend to a deeper set-up, but approximates their values over 0-1800m. I'll look to see about setting it from top and bottom T keeping the existing functional form. We could do a prettier functional form with, say, a thermocline depth parameter setting the change in T=f(z) shape, but I have not looked into it that closely.

alicebarthel:
Thanks @cbegeman and @xylar for your comments so far!

compass/ocean/__init__.py
cbegeman: I think that you can consider it a general baroclinic gyre test case and go with the shorter name. The way I see it, it is the responsibility of whomever might want to develop a baroclinic gyre with different properties to change the name at that point. The planar vs. spherical distinction can also happen down the line (e.g., as it is done for the `global_convergence` test case where the path of the generated test case includes `planar` vs. `spherical`). 

I think it will be clear that the default configuration is the one that is meant for comparison with the MITgcm output (if you find a way to include that output in the analysis). Or maybe it will only be apparent in the documentation, where you have references and maybe copy an image from the MITgcm page documenting their results. Or maybe you change the `default` test name to `mitgcm_comparison`. I'm not sure what the best route is at this point. @alicebarthel Maybe you have preferences here?

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
xylar: @alicebarthel, couldn't you have config options that are the temperatures in the middle of the top and bottom layers respectively?  Then, couldn't you take your existing functional form but stretch it linearly so it hits the desired top and bottom temperatures?

Alternatively, if you want to be very particular about having the same temperature profile independent of resolution, couldn't you do the same but instead of the middle of the top and bottom layers, you would use z = 0 and z=1800 m.  The default values would be chosen such that the temperature in the middle of the top and bottom layers is quite close to the value you currently get with your functional form.

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
```diff
@@ -0,0 +1,131 @@
...
+    normalVelocity = xr.zeros_like(ds.xEdge)
+    normalVelocity, _ = xr.broadcast(normalVelocity, ds.refBottomDepth)
+    normalVelocity = normalVelocity.transpose('nEdges', 'nVertLevels')
+    normalVelocity = normalVelocity.expand_dims(dim='Time', axis=0)
+
+    ds['temperature'] = temperature
+    ds['salinity'] = salinity
+    ds['normalVelocity'] = normalVelocity
+
+    omega = 2. * np.pi / 86164.
```
xylar, line 82: https://github.com/E3SM-Project/polaris/blob/0b742bea7e20dfa0e58fc800f3626ed02d966a24/polaris/ocean/tasks/geostrophic/init.py#L69

compass/ocean/tests/mitgcm_baroclinic_gyre/__init__.py
alicebarthel: @cbegeman Working on this `resolution` change now. A few questions/suggestions:
1) Are you satisfied with resolution in meters by default? It makes sense for turbulence closure since we go to fine resolution. Is it best to be consistent across cases? Happy to switch mine, just wanted to check. 
2) the comment blocks in your `forward` (L12, [L25](https://github.com/MPAS-Dev/compass/blob/f8af494f5c277c381d1b02320c1b9e297d01f9da/compass/ocean/tests/turbulence_closure/forward.py#L25)) and `initial_state` (L22, L34) still carry comments where resolution is expected to be string, though it is the float that is passed (as far as I understand). It may be good to propagate the update.
3) The [`run`](https://github.com/MPAS-Dev/compass/blob/f8af494f5c277c381d1b02320c1b9e297d01f9da/compass/ocean/tests/turbulence_closure/forward.py#L87-L93) in `forward` still assumes a string resolution with units. I couldn't find a call for it so maybe it's just stale code, but worth updating in case it ever get exercised.

compass/ocean/__init__.py
alicebarthel: Renamed the case. I like the idea of changing the default test name to mitgcm_comparison, so something to keep in mind for later.

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
alicebarthel: Moved the constant salinity value to the cfg file. The initial temperature profile is now set (still with a "magical" functional form and a fitted parameter) from the top and bottom temperature set at the surface (z=0m) and bottom (z = bottom_depth). The default values approximate the mitgcm stratification, but we could add a table to reach to be more exact.

compass/ocean/tests/mitgcm_baroclinic_gyre/gyre_test_case.py
alicebarthel: Added @cbegeman's suggestion.

compass/ocean/tests/mitgcm_baroclinic_gyre/initial_state.py
alicebarthel: done.

alicebarthel:
I made some offline plots for sanity checks on the initial and forcing files. The more useful ones are these 3:
forcing_temperatureSurfaceRestoringValue
forcing_windStressZonal
initial_temperature_fz
They contain numbers (mean, std) on the other variables to flag if something is modified in other fields.

alicebarthel:
I made some (offline) plots to check the simulation state -- as a sanity check. They still need minor adjustments (e.g. I am unsure about the surface heat flux calculation) but are promising. At this stage, the "mean" surface state was calculated from 30-day snapshots over year 3 (the last year of the 3_year_test). 
baroclinicgyre_80km_meansurfacestate_yr3
I plotted a spin-up plot up to year 33. As expected, 3 years is far from enough for the spin-up (the MITgcm example show year 100). 
baroclinicgyre_80km_KEft_toyr33

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,171 @@
+.. _baroclinic_gyre:
+
+baroclinic_gyre
+=======================
+
+The ``baroclinic_gyre`` test group implements variants of the
+baroclinic ocean gyre set-up from the 
+`MITgcm test case <https://mitgcm.readthedocs.io/en/latest/examples/baroclinic_gyre/baroclinic_gyre.html>`_.
```
xylar, line 8: I think sphinx is getting confused by having this same link twice.  I would suggest putting the following at the bottom of the page:
```
.. _MITgcm test case: https://mitgcm.readthedocs.io/en/latest/examples/baroclinic_gyre/baroclinic_gyre.html
```
And then take the URL out of this and the following link:
```suggestion
`MITgcm test case`_.
```

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,171 @@
...
+stress which is constant in time, :math:`\tau_{\lambda}`, varying sinusoidally in the
+north-south direction.
+
+.. figure:: ../images/baroclinic_gyre_config.png
+  :width: 95%
+  :align: center
+  :alt: baroclinic gyre configuration
+  :name: baroclinic_gyre_config
+
+  Schematic of simulation domain and wind-stress forcing function for baroclinic gyre numerical experiment. The domain is enclosed by solid walls. From `MITgcm test case <https://mitgcm.readthedocs.io/en/latest/examples/baroclinic_gyre/baroclinic_gyre.html>`_.
```
xylar, line 26: After moving the URL to the bottom of the page as mentioned above:
```suggestion
  Schematic of simulation domain and wind-stress forcing function for baroclinic gyre numerical experiment. The domain is enclosed by solid walls. From `MITgcm test case`_.
```

xylar (changes requested):
I had some issues mainly with the docs.  A few things to clean up there.

xylar:
I successfully ran the `80km/performance_test`, `80km/3_year_test` and `20km/performance_test`.  I didn't try to run the `20km/3_year_test`.

xylar:
@alicebarthel, the outputs you posted above look great to me!  I think we're almost there...

alicebarthel:
Thanks @xylar, that's super helpful. Attached are updated mean state plots for year 100 of the 80km. It looks nicely spun-up to me, at least dynamically: 
baroclinicgyre_80km_KEft_toyr100
baroclinicgyre_80km_meansurfacestate_yr100

cbegeman:
@alicebarthel This is awesome progress! Is this ready for review? Are you planning on adding a visualization or analysis step to aid comparison with MITgcm https://mitgcm.readthedocs.io/en/latest/examples/baroclinic_gyre/baroclinic_gyre.html#model-solution? The heat flux figure you posted looks like a good comparison with Fig 2.7. Do you think we want the analogs to Figs 4.8 or 4.9?

alicebarthel:
Thanks @cbegeman! For now, all the analysis is offline. Yes, there are a few more plots I want to make, including Figure 4.9b. Figure 4.8 may be good too - although at this stage the SSH contour looks good enough for me. Adding the barotropic streamfunction could be nice but feels lower priority for me. 
A question for you two: 
3 years is not enough for a full spin-up. The plots for MITgcm were made for year 100. 
1) Does it make sense to keep the 3_year_test even if it does not bring things to equilibrium? The expectation is for users to run it further as needed. For 80km, it looks like ~50 years is needed.  
2) Does it make sense to add an analysis/visualization step if the default run is not usually spun-up? 

At this stage, I am focusing on:
- [ ] Testing/debugging the restart capabilities. I have had issues but I suspect this could be a simple storage issue at this stage.
- [ ] enhancing the offline analysis to increase our confidence of the spun-up state (including at 20-km resolution) and include a readily available AMOC estimate.

I welcome comments and suggestions as I work on this.

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
alicebarthel: Thanks for the suggestion, I modified accordingly. The documentation built fine for me, and the link was correct.

compass/ocean/tests/baroclinic_gyre/forward.py
```diff
@@ -0,0 +1,118 @@
...
+            res_name = f'{int(resolution)}m'
+
+        res_params = {'80km': {'ntasks': 20,  # MODIFY
+                               'min_tasks': 2,
+                               'dt': "'01:00:00'",
+                               'btr_dt': "'00:03:00'",
+                               'mom_del4': "5.0e11",
+                               'run_duration': "'0000_03:00:00'"},
+                      '20km': {'ntasks': 128,
+                               'min_tasks': 10,
```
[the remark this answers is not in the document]
alicebarthel: Thanks, I was wondering about that. Thanks for the suggestion, I modified accordingly.

cbegeman:
> Does it make sense to keep the 3_year_test even if it does not bring things to equilibrium? The expectation is for users to run it further as needed. For 80km, it looks like ~50 years is needed.
Does it make sense to add an analysis/visualization step if the default run is not usually spun-up?

How about keeping the long test and having analysis/viz steps included in that test (and plot the last time slice in the output) but not include these steps in the performance test? I think it's reasonable to have the users change the duration. I didn't read through the docs yet, but let's make sure that ~50 year duration is recommended there.

compass/ocean/tests/baroclinic_gyre/forward.py
```diff
@@ -0,0 +1,118 @@
...
+    def run(self):
+        """
+        Run this step of the test case
+        """
+        run_model(self, partition_graph=True)
+
+        if self.name == '3_year_test':
+            replacements = {'config_do_restart': '.true.',
+                            'config_start_time': "'file'"}
+            self.update_namelist_at_runtime(replacements)
```
xylar, line 118: This isn't working because the name of the step is always `forward`, not `3_year_test`.  One fix is:
```suggestion
        if self.test_case.name == '3_year_test':
            replacements = {'config_do_restart': '.true.',
                            'config_start_time': "'file'"}
            self.update_namelist_at_runtime(replacements)
```
Or I think it might be more logical to store the `long` argument as an attribute and use it here:
```suggestion
        if self.long:
            replacements = {'config_do_restart': '.true.',
                            'config_start_time': "'file'"}
            self.update_namelist_at_runtime(replacements)
```

compass/ocean/tests/baroclinic_gyre/forward.py
```diff
@@ -0,0 +1,118 @@
...
+                               'dt': "'01:00:00'",
+                               'btr_dt': "'00:03:00'",
+                               'mom_del4': "5.0e11",
+                               'run_duration': "'0000_03:00:00'"},
+                      '20km': {'ntasks': 384,
+                               'min_tasks': 32,
+                               'dt': "'00:20:00'",
+                               'btr_dt': "'0000_00:00:20'",
+                               'mom_del4': "2.0e10 ",
+                               'run_duration': "'0000_00:01:00'"}}
```
xylar, line 55:
```suggestion
                               'run_duration': "'0000_01:00:00'"}}
```
I'm guessing this should be 1 hour (3 time steps) rather than 1 minute (1/20 of a time step, which gets rounded up to 1 time step).

compass/ocean/tests/baroclinic_gyre/initial_state.py
```diff
@@ -0,0 +1,151 @@
...
+    print(f'zmid: {ds.zMid[0,0,:].values}')
+    cc = 0.049  # 0.049 value from optimization on 1800m
+    aa = temp_top / np.log(cc)
+    bb = (cc ** (temp_bot / temp_top) - cc)
+    temperature = aa * np.log(bb * -1.0 * ds.zMid / bottom_depth + cc)
+    print(f'Tinit: {temperature[0,0,:].values}')
+    val_bot = aa * np.log(bb + cc)
+    val_top = aa * np.log(cc)
+    print(f'analytical bottom T: {val_bot} at \
+            depth : {bottom_depth}')
```
xylar, line 81: Here and in similar places, this is the preferred way to continue an f-string on a subsequent line:
```suggestion
    print(f'analytical bottom T: {val_bot} at '
            f'depth : {bottom_depth}')
```

alicebarthel: Thanks @xylar , I modified these. Let me know if I missed others.

alicebarthel:
@xylar @cbegeman This is ready for a review. I ran the performance and 3_year tests on Chrysalis without issue. I also ran the 3_year test for longer (50years, which takes ~2hours for 80km set-up). 
Thanks @xylar for working with me on the heat fluxes, this looks much better!
Results after 3 years: 
meansurfacestate_last1years
spinup_ft
Results after 50 years (average over last 10years): 
meansurfacestate_last10years
spinup_ft
time_avg_moc_last10years

xylar:
@alicebarthel, that looks awesome!  I'm glad we got the surface restoring worked out.

I'll review it tomorrow.

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,180 @@
+.. _baroclinic_gyre:
+
+baroclinic_gyre
```
xylar, line 3: This needs to be added to the index.rst for test groups, just as in the developer's guide.

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,180 @@
...
+The grid employs spherical polar coordinates with 15 vertical layers.
+The configuration is similar to the double-gyre setup first solved numerically
+in `Cox and Bryan (1984) <https://journals.ametsoc.org/view/journals/phoc/14/4/1520-0485_1984_014_0674_anmotv_2_0_co_2.xml>`_: the model is configured to
+represent an enclosed sector of fluid on a sphere, spanning the tropics to mid-latitudes,
+:math:`60^{\circ} \times 60^{\circ}` in lateral extent.
+The fluid is :math:`1.8`\ km deep and is forced by a zonal wind
+stress which is constant in time, :math:`\tau_{\lambda}`, varying sinusoidally in the
+north-south direction.
+
+.. figure:: ./images/baroclinic_gyre_config.png
```
xylar, line 20: We typically limit images to 500 pixels along the longer size to keep the repo small.  This file isn't super large but what do you think about shrinking it by a factor of 4?

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,180 @@
...
+
+Initial state
+-------------
+
+Initially the fluid is stratified
+with a reference potential temperature profile that varies from (approximately) :math:`\theta=30.6 \text{ } ^{\circ}`\ C
+in the surface layer to :math:`\theta=1.56 \text{ } ^{\circ}`\ C in the bottom layer. To ensure that the profile is independent of the vertical discretization, the profile is now set by a surface value (at the top interface) and a bottom value (at the bottom interface), set in the `.cfg` file. The default values have been chosen for the layer values (calculated with `zMid`) to approximate the discrete values presented in the MITgcm test case. The temperature functional form (and inner parameter `cc`  was determined by fitting an analytical function to the MITgcm discrete layer values (originally ranging from 2 to :math:`30 \text{ } ^{\circ}`\ C. If the `bottom_depth` is different from the default 1800m value, the temperature profile is stretched in the vertical to fit the surface and bottom temperature constraints, but the thermocline depth and the discrete layer values will move away from the MITgcm test case.
```
xylar, line 62:
```suggestion
in the surface layer to :math:`\theta=1.56 \text{ } ^{\circ}`\ C in the bottom layer. To ensure that the profile is independent of the vertical discretization, the profile is now set by a surface value (at the top interface) and a bottom value (at the bottom interface), set in the ``.cfg`` file. The default values have been chosen for the layer values (calculated with ``zMid``) to approximate the discrete values presented in the MITgcm test case. The temperature functional form (and inner parameter ``cc``  was determined by fitting an analytical function to the MITgcm discrete layer values (originally ranging from 2 to :math:`30 \text{ } ^{\circ}`\ C. If the ``bottom_depth`` is different from the default 1800m value, the temperature profile is stretched in the vertical to fit the surface and bottom temperature constraints, but the thermocline depth and the discrete layer values will move away from the MITgcm test case.  
```

xylar: Maybe break up these long lines?

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,180 @@
...
+baroclinic_gyre test case for a short (3 time steps) test run and validation of
+prognostic variables for regression testing.
+
+3_year_test
+-----------
+
+``ocean/baroclinic_gyre/3_year_test`` is an additional version of the
+baroclinic_gyre test case with a longer (3-year) spin-up. By default, it 
+includes monthly mean output, and plots the mean state of the simulation for the last 1 year (option in the config file). 
+Note that for the 80km configuration, the estimated time to equilibration is roughly 50 years (approx 3hours on default layout).
```
xylar, line 176:
```suggestion
Note that for the 80km configuration, the estimated time to equilibration is roughly 50 years (approx 3 hours of compute time on default layout).  
```

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,180 @@
...
+    restoring_temp_timescale = 30.
+
+performance_test
+----------------
+
+``ocean/baroclinic_gyre/performance_test`` is the default version of the
+baroclinic_gyre test case for a short (3 time steps) test run and validation of
+prognostic variables for regression testing.
+
+3_year_test
```
xylar, line 170: Somewhere here (or above), I would say that a user can run the `forward` step several times, advancing the simulation by 3 years each time.  Or that they can alter the namelist file to increase the duration of the simulation.

compass/ocean/tests/baroclinic_gyre/forward.py
```diff
@@ -0,0 +1,122 @@
...
+                              streams='streams.forward',
+                              template_replacements=replacements)
+        options = dict()
+        for option in ['dt', 'btr_dt', 'mom_del4', 'run_duration']:
+            options[f'config_{option}'] = res_params[option]
+        if long:
+            # run for 3 years instead of 3 time steps
+            options['config_start_time'] = "'0001-01-01_00:00:00'"
+            options['config_stop_time'] = "'0004-01-01_00:00:00'"
+            options['config_run_duration'] = "'none'"
```
xylar, line 94:
```suggestion
            options['config_start_time'] = "'0001-01-01_00:00:00'"
            options['config_stop_time'] = "'0004-01-01_00:00:00'"
            options['config_run_duration'] = "'none'"
```
I wonder if it might not be better to set a 3-year run duration here.  The problem with setting a start and stop time is that it somewhat defeats the purpose of switching to "restart" mode at the end of the first run of the forward step.  If a user were to run a second time, the simulation would immediately finish because it's already at year 4.  But if you instead set run duration, you can run for another 3 years each time you rerun.

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
xylar: Note however that you would need to make some changes I suggested to `forward.py` for this to actually work.

compass/ocean/tests/baroclinic_gyre/moc.py
```diff
@@ -0,0 +1,104 @@
...
+        dlat = 0.25  # set in config next
+        # latbins = np.arange(15.5, 75.5, 0.25)
+        latBins = np.arange(lat_min + 2 * dlat,
+                            lat_max + 2 * dlat, dlat)
+        nz = self.config.getint('vertical_grid', 'vert_levels')
+
+        dsMesh = xarray.open_dataset('../initial_state/initial_state.nc')
+
+        ds = xarray.open_mfdataset(
+            '{}/timeSeriesStatsMonthly*.nc'.format(in_dir),
```
xylar, line 60:
```suggestion
            f'{in_dir}/timeSeriesStatsMonthly*.nc',
```

compass/ocean/tests/baroclinic_gyre/moc.py
```diff
@@ -0,0 +1,104 @@
...
+        dsMOC = xarray.Dataset()
+        dsMOC['xtime_startMonthly'] = ds.xtime_startMonthly
+        dsMOC['xtime_endMonthly'] = ds.xtime_endMonthly
+        dsMOC['moc'] = (["Time", "latBins", "nVertLevelsP1"], moc)
+        dsMOC.coords['latBins'] = latBins
+        dsMOC.coords['nVertLevelsP1'] = np.arange(nz + 1)
+        dsMOC.moc.attrs['units'] = 'Sv'
+        dsMOC.moc.attrs['description'] = \
+            'zonally-averaged meridional overturning streamfunction'
+        outputFileName = '{}/moc.nc'.format(out_dir)
```
xylar, line 79:
```suggestion
        outputFileName = f'{out_dir}/moc.nc'
```

compass/ocean/tests/baroclinic_gyre/moc.py
```diff
@@ -0,0 +1,104 @@
...
+        # if file_complete(ds, outputFileName):
+        #    return
```
xylar, line 81: Remove?

compass/ocean/tests/baroclinic_gyre/moc.py
```diff
@@ -0,0 +1,104 @@
...
+        dsMesh = xarray.open_dataset('../initial_state/initial_state.nc')
+
+        ds = xarray.open_mfdataset(
+            '{}/timeSeriesStatsMonthly*.nc'.format(in_dir),
+            concat_dim='Time', combine='nested')
+
+        # print(dsMesh)
+        # print(ds)
+        # print(latBins)
+        # print(nz)
```
xylar, line 66: Remove?

compass/ocean/tests/baroclinic_gyre/viz.py
```diff
@@ -0,0 +1,140 @@
...
+        amoc = "max MOC = {:.1e}".format(round(np.max(moc), 1))
+        maxloc = 'at lat = {} and z = {}m'.format(
+            latbins[idx[-1]].values, int(dsMesh.refInterfaces[idx[0]].values))
+        maxval = 'max MOC = {:.1e} at lat={}-{}'.format(
+            round(np.max(moc[:, 175]), 1),
+            latbins[175 - 1].values, latbins[175].values)
+        plt.annotate(amoc + '\n' + maxloc + '\n' + maxval,
+                     xy=(0.01, 0.05), xycoords='axes fraction')
+        plt.colorbar()
+        plt.savefig('{}/time_avg_moc_last{}years.png'.format(out_dir, avg_len))
```
xylar, line 72:
```suggestion
        plt.savefig(f'{out_dir}/time_avg_moc_last{avg_len}years.png')
```

compass/ocean/tests/baroclinic_gyre/viz.py
```diff
@@ -0,0 +1,140 @@
...
+        plt.gca().invert_yaxis()
+        plt.ylabel('Depth (m)')
+        plt.xlabel('Latitude')
+        idx = np.unravel_index(np.argmax(moc), moc.shape)
+        amoc = "max MOC = {:.1e}".format(round(np.max(moc), 1))
+        maxloc = 'at lat = {} and z = {}m'.format(
+            latbins[idx[-1]].values, int(dsMesh.refInterfaces[idx[0]].values))
+        maxval = 'max MOC = {:.1e} at lat={}-{}'.format(
+            round(np.max(moc[:, 175]), 1),
+            latbins[175 - 1].values, latbins[175].values)
```
xylar, line 68: Ideally, break this into several lines where you store these various inputs in variables and then use f-strings rather than the `.format()` methods to put them in place.

compass/ocean/tests/baroclinic_gyre/viz.py
```diff
@@ -0,0 +1,140 @@
...
+        plt.savefig('{}/time_avg_moc_last{}years.png'.format(out_dir, avg_len))
+
+    def _plot_spinup(self, mon_dir, dsMesh, out_dir):
+        """
+        Plot the timeseries of monthy layer-mean
+        kinetic energy and temperature for the test case
+        """
+
+        ds = xarray.open_mfdataset(
+            '{}/timeSeriesStatsMonthly*.nc'.format(mon_dir),
```
xylar, line 81:
```suggestion
            f'{mon_dir}/timeSeriesStatsMonthly*.nc',
```

compass/ocean/tests/baroclinic_gyre/viz.py
```diff
@@ -0,0 +1,140 @@
...
+        ax[0].set_ylabel('Layer Mean Kinetic Energy ($m^2 s^{-2}$)')
+        ax[1].set_ylabel(r'Layer Mean Temperature ($^{\circ}$C)')
+        plt.savefig('{}/spinup_ft.png'.format(out_dir), bbox_inches='tight')
+
+    def _plot_mean_surface_state(self, mon_dir, dsMesh, out_dir):
+
+        lon = 180. / np.pi * dsMesh.variables['lonCell'][:]
+        lat = 180. / np.pi * dsMesh.variables['latCell'][:]
+        ds = xarray.open_mfdataset(
+            '{}/timeSeriesStatsMonthly*.nc'.format(mon_dir),
```
xylar, line 107:
```suggestion
            f'{mon_dir}/timeSeriesStatsMonthly*.nc',
```

compass/ocean/tests/baroclinic_gyre/viz.py
```diff
@@ -0,0 +1,140 @@
...
+        plt.colorbar(temp, ax=ax[1])
+        ax[2].tricontour(lon, lat, np.mean(ds.timeMonthly_avg_ssh[-12 * avg_len:, :], axis=0),  # noqa: E501
+                         levels=14, linewidths=.8, colors='k')
+        hf = ax[2].tricontourf(lon, lat, np.mean(heatflux[-12 * avg_len:, :], axis=0),  # noqa: E501
+                               levels=np.linspace(- absmax, absmax, 27), cmap="RdBu_r")  # noqa: E501
+        plt.colorbar(hf, ax=ax[2])
+
+        ax[0].set_title('SSH (m)')
+        ax[1].set_title(r'SST ($^\circ$C)')
+        ax[2].set_title('Heat Flux (W/m^{2})')
```
xylar, line 134:
```suggestion
        ax[2].set_title('Heat Flux (W/m$^{2}$)')
```

compass/ocean/tests/baroclinic_gyre/viz.py
```diff
@@ -0,0 +1,140 @@
...
+
+        ax[0].set_ylabel(r'Latitude ($^\circ$)')
+        for axis in ax:
+            axis.set_xlabel(r'Longitude ($^\circ$)')
+        plt.savefig('{}/meansurfacestate_last{}years.png'.format(
+            out_dir, avg_len), bbox_inches='tight')
```
xylar, line 140:
```suggestion
        plt.savefig(f'{out_dir}/meansurfacestate_last{avg_len}years.png',
                    bbox_inches='tight')
```

xylar (commented):
@alicebarthel, this is some amazing work!  It's very nearly there!

I was able to run the 80km tests without a hitch!  I'm running the 20 km tests now.  In the meantime, I have several small suggestions that you could make, or we could discuss further.

cbegeman (approved):
I think this looks great! I just visually inspected the code. Would you like me to run tests on any particular machine or build docs?

docs/developers_guide/ocean/test_groups/baroclinic_gyre.rst
```diff
@@ -0,0 +1,70 @@
...
+~~~~~~~~~~~~~
+
+The class :py:class:`compass.ocean.tests.baroclinic_gyre.initial_state.InitialState`
+defines a step for setting up the initial state for each test case.
+
+First, a mesh appropriate for the resolution is generated using
+:py:class:`compass.mesh.QuasiUniformSphericalMeshStep`.  Then, the mesh is
+culled to keep a single ocean basin (with lat, lon bounds set in `.cfg` file).  A vertical grid is generated,
+with 15 layers of thickness that increases linearly with depth (10m increase by default), from 50m at the surface to 190m at depth (full depth: 1800m).
+Finally, the initial temperature field is initialized with a vertical profile to approximate the discrete values set in the `MITgcm test case <https://mitgcm.readthedocs.io/en/latest/examples/baroclinic_gyre/baroclinic_gyre.html>`_, uniform in horizontal space. The surface and bottom values are set in the `.cfg` file. Salinity is set to a constant value (34 psu, set in the `.cfg` file)  and initial
```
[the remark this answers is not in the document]
xylar: Yes, 79 characters actually. The width of a default terminal is 80, so that's the reasoning.

compass/ocean/tests/baroclinic_gyre/baroclinic_gyre.cfg
```diff
@@ -0,0 +1,55 @@
...
+
+# Surface temperature restoring
+restoring_temp_min = 0.
+restoring_temp_max = 30.
+
+# Restoring timescale for surface temperature (in days)
+restoring_temp_timescale = 30.
+
+# config options for the mean state visualization
+[mean_state_viz]
```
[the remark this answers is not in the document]
alicebarthel: Thanks! I modified this to include both viz and moc calculation parameters.

docs/developers_guide/ocean/api.rst
```diff
@@ -83,6 +83,26 @@ buttermilk_bay
...
+   BaroclinicGyre
+
+   GyreTestCase
+   GyreTestCase.configure
+
+   forward.Forward
+   forward.Forward.run
+
+   initial_state.InitialState
+   initial_state.InitialState.run
```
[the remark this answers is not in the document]
alicebarthel: Thanks! Is this supposed to be alphabetical or does the order matter?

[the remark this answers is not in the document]
xylar: Yes, alphabetical is easiest.

docs/users_guide/ocean/test_groups/index.rst
```diff
@@ -12,6 +12,7 @@ coming months.
    :titlesonly:
 
    baroclinic_channel
+   baroclinic_gyre
```
alicebarthel, line 15: @xylar Is this what you meant in your review of the users_guide above?

docs/users_guide/ocean/test_groups/baroclinic_gyre.rst
alicebarthel: @xylar I included that change in docs/users_guide/ocean/test_groups/index.rst. Am I missing something?

xylar: Sorry, I saw that and thought I deleted this comment. Please ignore it.

compass/ocean/tests/baroclinic_gyre/forward.py
alicebarthel: Great point, @xylar, thanks for noticing. I've usually modified it manually each time, but setting up for re-running it for 3years is much more user-friendly.

xylar (approved):
Looks great!  Thanks for making the requested changes.

PR #548, opened 2023-03-01
Fix `isomip_plus` streamfunction computation

xylar:
The fix involves replacing NaNs in transport with zeros, so that the resulting sums are not themselves contaminated with NaNs.

This merge also contains some linting-related clean up.

Checklist

### DIFF
--- a/compass/ocean/tests/isomip_plus/streamfunction.py
+++ b/compass/ocean/tests/isomip_plus/streamfunction.py
@@ -1,13 +1,14 @@
-import xarray
+import os
+
 import numpy
+import progressbar
 import scipy.sparse
 import scipy.sparse.linalg
-import progressbar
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I was able to run the 5 km zstar Ocean test case (including `simulation`, `streamfunction` and `viz` steps) with this fix, whereas I got the error reported in https://github.com/MPAS-Dev/compass/pull/471#pullrequestreview-1237942887 without it.

See:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230301/ocean0_5km_fix_osf2/ocean/isomip_plus/5km/z-star/Ocean0
```

xylar:
Great, thanks @cbegeman!

PR #550, opened 2023-03-06
Add `jigsaw-python` as submodule

xylar:
Checklist

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -97,6 +97,7 @@ jobs:
       - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
         name: Install compass
         run: |
+          git config --global url."https://github.com/".insteadOf "git@github.com:"
           ./conda/configure_compass_env.py \
             --env_only \
             --env_name compass_test \
--- a/.github/workflows/docs_workflow.yml
+++ b/.github/workflows/docs_workflow.yml
@@ -49,6 +49,7 @@ jobs:
       - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
         name: Install compass
         run: |
+          git config --global url."https://github.com/".insteadOf "git@github.com:"
           ./conda/configure_compass_env.py \
             --env_only \
             --env_name compass_test \
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "MALI-Dev"]
 	path = MALI-Dev
 	url = git@github.com:MALI-Dev/E3SM.git
+[submodule "jigsaw-python"]
+	path = jigsaw-python
+	url = git@github.com:dengwirda/jigsaw-python.git
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -243,7 +243,7 @@ def get_env_setup(args, config, machine, compiler, mpi, env_type, source_path,
 def build_conda_env(env_type, recreate, mpi, conda_mpi, version,
                     python, source_path, conda_template_path, conda_base,
                     env_name, env_path, activate_base, use_local,
-                    local_conda_build, logger, local_mache):
+                    local_conda_build, logger, local_mache, update_jigsaw):
 
     if env_type != 'dev':
         install_miniconda(conda_base, activate_base, logger)
@@ -295,7 +295,10 @@ def build_conda_env(env_type, recreate, mpi, conda_mpi, version,
     else:
         spec_filename = None
 
-    if not os.path.exists(env_path) or recreate:
+    if not os.path.exists(env_path):
+        recreate = True
+
+    if recreate:
         print(f'creating {env_name}')
         if env_type == 'dev':
             # install dev dependencies and compass itself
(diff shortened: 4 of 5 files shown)


xylar:
So far, I have just tested QU240 mesh and init, and they are bit-for-bit on my Linux laptop between the old (conda-package) approach and the new (submodule) one.  I'll try the `pr` and `full_integration` suites on Chrysalis as soon as I have time.

xylar:
@matthewhoffman, @trhille, @cbegeman and @mark-petersen

I'm not asking for a review on this PR but I'm making you aware of it and would welcome your comments.

@dengwirda requested that `jigsaw-python` be added as a submodule of compass (and eventually polaris).  His reasoning as I understand it is twofold.  First, it will allow development of jigsaw (and jigsaw-python) alongside compass (much as we currently support development of MPAS-Ocean and MALI alongside compass).  Second, he hopes that it will increase the visibility of jigsaw (and jigsaw-python) within E3SM workflows.

@dengwirda, please add as you see fit.

xylar:
## Testing

I ran the `pr` and `full_integration` suites on Chrysalis (with Intel and Gnu, respectively) and got bit-for-bit results comparing with the current `main`.

Results are here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230306
```

I will note that the `ocean_global_convergence_qu_cosine_bell` took 10 minutes with this branch but only 3 on `main`.  I'm hoping that was a random fluke.  There is no JIGSAW activity in that test because the initial conditions are cached.

xylar:
I need to note that `./conda/configure_compass_env.py` will take about 2 1/2 minutes longer than it did before this change.

cc @matthewhoffman, @trhille, @cbegeman and @mark-petersen

matthewhoffman:
@xylar & @dengwirda , thanks for the explanation on this.  I certainly want to support jigsaw flexibility and visibility , but I want to understand some of the details better.  In terms of developing jigsaw concurrently with compass, how much need is anticipated for that workflow?  I personally don't imagine being competent going down that road, but I'm curious what scenarios Darren might envision.  On a related note, could that functionality be accomplished by installing the jigsaw conda package into compass in edit mode, without needing to have jigsaw be a submodule within compass?  If such an arrangement is possible, that might be a smoother solution for a workflow that may only require occasional usage.  I'd also like to learn more about the goal of increasing jigsaw visibility, because I'm not sure this change will achieve that - I think the only people who'd notice/understand the difference would be the people flagged on this PR - but I can think of some other ways that would be more effective at promoting jigsaw as an important component in its own right.  I am open to this PR, but I'm mainly concerned about changes that increase the time to create or load environments (already a bit disruptively slow), so want to make sure this change would actually achieve the desired effects.

xylar:
>  On a related note, could that functionality be accomplished by installing the jigsaw conda package into compass in edit mode, without needing to have jigsaw be a submodule within compass? If such an arrangement is possible, that might be a smoother solution for a workflow that may only require occasional usage.

@matthewhoffman, unfortunately, conda packages can't be installed in edit mode.  Python packages can be installed from source code in edit mode.  So that's one of the things accomplished by having `jigsaw-python` as a submodule: we can install both it and `jigsaw` from source.  In theory, we could install `jigsaw` from the conda package and `jigsaw-python` from source but that would remove any flexibility to update `jigsaw-python` to a non-release version (it wouldn't be compatible with a conda package of `jigsaw` otherwise).

dengwirda:
Thanks very much for adding this @xylar. Following up on a few points from above:
- A big reason for doing this is to make it easier for me to bring new jigsaw features into the workflow via branches, rather than trying to have folks update conda packages and whatnot, which I've not had much success employing. Like with E3SM, MPAS, MALI, etc, I'm hoping it should be relatively straightforward to simply build compass using a particular jigsaw branch to test out new features.
- Gaining a little more visibility for jigsaw is something I'd like to achieve as well --- currently the full library is buried in a way I'd like to avoid going forward. @matthewhoffman and all, please let me know if you have any thoughts or suggestions on this.
- Re: compile time, I appreciate this is a cost to pay with this change, but I'm hoping with this occurring only at the `configure_compass_env` and not `load_dev_compass` step it's not too bad overall, with the full E3SM build time typically being quite a big burden anyway. I had worried that jigsaw's compile time might be due to my heavy use of C++ templates, but building with `-ftime-report` suggests this isn't the case, with most time apparently spent in the more general gcc / clang optim. passes.

matthewhoffman:
@dengwirda , thanks for the discussion about this.  In terms of the goal of increased visibility, for the typical user or non-user observer of compass, I'm skeptical this changes the visibility level much.  Xylar's suggestion of providing instructions for how one can manipulate jigsaw within a compass environment via a _separate_ jigsaw repo might be a stronger way to emphasize it as a distinct codebase.

Regarding the load time for configure_compass_env.py, I'm still concerned about increasing that.  As Xylar says, every time we begin work on a new branch (which might be every couple days or more for some people), the configure script needs to be run.  I'm already finding I'm slightly discouraged from playing around with new ideas knowing I have to sit and wait for that before I can get too far along.  For MALI applications, there also is a spack layer that needs to get configured that has added a few additional minutes, so I'm more sensitive to extending that further.

In terms of your first item about making co-development easier, we had bad experiences with submodules for enabling co-development in the past.  When we had the MPAS components in E3SM as submodules, it was a real headache, and we only saw benefits after injecting the MPAS components directly into the E3SM repo.  I'm not suggesting you would want to do that with jigsaw, but my point is I'm not sure using a submodule is much better than the conda env approach currently being used.

dengwirda:
Okay, fair enough --- feel free to close this then.
Just to reiterate that me providing jigsaw as a cona package is not something that I've found to be particularly successful, though I appreciate it's convenient for others.

xylar:
One option I thought of was that we would only install JIGSAW:
* if the conda environment doesn't exist yet
* if the `--recreate` flag is specified
* if you specify a `--update_jigsaw` flag (which I would add)
If you ran with an existing conda-environment name and didn't add one of the flags above, the version of JIGSAW would be unchanged and would not be built.

Changes to `jigsaw-python` would still affect the conda environment without any action required because I install `jigsaw-python` in editable mode (`python -m pip install -e .`).

xylar:
@matthewhoffman, would that help you with your typical workflow?  Or do you create a new environment each time you start a new project?  I've been mostly using the same `compass_test` environment for everything unless I'm doing development on two branches at once.

xylar:
@dengwirda, you said:
> Just to reiterate that me providing jigsaw as a cona package is not something that I've found to be particularly successful

I guess even after several conversations, I don't quite understand the nature of this lack of success.  Is it in your own workflow of working on JIGSAW and JIGSAW-Python (e.g. in combination with compass)?  Or in terms of trying to get more of us involved in developing JIGSAW(-Python)?  Or having us use and test new JIGSAW(-Python) features?

For all of these, it seems like these are partly problems of a technical nature that could be handled even without JIGSAW as a submodule (e.g. just installing things out of the JIGSAW repo directly).  For others, I'm afraid I don't think having the submodule will change things.  If you're not getting the level of engagement you'd like, we need to  have more in-person or telecon discussions about what we should be using more effectively from JIGSAW and how we might benefit from either using the latest code or making our own changes.

xylar:
On chrysalis, a call to `./conda/configure_compass_env.py` with:
* `--recreate` takes 5m 51s
* `--update_jigsaw` takes 5m 45s
* not updating jigsaw (either flag above) takes 1m 3s

xylar:
Running with `--recreate` on `main` takes 1m 20s.

xylar:
Based on the feedback we've gotten, I do think I'm going to close this. We can revisit it in the future if we can come up with a good strategy for making the extra install time worthwhile.

@dengwirda, in the meantime, we would all welcome any contributions you want to make to the documentation both to improve the visibility of JIGSAW and to provide instructions for installing a development branch into the compass conda environment.

PR #489, opened 2022-12-27
Add sea-ice graph partitioning to `files_for_e3sm` test case

xylar:
This merge also generalizes `files_for_e3sm` so it can be run with an existing mesh and graph file, not just one generated by compass.  Thus, it replaces `make_diagnostics_files`, which performed a subset of this functionality for an existing mesh.

Checklist
* [x] User's Guide has been updated
* [x] Developer's Guide has been updated
* [x] API documentation in the Developer's Guide (`api.rst`) has any new or modified class, method and/or functions listed
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

Closes #474

PR #559, opened 2023-03-11
Make mapping file for sea-ice partitioning in compass 

xylar:
This allows us to use `pyremap`, supplying a parallel executable and running with multiple MPI tasks.

It will also be more adaptable to task parallelism in the future.

This merge updates to `mpas_tools 0.19.0` to bring in functionality needed to support this change.

Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+        exclude: conda/recipe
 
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I made sea-ice mapping files using this approach:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230311/ec30to60e2r3_file_for_e3sm_metadata2/ocean/global_ocean/files_for_e3sm/seaice_graph_partition
```

The results look reasonable in Paraview (similar to #489)

PR #573, opened 2023-03-23
Change algorithm for constraining resources

### LINKED ISSUE #568: `cpus_per_task` should be constrained by `constrain_resources()` to `cores_per_node`
We want a  way for `cpus_per_task` to be greater than `cores_per_node` on some machines with smaller nodes.  This would allow us to use all the cores on a node even on big nodes, but to make sure not to use *more* cores than are on a node (currently not enforced!) on smaller nodes.

The situation I have in mind is when we're using thread parallelism in python (e.g. for creating region masks).

xylar:
Previously, nothing was preventing `cpus_per_task` from being more than the number of cores on a node.  This is inconvenient because unsafe or non-performant behavior will occur if more python threads are used than cores on a node.

This merge changes the way resources are constrained.  First, `cpus_per_task` is constrained to be less than the lower of the number of cpus on a node or the number of total cores available. Then, we check to see if `cpus_per_task` is smaller than the minimum allowed for the given step.  Next, we allow all tasks to have the same `cpus_per_task` and constrain the number of tasks not to exceed the total available resources. Finally, we check to make sure the number of tasks is not below the allowed minimum for the step.

There are 4 typical resource requirements for steps:
* serial (1 task and 1 cpu per task)
* threading or multiprocessing on a node (1 task, multiple cpus per task)
* MPI jobs without threading (multiple tasks, 1 cpu per task)
* MPI jobs with threading (multiple tasks, a few cpus per task)

This algorithm should handle all of these without difficulty.

Checklist

closes #568

### DIFF
--- a/compass/step.py
+++ b/compass/step.py
@@ -298,7 +298,8 @@ class Step:
         Parameters
         ----------
         available_resources : dict
-            The total number of cores available to the step
+            A dictionary containing available resources (cores, tasks, nodes
+            and cores_per_node)
         """
         mpi_allowed = available_resources['mpi_allowed']
         if not mpi_allowed and self.ntasks > 1:
@@ -307,24 +308,21 @@ class Step:
                 'Please switch to a compute node.')
 
         available_cores = available_resources['cores']
-        if self.ntasks == 1:
-            # just one task so only need to worry about cpus_per_task
-            self.cpus_per_task = min(self.cpus_per_task, available_cores)
-            cores = self.cpus_per_task
-        elif self.cpus_per_task == self.min_cpus_per_task:
-            available_tasks = available_cores // self.cpus_per_task
-            self.ntasks = min(self.ntasks, available_tasks)
-            cores = self.ntasks * self.cpus_per_task
-        else:
-            raise ValueError('Unsupported step configuration in which '
-                             'ntasks > 1 and  '
-                             'cpus_per_task != min_cpus_per_task')
-
-        min_cores = self.min_cpus_per_task * self.min_tasks
-        if cores < min_cores:
+        cores_per_node = available_resources['cores_per_node']
+        self.cpus_per_task = min(self.cpus_per_task,
+                                 min(available_cores, cores_per_node))
+        if self.cpus_per_task < self.min_cpus_per_task:
             raise ValueError(
-                f'Available cores ({cores}) is below the minimum of '
-                f'{min_cores} for step {self.name}')
+                f'Available cpus_per_task ({self.cpus_per_task}) is below the '
+                f'minimum of {self.min_cpus_per_task} for step {self.name}')
+
+        available_tasks = available_cores // self.cpus_per_task
+        self.ntasks = min(self.ntasks, available_tasks)
+
+        if self.ntasks < self.min_tasks:
+            raise ValueError(
+                f'Available number of MPI tasks ({self.ntasks}) is below the '
+                f'minimum of {self.min_tasks} for step {self.name}')
 
     def setup(self):
         """


xylar:
@matthewhoffman and @trhille, I'd appreciate a review of this PR from one (and preferably both) of you if you can.  In part, I want you to let me know if you agree with the logic for constraining resources.  In part, I want to make sure you are also aware of how `ntasks`, `min_tasks`, `cpus_per_task` and `min_cpus_per_task` are used so you can help me share that knowledge with other compass developers going forward.

matthewhoffman:
@xylar , thanks for flagging us about this.  I sort of understand these concepts but am still fuzzy on some details.
 
A few thoughts/questions:
1. Is there somewhere with definitions for the concepts of cores, cpus, tasks, and threads?  Also, could those definitions and the logic being introduced in this PR be added to the docs somewhere?  Probably here: https://mpas-dev.github.io/compass/latest/developers_guide/organization.html#constrain-resources
2. How is hyperthreading handled supported?  (maybe the answer to this becomes obvious once I understand the concepts in the previous question)  On Cori, for example, with 64 logical cores per node but which supports hyperthreading two tasks per core, we would see a speedup running 128 tasks per node.  It was not as big of a speedup as running on 2 nodes with 64 tasks per node each, but it was significantly faster than using 64 tasks on one node, so we typically didn't want to leave that cost savings on the table.

xylar:
@matthewhoffman, let me know if the docs I added address your first point.

docs/developers_guide/organization.rst
```diff
@@ -1105,6 +1105,32 @@ attributes.  Another reason might be to set these attributes using an algorithm
...
+
+``ntasks``
+    The target number of MPI tasks that a step will use if the resources are
+    available.
+
+``min_tasks``
+    The minimum number of MPI tasks for a step.  If too few resources are
+    available, the step will not run.
+
+``cpus_per_task``
```
matthewhoffman, line 1119: For straight MPI applications without OpenMP, this will always be 1, right?  If so, can you add a statement to that effect?

matthewhoffman:
@xylar , thanks for the explanation - that is very helpful.  One aspect I'm still fuzzy on is how you identify or adjust the number of tasks per node.  My understanding is that is defined in the machine files, and you could override it by adjusting hyperthreading or the base value in special situations (e.g. needing to lower the number of tasks to avoid running out of memory in some memory-intensive operation) as you described above.  Is that correct?  So in test cases where we define `ntasks` we leave the tasks per node definition to the machine file.  i.e. there is not a way to prescribe the number of nodes you want to use directly - you instead define the number of tasks, and the machine definition will also be used to calculate the number of nodes required, correct?

xylar:
@matthewhoffman, my understanding is that tasks per node isn't something you get to set in Slurm.  You set the number of tasks with `--ntasks` and let it do its thing.  It decides how to distribute the tasks across the available nodes.  I believe that would typically be to have about the same number of tasks on each node, so if you ask for 5 nodes at 30 tasks, it will put 6 on each node.

Regarding memory, we'll have something more sophisticated later with task parallelism but for now we don't have a good way of saying that a job shouldn't use all the CPUs on a node because it will run out of memory that way.  I guess your best bet would be to make sure you have well over the target number of MPI tasks in that case (i.e. nodes*cores per node >> ntasks) so things work out.  But that's tricky without a separate and reliable way to specify how much memory a task needs.  Again, that's coming later (in polaris).

docs/developers_guide/organization.rst
xylar: I added this.

docs/developers_guide/organization.rst
```diff
@@ -1105,6 +1105,35 @@ attributes.  Another reason might be to set these attributes using an algorithm
...
+    If ``ntasks > 1``, this is typically a number of threads used by each
+    MPI tasks (e.g. with OpenMP threading).  If ``ntasks == 1``, this may
+    be the number of target cores used in on-node parallelism like python or
+    c++ threading, or python multiprocessing.  ``cpus_per_task`` will
+    automatically be constrained to be less that the number of cores on a node
```
trhille, line 1124:
```suggestion
    automatically be constrained to be less than or equal to the number of cores on a node
```

docs/developers_guide/organization.rst
```diff
@@ -1105,6 +1105,35 @@ attributes.  Another reason might be to set these attributes using an algorithm
...
+    The target number of MPI tasks that a step will use if the resources are
+    available.
+
+``min_tasks``
+    The minimum number of MPI tasks for a step.  If too few resources are
+    available, the step will not run.
+
+``cpus_per_task``
+    If ``ntasks > 1``, this is typically a number of threads used by each
+    MPI tasks (e.g. with OpenMP threading).  If ``ntasks == 1``, this may
```
trhille, line 1121:
```suggestion
    MPI task (e.g. with OpenMP threading).  If ``ntasks == 1``, this may
```

trhille (commented):
@xylar, this logic makes sense to me. Is there any testing you'd like us to do?

xylar:
@trhille, sorry, I thought I had said something about testing.  I would recommend running the `full_integration` suite somewhere and making sure it's fine.

trhille (approved):
All `full_integration` test cases passed on Perlmutter.

matthewhoffman:
@xylar , regarding tasks per node, we often used to set `#SBATCH --ntasks-per-node=64` to enforce hyperthreading on Cori-Haswell, for example (or 256 on Perlmutter-cpu).  What I'm still not sure about is if we wanted to use, e.g., 256 MPI tasks per node on Perlmutter (assuming it didn't have the hanging problem), would we change line 37 or 41 here?
https://github.com/MPAS-Dev/compass/blob/main/compass/machines/pm-cpu.cfg#L36-L41
```
# cores per node on the machine
cores_per_node = 128

# threads per core (set to 1 because trying to hyperthread seems to be causing
# hanging on perlmutter)
threads_per_core = 1
```

Regarding the memory issue, that's not something I'm concerned about anytime soon.

xylar:
> What I'm still not sure about is if we wanted to use, e.g., 256 MPI tasks per node on Perlmutter (assuming it didn't have the hanging problem), would we change line 37 or 41 here?

You would change line 41:
```
# cores per node on the machine
cores_per_node = 128

# threads per core (allow hyperthreading)
threads_per_core = 2
```

xylar:
Sorry, you would change both.  The `cores_per_node` is just a sanity check.  It gets computed based on what Slurm sees:
https://github.com/MPAS-Dev/compass/blob/main/compass/parallel.py#L32-L41
So you'd want:
```
# cores per node on the machine
cores_per_node = 256

# threads per core (allow hyperthreading)
threads_per_core = 2
```
It would use the `treads_per_node` you specify and it would verify that the `cores_per_node` was what you said or issue a warning if they differ.  (But it trusts Slurm over the config file in this case.)

xylar:
The reason for this is that `cores_per_node` is useful to have at setup time (e.g. for setting up job scripts) but it can only be verified at runtime.

matthewhoffman:
@xylar , thanks for the explanation.  Could you add that info about how to enable/disable hyperthreading in the docs you had added?

xylar:
Hmm, I don't think that really relates to the resources for a single step.  It's more of a machine-level thing.  I don't think it belongs in this PR.  I'd be happy to have you make a separate PR to add that elsewhere in the documentation.  I'm not sure I have time to figure out the right place at the moment.

xylar:
I think the pages to add these to would be:
https://github.com/MPAS-Dev/compass/blob/24d63ef4f3da34ecea32fa2724430dc3d0b0dbc6/docs/users_guide/machines/chicoma.rst
https://github.com/MPAS-Dev/compass/blob/24d63ef4f3da34ecea32fa2724430dc3d0b0dbc6/docs/users_guide/machines/perlmutter.rst

I can do that here but again I think it's a bit of scope creep.

xylar:
Okay, I'll open a separate PR now to do those edits.

xylar:
Added in https://github.com/MPAS-Dev/compass/pull/579

matthewhoffman (approved):
All looks good.  Thanks, @xylar , and thanks, @trhille for testing.

xylar:
Thanks @matthewhoffman!

PR #574, opened 2023-03-23
Update guidance on switching between branches

matthewhoffman:
Update to docs to add details on how conda environments are used when switching between compass branches.

Checklist

### DIFF
--- a/docs/developers_guide/quick_start.rst
+++ b/docs/developers_guide/quick_start.rst
@@ -297,9 +297,10 @@ You can do this by simply re-executing
 from the *root of the repo* before proceeding.
 
 Similarly, if you are developing or using multiple ``compass`` branches
-but you use a different directory for each (creating the directories with
- ``git worktree``), you will need to make sure the version of the
- ``compass`` package in your conda environment is the one you want.
+but you use a different directory for each
+(creating the directories with ``git worktree``),
+you will need to make sure the version of the ``compass`` package
+in your conda environment is the one you want.
 If your branches use the same ``compass`` version (so the dependencies
 are the same), you can use the same conda environment
 (e.g. ``dev_compass_<version>``) for all of them.  But you will only
--- a/docs/developers_guide/quick_start.rst
+++ b/docs/developers_guide/quick_start.rst
@@ -237,14 +237,20 @@ compilers, MPI and libraries needed to build and run MPAS components, and
 set environment variables needed for MPAS or ``compass``.  It will also set an
 environment variable ``LOAD_COMPASS_ENV`` that points to the activation script.
 ``compass`` uses this to make an symlink to the activation script called
-``load_compass_env.sh`` in the work directory.
+``load_compass_env.sh`` in the work directory.  When the load script is
+executed from the base of the compass repository (i.e., as
+``source ./load_dev_compass_<version>_<machine>_<compiler>_<mpi>.sh``),
+it will install the version of the ``compass`` package from that location into the associated
+conda environment.  When the load script is executed from the work
+directory through the symlink, it will activate the associated conda
+environment, but does *not* install the ``compass`` package into the conda
+environment; it is assumed that is already up to date from when the conda
+environment was created or last updated.
 
-If you switch between different ``compass`` branches, it is safest to rerun
-``./conda/configure_compass_env.py``  with the same arguments as above to make
-sure dependencies are up to date and the ``compass`` package points to the
-current directory.  If you are certain that no ``compass`` dependencies are
-different between branches, you can also simply source the activation script
-(``load_dev_compass*.sh``) in the branch.
+It is generally recommended to activate the ``compass`` environment (from
+either the compass repo or via the workdir symlink) from a
+clean environment.  Unexpected behavior may occur if activating a different
+``compass`` environment after having one already activated.
 
 Once you have sourced the activation script, you can run ``compass`` commands
 anywhere, and it always refers to that branch.  To find out which branch you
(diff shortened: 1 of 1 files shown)


matthewhoffman:
@xylar , this is my attempt to update the docs based on the conversation we had last week about how to switch between compass branches in different situations.  Feel free to edit/modify and push directly to my branch.

docs/developers_guide/quick_start.rst
```diff
@@ -237,14 +237,20 @@ compilers, MPI and libraries needed to build and run MPAS components, and
...
-If you switch between different ``compass`` branches, it is safest to rerun
-``./conda/configure_compass_env.py``  with the same arguments as above to make
-sure dependencies are up to date and the ``compass`` package points to the
-current directory.  If you are certain that no ``compass`` dependencies are
-different between branches, you can also simply source the activation script
-(``load_dev_compass*.sh``) in the branch.
+``load_compass_env.sh`` in the work directory.  When the load script is
+executed from the base of the compass repository (i.e., as
+``source ./load_dev_compass_<version>_<machine>_<compiler>_<mpi>.sh``),
+it will install that version of the ``compass`` repo into the associated
```
xylar, line 243:
```suggestion
it will install the version of the ``compass`` package from that location into the associated
```

docs/developers_guide/quick_start.rst
```diff
@@ -237,14 +237,20 @@ compilers, MPI and libraries needed to build and run MPAS components, and
...
+conda environment.  When the load script is executed from the work
+directory through the symlink, it will activate the associated conda
+environment, but does *not* install the ``compass`` repo into the conda
```
xylar, line 246:
```suggestion
environment, but does *not* install the ``compass`` package into the conda
```

docs/developers_guide/quick_start.rst
```diff
@@ -259,6 +265,41 @@ the branch is.  If you do not use the worktree approach, you will also need to
...
+for the entire package structure to be re-indexed.
+In those situations, it is safest to rerun
+``./conda/configure_compass_env.py``  with the same arguments as above to make
+sure dependencies are up to date and the ``compass`` package points to the
+current directory.
+
+In complex situations, it is also possible to define a user-defined conda
+environment name to avoid reusing the same conda environment with multiple
+compass branches.  This can be done with the ``--env-name`` argument to
+``./conda/configure_compass_env.py``.
```
xylar, line 301: This is not quit correct and needs some modification.  Before I do that, I need to understand the use case you have in mind.  Is this for using `git checkout` within the same directory or do you have in mind the `git worktree` approach.  I think we need a paragraph describing the right approach for each of these workflows but I want to understand what you're describing here.

It seems like there are a couple of misconceptions here.  First, you should change branches, and then reinstall `compass` (e.g. by sourcing the load script).  That would take care of any issues with files that have been added or deleted.  You don't need to rerun `./cona/configure_compass_env.py` unless dependencies need to be updated.  (It will also reinstall the `compass` package so it will take care of the case you're describing but so would just sourcing the load script after switching branches, rather than before.)

You start out mentioning switching between different `compass` environments but the workflow you describe is just installing different `compass` packages  into the same conda environment.  So I think we need to change that description.

Basically everywhere you describe the `compass` repo, I think you mean the `compass` package, which is effectively the `compass` directory within the `compass` repo (so not the docs, deployment scripts, utilities, submodules, etc.).

I'll revise this as soon as I get clarification on the workflow you want to describe here (`git checkout` vs. `git worktree`).

xylar (commented):
@matthewhoffman, I could use some clarification.  Then, I'll suggest some further revisions.

docs/developers_guide/quick_start.rst
xylar:
```suggestion
Many developers are switching between different ``compass`` branches.
We have 2 main workflows for doing this: checking out different branches
in the same directory (with ``git checkout``) or creating new directories for
each branch (with ``git worktree``).  Either way, you need to be careful that
the version of the ``compass`` package that is installed in the conda
environment you are using is the one you want.  But how to handle it
differs slightly between these workflows.

If you are developing or using multiple ``compass`` branches in the same
directory (switching between them using ``git checkout``), you will need
to make sure you update your ``compass`` environment after changing
branches.  Often the branches you're developing will make use of the 
same conda environment, because they are using the same 
``compass`` version (so the dependencies aren't changing).  The same 
conda environment (e.g. ``dev_compass_<version>``) can safely be used
with multiple branches if you explicitly reinstall the ``compass`` package
you want to use into the conda environment *after* moving a new branch.
You can do this by simply re-executing 
``source ./load_dev_compass_<version>_<machine>_<compiler>_<mpi>.sh``
from the *root of the repo* before proceeding.

Similarly, if you are developing or using multiple ``compass`` branches 
but you use a different directory for each (creating the directories with
 ``git worktree``), you will need to make sure the version of the
 ``compass`` package in your conda environment is the one you want.
If your branches use the same ``compass`` version (so the dependencies
are the same), you can use the same conda environment 
(e.g. ``dev_compass_<version>``) for all of them.  But you will only
be able to test one of them at a time.  You will tell the conda environment
which branch to use by running
``source ./load_dev_compass_<version>_<machine>_<compiler>_<mpi>.sh``
from the *root of the directory (worktree) you want to work with* before
proceeding.

In both of these workflows, you can modify the ``compass`` code and the conda
environment will notice the changes as you make them.  However, if you have 
added or removed any files during your development, you need to source the
load script again:
``source ./load_dev_compass_<version>_<machine>_<compiler>_<mpi>.sh``
in the root of the repo or worktree so that the added or removed files will be
accounted for in the conda environment.

If you know that dependencies that the ``compass`` has different dependencies
in a branch or worktree you are working on compared to a previous branch
you have worked with (or if you aren't sure), it is safest to not just reinstall
the ``compass`` package but also to check the dependencies by re-running:
``./conda/configure_compass_env.py``  with the same arguments as above.
This will also reinstall the ``compass`` package from the current directory.

If you need more than one conda environment (e.g. because you are testing
multiple branches at the same time), you can choose your own name
for the conda environment.  Typically, this might be something related to the
name of the branch you are developing.  This can be done with the 
``--env-name`` argument to ``./conda/configure_compass_env.py``.  You
can reuse the same custom-named environment across multiple branches
if that is useful.  Just remember to reinstall ``compass`` each time you
switch branches.
```

xylar: Rereading the text, I *think* you were referring more to the `git worktree` approach because you refer to "the root of the repo you want to work with".  Either way, I think my suggested revisions apply to both.  Could you give this a read and incorporate it in some form?

Thanks so much for getting the ball rolling on this!

matthewhoffman: @xylar , thanks for editing my description.  I was indeed primarily thinking about the worktree workflow, but I was trying to write the description in a workflow-agnostic way, because I thought the steps did not really change.  What you have written here helps clarify the subtleties and fixes some issues I was confused about. 

Note there are typos in lines 294 and 320, but github won't let me make suggestions on a suggestion so I'll commit your change and then fix them after.

matthewhoffman:
@xylar , I've incorporated the changes you made/suggested.  Give this enough look when you have a chance.  I'll build the docs and confirm they look correct if you are happy with the content.

xylar:
Yep, this looks good!  Please do a local build (just in case) and I'll merge.

matthewhoffman:
@xylar , I checked and fixed the formatting, so this is ready to go from my perspective.

xylar (approved):

xylar:
Thanks @matthewhoffman!

PR #575, opened 2023-03-23
Add WCwISC14 and ARRMwISC12to60 meshes

xylar:
Checklist


xylar:
I am currently running these new meshe through their paces here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230323/arrmwisc
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230323/wcwisc14
```

xylar:
@milenaveneziani and @vanroekel, these are not even close to the E3SM v3 versions of these meshes yet, I'm just adding the support for creating versions of each mesh with ice-shelf cavities and making sure they "work".  Once my test are done (hopefully tomorrow), I'll ask you to have a quick glance at the meshes and any other pieces of the compass process you want.  This doesn't have to be a super thorough review since we'll add some more things based on https://github.com/MPAS-Dev/compass/discussions/565 before we do more thorough testing.

xylar:
@milenaveneziani, the ARRMwISC12to60 is ready to review.  While the `files_for_e3sm` step hasn't been run, I don't think it matters since this isn't the production version of the mesh anyway.  I just want to make sure things look okay.

I'd suggest looking at the initial condition (maybe using the paraview extractor):
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230323/arrmwisc/ocean/global_ocean/ARRMwISC10to60/PHC/init/initial_state
```
I mostly just want your sense that nothing looks crazy.

xylar:
@vanroekel, same for you for the WCwISC14.  The `files_for_e3sm` step started but didn't complete (because making balanced sea-ice partition files is pretty time consuming).  I don't think it matters since this isn't the production version of the mesh anyway.  I just want to make sure things look okay.

Again, I'd suggest looking at the initial condition (maybe using the paraview extractor):
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230323/wcwisc14/ocean/global_ocean/WCwISC14/PHC/init/initial_state
```
I mostly just want your sense that nothing looks crazy.

vanroekel:
@xylar I took a look through that file you pointed me at in paraview and everything looks reasonable to me.  I can see the WC14 enhancement, T&S appear in bounds throughout the column and can see the ISC regions in the south.  Let me know if you want me to dump in any paraview screenshots, would be happy to do so.

xylar:
@vanroekel, would you be willing to "officially" approve the PR based on the testing you've already done?

milenaveneziani:
I checked the initial state for ARRMwISC10to60 with paraview and everything looks as expected (I also compared `bottomDepth` to the bathymetry for the ARRM mesh without ice cavities, and they are similar). I checked `botttomDepth`, `layerthickness`, `temperature`, and `salinity`. Nothing out of the ordinary. 
Ice cavities are there where they should be.

milenaveneziani:
One thing I wanted to report for record keeping that has nothing to do with this particular mesh, but likely with the PHC data that we use to produce the initial condition, is this small warm pool in the Northwestern Passages just north of Hudson Bay in the middle of winter:

Screen Shot 2023-03-28 at 9 37 56 AM

Just something to be aware of, and pinging @proteanplanet here as well.

milenaveneziani (approved):
Approved based on viewing of the initial_state file.

dengwirda:
@milenaveneziani, I've seen that warm blob in lots of config.'s I believe, and assume it's coming from the underlying data/interpolation somehow.
It doesn't look to me that the raw [PHC SST climatology](http://psc.apl.washington.edu/nonwp_projects/PHC/AnnualGraphicsP3.html) (assuming this is the right one) has the same feature though, which feels a bit weird...
I assume this only affects the spin-up, but starting from a better climatology seems a good thing to improve.

milenaveneziani:
Yes, I agree @dengwirda. Best would be to take a look at the PHC files that compass uses to generate the initial condition. The blob is not only warm but also fresh btw. @xylar: could you point me to the original PHC files?

We do use WOA18 climatologies in MPAS-Analysis, so we do have that data ready, I think, if we decide to give it a try.

xylar:
>  @xylar: could you point me to the original PHC files?

@milenaveneziani, here are the files we use.  @maltrud produced them as far as I'm aware so he might have some thoughts.

https://web.lcrc.anl.gov/public/e3sm/mpas_standalonedata/mpas-ocean/initial_condition_database/PotentialTemperature.01.filled.60levels.PHC.151106.nc
https://web.lcrc.anl.gov/public/e3sm/mpas_standalonedata/mpas-ocean/initial_condition_database/Salinity.01.filled.60levels.PHC.151106.nc

maltrud:
I'll have a look at the original vs filled files.

maltrud:
that weird blob is in the original PHC2.0 data.  I didn't realize this was still being used (it's from 15 years ago!).  I switched to PHC 3.0 when it came out.  I'd say we should switch to WOA (2018?).
Screen Shot 2023-03-30 at 11 33 32 AM

milenaveneziani:
I thought that might be the case. Agreed about switching to WOA18.

PR #577, opened 2023-03-27
Add missing land-ice mask file to cull mesh output 

xylar:
This missing output means that the cached mesh steps with ice-shelf cavities don't work with non-cached init steps.

Mesh and init cache files are updated here, too, to include the missing `land_ice_mask.nc`.

Checklist


xylar:
## Testing

The `pr` test suite passes and is BFB with the baseline except for the newly cached QUwISC and ECwISC tests, which have a different number of cells (possibly because of updated geometric features) than the previous cached files.

PR #582, opened 2023-03-29
Switch wisc meshes to z-star

xylar:
Checklist


cbegeman (approved):
Looks consistent with what we talked about. Would you like me to run some tests?

xylar:
@cbegeman, that would be wonderful!  If you can run the `ECwISC30to60` `mesh`, `init` and `performance` tests, that would be a really great starting point.

xylar:
If that goes well, we would also want to run `dynamic_adjustment`.  That'll be a more rigorous test.  Then, we'd probably turn on melting and run the `simulation` step for longer.

xylar:
It could be that `init` fails.  I can't recall if `global_ocean` in init mode is prepared for that coordinate.  I hope so!

cbegeman:
@xylar The model crashes (invalid state) at the second time step of ssh_adjustment when dt=30s

xylar:
Thanks.  Not promising, it sounds like.  Presumably, waves are big enough to evacuate thin water columns.

cbegeman:
@xylar I just reran the case with wetting and drying turned on and it doesn't appear that thin layers are actually evacuated. The `config_verify_not_dry` functionality should have caught that.

xylar:
Thanks @cbegeman.  I'll need to revisit this, then.  I think I will prioritize other changes (like bathymetry, initial condition and vertical grid) first since they may affect this behavior.

xylar:
I'm going to close this.  I don't think we will be able to switch to the z-star vertical coordinate here on compass, and we should focus on doing that as part of the move to polaris this fall.

PR #583, opened 2023-03-29
Remove Ifpack and ML preconditioners from landice test inputs

mcarlson801:
This PR removes optional/unused preconditioners from landice test inputs that use Trilinos packages (Ifpack/ML) that we no longer use for our builds. Having these optional preconditioners cause invalid parameter exceptions when running landice tests using newest Albany builds. These tests don't use the Ifpack/ML preconditioners anyway so this won't change the behavior of these tests.

Checklist

### DIFF
--- a/compass/landice/tests/circular_shelf/albany_input.yaml
+++ b/compass/landice/tests/circular_shelf/albany_input.yaml
@@ -90,12 +90,6 @@ ANONYMOUS:
 #             Preconditioner Information
               Preconditioner Type: Ifpack2
               Preconditioner Types:
-                Ifpack:
-                  Overlap: 1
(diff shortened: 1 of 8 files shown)


mcarlson801:
### Testing

Confirmed that landice tests now pass on Perlmutter with latest version of Albany and Trilinos-develop. Attached test log:
[biweekly_log_pm_cpu_Compass_test.txt](https://github.com/MPAS-Dev/compass/files/11102683/biweekly_log_pm_cpu_Compass_test.txt)

matthewhoffman (approved):
@mcarlson801 , this looks great.  Thanks for taking care of it and opening the PR.

Issue #593, opened 2023-04-11
Topography remapping PR broke several tests


xylar:
The topography remapping PR (#566) didn't get adequate testing in situations where it is *not* used and breaks some tests as a result:
* We need to cache new mesh and initial condition steps for EC (and maybe SO) meshes, because the current cached files don't include the bathymetry remapping
*  the QU(wISC)240 `culled_mesh` steps are attempting to read a non-existent `topography.nc` file.

PR #594, opened 2023-04-11
Update cached EC mesh and init files

xylar:
These need to be updated after recent changes to topography remapping and vertical grid.

Checklist

Partially addresses #593


xylar:
## Testing

The `pr` suite runs again on Chrysalis with this fix (except for the broken test `ocean/global_ocean/QU240/mesh` fixed in #596).

PR #596, opened 2023-04-11
Only cull topography for cull_mesh steps that have it

### LINKED ISSUE #593: Topography remapping PR broke several tests

xylar:
Some `cull_mesh` steps take a remapped topography as an input, while others do not (for added speed and to keep resources under control).  This merge fixes a bug in which the `cull_mesh` step was trying to cull the topogrpahy file even if none is present.

Checklist

closes #593

### DIFF
--- a/compass/ocean/mesh/cull.py
+++ b/compass/ocean/mesh/cull.py
@@ -383,7 +383,8 @@ def _cull_mesh_with_logging(logger, with_cavities, with_critical_passages,
                 '--engine', netcdf_engine]
         check_call(args, logger=logger)
 
-    _cull_topo()
+    if has_remapped_topo:
+        _cull_topo()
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the `pr` test suite on Chrysalis with this branch and the `ocean/global_ocean/QU240/mesh/cull_mesh` step now runs successfully (as does the rest of the suite).

PR #599, opened 2023-04-12
Reduce no. of processes for QU240 cull_mesh step

### LINKED ISSUE #598: QU240 `cull_mesh` performance worse with more processes
I increased the number of python processes used for mesh culling in https://github.com/MPAS-Dev/compass/pull/566 from 18 to 128.  This actually seems to have made performance worse for the QU240 mesh, so this parameter needs to be mesh dependent.

xylar:
It appears that 128 processes are too many and slow down the step significantly.

Checklist

closes #598


xylar:
# Testing

I ran the `pr` suite on Perlmutter.  The `ocean/global_ocean/QU240/mesh` took 2 minutes with this change, whereas it was taking over 8 minutes with 128 processes.

PR #600, opened 2023-04-12
Change np.bool to bool

mark-petersen:
Without this change, the test `ocean/tides/Icos7/init` fails. The error message says that  `np.bool` is no longer supported. With this change, the test case runs successfully on perlmutter using `load_dev_compass_1.2.0-alpha.4_pm-cpu_gnu_mpich.sh`.

The flake linter changed the order of the import lines.

### DIFF
--- a/compass/ocean/tests/tides/init/dem_pixel.py
+++ b/compass/ocean/tests/tides/init/dem_pixel.py
@@ -1,9 +1,10 @@
 
-import os
-import numpy as np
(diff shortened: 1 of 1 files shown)


xylar (approved):
Thanks @mark-petersen!  This looks correct to me and I know from other work I've done that `numpy` recently dropped its redundant references to python datatypes like `bool`, `int` and `float`.

PR #590, opened 2023-04-07
Refactor land-ice mesh generation test cases

trhille:
Move build_cell_width() to compass.landice.mesh and create new build_MALI_mesh() method to greatly reduce code redundancy between all mesh-generating test cases.

Also add make_region_masks() to compass.landice.mesh and add region masks for Greenland.

Checklist
* [x] User's Guide has been updated
* [x] Developer's Guide has been updated
* [x] API documentation in the Developer's Guide (`api.rst`) has any new or modified class, method and/or functions listed
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #602, opened 2023-04-14
Add isunnguata sermia mesh generation test case

trhille:
Add a new `issunguata_sermia` test case with a `mesh_gen` test group.

Checklist


trhille:
force-push from [1bddfd9 ](https://github.com/MPAS-Dev/compass/commit/1bddfd97669b4879933ad63ea9e25ca5d17736e4)to [728cc4a ](https://github.com/MPAS-Dev/compass/commit/728cc4ad89c43d7d0188efa5c359d578eb3dd96c) was a rebase onto `compass/main` to incorporate changes after merge of https://github.com/MPAS-Dev/compass/pull/590.

trhille:
Testing
-------
Running `mesh_gen` created the mesh without issues. It's closer to 0.8–8 km than 1–10km, but at only 3900 cells this doesn't seem like a major issue.

image

matthewhoffman (changes requested):
This looks pretty standard, so I don't have any concerns with the implementation.

Two observations:
* Official name of this glacier has two n's and one s: Isunnguata Sermia (https://tc.copernicus.org/articles/9/2215/2015/tc-9-2215-2015.pdf)
* The cell spacing near the terminus looks strange and not similar to the velocity field.  Do you know why that is?

I'm happy to entertain your opinion about to what extent either of these items should be corrected.

trhille:
Thanks @matthewhoffman. I'm happy to change the name of the test case. Good catch on the velocity-cell spacing mismatch. I believe this stems from the 8km dataset that is used to set cell spacing, while the 1km dataset is used to populate the velocity field after the mesh has been created and culled. I notice that this also leads to an issue with ice thickness abutting the mesh boundary, so I'll fix that.

trhille:
Force-push was a rebase onto `origin/main` in order to get new python cell culler.
Changing from the 8km to 2 km data set to define cell width results in much better cell spacing:
image

Note: the `gridSpacing` variable is now identical to the `cellQuality` variable. This seems wrong.

xylar:
> Note: the gridSpacing variable is now identical to the cellQuality variable. This seems like a bug introduced elsewhere in COMPASS.

That would probably be in MPAS-Tools in the cell culler?  If you want to follow up, that would be appreciated.  But I also don't think we use the cell quality for anything...

matthewhoffman:
@trhille , this looks ready to merge with exception of correcting the glacier spelling (which is a lot of changes in file names, variables, docs, etc.).  Do you want to do that or leave it as is?  My preference would be to fix it, and I'm happy to help do that.

trhille:
@matthewhoffman, thanks for the reminder. I'll fix that.

trhille:
Note: requires https://github.com/MPAS-Dev/MPAS-Tools/pull/551 to pass.

matthewhoffman (approved):
Thanks for the updates, @trhille !  I confirmed I can reproduce the same result.

image

PR #603, opened 2023-04-14
Add job script to steps

xylar:
It is often convenient to run steps on their own, so it will be convenient to have a job script (and a link to the load script) for each step.

Checklist


xylar:
@matthewhoffman, @trhille and @cbegeman, just making you aware of this change.

xylar:
## Testing

I set up the `pr` test suite  on Chrysalis and verified that job scripts were created in each of several steps I examined.  I successfully ran the job scripts for the `ocean/baroclinic_channel/10km/default/initial_state` and `.../forward` steps one after the other.  I verified that the number of nodes in the job scripts for the `ocean/global_convergence/cosine_bell/QU60/mesh`, `.../init` and `.../forward` steps were correct (1, 1 and 2, respectively).

I then successfully ran the `pr` suite.

PR #605, opened 2023-04-15
Adjust #cells in baroclinic channel to define 160x500km domain

dengwirda:
This PR adjusts the number of cells used in meshes for the baroclinic channel test case to ensure the 'standard' 160x500km domain is defined. Previously, the channel has had a shorter y-length of sqrt(3)/2 * 500 => 433km.

- When using the `make_planar_hex_mesh` routine, the number of cells along the y-axis should be scaled by `2/sqrt(3)` to account for hexagonal geometry.

The channel is a standard community test case, and use of the 160x500km domain enables model intercomparison, e.g.:

- Ilıcak, M., Adcroft, A.J., Griffies, S.M. and Hallberg, R.W., 2012. Spurious dianeutral mixing and the role of momentum closure. Ocean Modelling, 45, pp.37-58.

- Petersen, M.R., Jacobsen, D.W., Ringler, T.D., Hecht, M.W. and Maltrud, M.E., 2015. Evaluation of the arbitrary Lagrangian–Eulerian vertical coordinate method in the MPAS-Ocean model. Ocean Modelling, 86, pp.93-113.

### DIFF
--- a/compass/ocean/tests/baroclinic_channel/__init__.py
+++ b/compass/ocean/tests/baroclinic_channel/__init__.py
@@ -44,13 +44,13 @@ def configure(resolution, config):
         Configuration options for this test case
     """
     res_params = {'10km': {'nx': 16,
-                           'ny': 50,
+                           'ny': 58,  # NB. 2/sqrt(3) due to hexagons
                            'dc': 10e3},
                   '4km': {'nx': 40,
-                          'ny': 126,
+                          'ny': 144,
                           'dc': 4e3},
                   '1km': {'nx': 160,
-                          'ny': 500,
+                          'ny': 578,
                           'dc': 1e3}}
 
     comment = {'nx': 'the number of mesh cells in the x direction',
--- a/compass/ocean/tests/baroclinic_channel/__init__.py
+++ b/compass/ocean/tests/baroclinic_channel/__init__.py
@@ -1,9 +1,9 @@
-from compass.testgroup import TestGroup
 from compass.ocean.tests.baroclinic_channel.decomp_test import DecompTest
 from compass.ocean.tests.baroclinic_channel.default import Default
 from compass.ocean.tests.baroclinic_channel.restart_test import RestartTest
 from compass.ocean.tests.baroclinic_channel.rpe_test import RpeTest
 from compass.ocean.tests.baroclinic_channel.threads_test import ThreadsTest
+from compass.testgroup import TestGroup
 
 
 class BaroclinicChannel(TestGroup):


xylar:
Thanks @dengwirda.  I totally agree that this change is needed.  Could you describe any testing you did of this branch?

xylar:
If you don't mind, I will push a commit that fixes the linting issue (unrelated to your changes) that that CI is pointing out.

dengwirda:
Thanks @xylar. I've run the 1km and 4km cases out to 100 days on `pm-cpu` --- @alicebarthel and I have been looking into whether MPAS-O produces the same long timescale dynamics as other models, and cleaning up the channel geometry is a first step on this. Based on some quick code/visual checks, it looks like the ICs, etc are correctly scaled onto the 500km geometry, and that the spin-down proceeds as expected.

xylar:
Sounds good.  Do you mind me pushing that linting commit I mentioned above?

xylar:
## Testing

I have run all the `baroclinic_channel` test cases except `1km/rpe_test` successfully (that last one takes a long time so it's still running).

I verified that the maximum of `yCell` as close to 500 km in all 3 mesh configurations (whereas I agree that it was close to 440 km in the configurations before this change).

dengwirda:
@xylar yes please do push to clean up the linting --- I'm not 100% on what needs to change there.
One additional note re: the `make_planar_hex_mesh` routine is that it culls off rows of cells when you ask for non-periodic boundaries, making the min(coord) value not quite 0.0, and so requires a little fiddling to ensure the overall domain size is as expected.

xylar:
> One additional note re: the make_planar_hex_mesh routine is that it culls off rows of cells when you ask for non-periodic boundaries, making the min(coord) value not quite 0.0, and so requires a little fiddling to ensure the overall domain size is as expected.

Ah, yes, of course.  I was a little worried that the max of `yCell` seemed a bit larger than it strictly needed to be but the deleted first and last rows would account for that.  Yes, agreed, it is tricky.

Also, what do you consider the location of the top and bottom boundaries, given that they are jagged?

xylar (approved):
I ran all `baroclinic_channel` tests (including all 3 RPE tests for 20 days) on Chrysalis.  All ran successfully.  I didn't look at the RPE results in detail but they can be found at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230415/bcn_chan_fix_ny
```
if anyone wants to have a look.

dengwirda:
@xylar I agree it's hard to get an exact match with the `make_planar_hex_mesh` tools, due to jagged-ness and integer+even-numbered-regular-hexagon constraints. I think as long as `max(yCell)-min(yCell) `and `max(xCell)-min(xCell)` are within `dc` of the desired bounds this should be close enough in practice, and should converge as `dc => 0`.

@alicebarthel and all, it looks like running with the proper length channel results in a little more cool surface water remaining at 200 days (for `dc=1km, nu=20`):

mpaso_bcl_channel_160x500km_day200

which improves on the 'short' MPAS-O channel:

baroclinic_channel_SST

but is still somewhat different to the MITgcm and ROMS results shown in Ilicak et al.

alicebarthel (approved):
Thanks @dengwirda, the changes look good to me by visual inspection. Thanks for pursuing this, for some reason I only had the x-geometry in mind when we first talked about this. 
Happy to approve as is, but let me know if you'd like me to run some additional tests.

xylar:
Thanks @dengwirda for fixing this and @alicebarthel for the review!

PR #615, opened 2023-04-21
Fix substitution of `levels` at runtime in `global_ocean`

### LINKED ISSUE #614: Mesh metadata not including number of vertical levels correctly
The mesh description is ending up with `... with autodetect vertical level`, where `autodetect` is supposed to have been replaced with the number of vertical levels.

xylar:
We need a different placeholder in config options for situations where we need to replace them (not using extended interpolation in the config file) at runtime.  Currently, this only applies to `levels`, the number of vertical levels but it could apply to `creation_date` and `bottom_depth` so we support these in the same way.

Checklist

closes #614


xylar:
## Testing

I ran `mesh` and `init` test cases for the QUwISC240 mesh and with this change, I see the expected mesh description in the metadata:
```
		:MPAS_Mesh_Description = "MPAS quasi-uniform mesh for E3SM version 3 at 240-km global resolution with 16 vertical level" ;

```

PR #616, opened 2023-04-22
Create ensemble of branch runs for projections

matthewhoffman:
This PR introduces a branch_ensemble test within the ensemble_generator test group that can be used to set up an ensemble of branch runs from the spinup ensemble for the purpose of running projections

Checklist


compass/landice/tests/ensemble_generator/branch_ensemble/branch_run.py
```diff
@@ -0,0 +1,184 @@
...
+        self.config.set('job', 'job_name', f'uq_{self.name}')
+        machine = self.config.get('deploy', 'machine')
+        pre_run_cmd = ('LOGDIR=previous_logs_`date +"%Y-%m-%d_%H-%M-%S"`;'
+                       'mkdir $LOGDIR; cp log* $LOGDIR; date')
+        post_run_cmd = "date"
+        write_job_script(self.config, machine,
+                         target_cores=self.ntasks, min_cores=self.min_tasks,
+                         work_dir=self.work_dir,
+                         pre_run_commands=pre_run_cmd,
+                         post_run_commands=post_run_cmd)
```
matthewhoffman, line 167: @xylar , when setting up ensembles, we ideally would be able to customize the job script to have the name of the run, as well as these pre and post run commands.  However, because `compass setup` writes each step's job script (https://github.com/MPAS-Dev/compass/blob/main/compass/setup.py#L315) , **after** the step has been set up (https://github.com/MPAS-Dev/compass/blob/main/compass/setup.py#L295), whatever job script gets written here in the step's `setup` method gets overwritten.  This means that the pre and post commands here get lost, and all ensemble members have the job name of the last processed ensemble member (e.g. when submitting a 100 run ensemble, in looking at squeue, all jobs have the name `run100`)
The ensemble_generator does not follow normal compass protocols, so maybe there is not a good workaround to this situation, but I was wondering if you had any ideas.

The ideas I have considered are:
1. somehow being able to disable the writing of job script for a step (not sure how that would be controlled), so that the clobbering wouldn't happen
2. being able to have a step's `setup` pass back arguments that can optionally be used by `write_job_script`
3. have a 'post-setup' method in a step that `compass setup` calls at the very end of `setup_case` where the job script could be overwritten

All of these solutions seem pretty awkward and invasive to solve a very specific situation.  Can you think of any other ideas?  It's not crucial that the job scripts get customized like this, but it does help a lot with managing large ensembles.

matthewhoffman:
Compiled docs here: 
https://portal.nersc.gov/cfs/fanssie/hoffman2/compass-docs-branch-ensemble/

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -182,11 +270,31 @@
...
+# --------------
+# Get needed fields from input file
+# --------------
+
+DSinput = xr.open_dataset(input_file_path)
+observedSurfaceVelocityX = DSinput['observedSurfaceVelocityX'].values[0, :]
+observedSurfaceVelocityY = DSinput['observedSurfaceVelocityY'].values[0, :]
+obsSpdUnc = DSinput['observedSurfaceVelocityUncertainty'].values[0, :]
+obsSpd = (observedSurfaceVelocityX**2 + observedSurfaceVelocityY**2)**0.5
+obsSpdUnc += obsSpd * 0.25
```
trhille, line 282: Do we actually want this hard-coded?

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -202,61 +310,149 @@
...
-        if GLobs is not None and len(indices) > 0:
+        iceArea = f.variables['totalIceArea'][:] / 1000.0**2  # in km^2
+        grdVol = f.variables['groundedIceVolume'][:] / (1.0e12 / rhoi)  # in Gt
+        BMB = f.variables['totalFloatingBasalMassBal'][:] / -1.0e12  # in Gt/yr
+        grSMB = f.variables['totalGroundedSfcMassBal'][:] / -1.0e12  # in Gt/yr
+        groundingLineFlux = f.variables['groundingLineFlux'][:] \
+            / 1.0e12  # Gt/yr
+        GLMigFlux = f.variables['groundingLineMigrationFlux'][:] \
+            / 1.0e12  # Gt/yr
+        w = 50
```
trhille, line 332: What's this?

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -202,61 +310,149 @@
...
+        GLMigFlux2 = np.convolve(GLMigFlux, np.ones(w), 'same') / w
```
trhille, line 333: No idea what's going on here. Needs a comment.

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -202,61 +310,149 @@
...
-            qoi_info['total area']['values'][idx] = iceArea[ii] - iceArea[0]
-
-            qoi_info['GL flux']['values'][idx] = groundingLineFlux[ii]
+            surfaceSpeed = DS['surfaceSpeed'].values[ii, :]
+            thickness = DS['thickness'].values[ii, :]
+            cellMask = DS['cellMask'].values[ii, :]
+            areaCell = DS['areaCell'].values[0, :]
+
+            # only evaluate where modeled ice exists and observed speed
+            # uncertainy is a reasonable value (<~1e6 m/yr)
```
trhille, line 389: is 1e6 m/yr a reasonable value?

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -202,61 +310,149 @@
...
+            # plot time series
+            # plot 1
+            ax_ts_vaf.plot(years, VAF - VAF[0], linewidth=lw, color=col,
+                           alpha=alph)
+            if label_runs:
+                ax_ts_vaf.annotate(run, (years[-1], VAF[-1] - VAF[0]),
+                                   fontsize=6)
+            ax_ts_grvol.plot(years, grdVol - grdVol[0], linewidth=lw,
+                             color=col, alpha=alph)
+            if label_runs:
```
trhille, line 427: Probably better to do all the plot() commands first, and then only have one `if label_runs:` with all the annotate() commands inside.

trhille (changes requested):
@matthewhoffman, this is excellent! I have some minor comments, but it's pretty much ready to go.

compass/landice/tests/ensemble_generator/plot_ensemble.py
```diff
@@ -202,61 +310,149 @@
...
+        grdVol = f.variables['groundedIceVolume'][:] / (1.0e12 / rhoi)  # in Gt
+        BMB = f.variables['totalFloatingBasalMassBal'][:] / -1.0e12  # in Gt/yr
+        grSMB = f.variables['totalGroundedSfcMassBal'][:] / -1.0e12  # in Gt/yr
+        groundingLineFlux = f.variables['groundingLineFlux'][:] \
+            / 1.0e12  # Gt/yr
+        GLMigFlux = f.variables['groundingLineMigrationFlux'][:] \
+            / 1.0e12  # Gt/yr
+        w = 50
+        GLMigFlux2 = np.convolve(GLMigFlux, np.ones(w), 'same') / w
+        GLflux2 = groundingLineFlux + GLMigFlux2
```
alexolinhager, line 334: This line gives an error when I run it: `ValueError: operands could not be broadcast together with shapes (42,) (50,)`. It seems like the hardcording of `w=50` in line 332 is causing issues because `groundingLineFlux` has different length in this instance

compass/landice/tests/ensemble_generator/plot_ensemble.py
alexolinhager, line 332: Should this be `len(groundingLineFlux)` so that `groundingLineFlux` and GLMigFlux2` are the same size in line 334?

alexolinhager (commented):
The ensemble runs successfully but it looks like there is a bug in `plot_ensemble.py`. May need to make a change there before merging

compass/landice/tests/ensemble_generator/ensemble_generator.cfg
```diff
@@ -27,15 +27,15 @@ sampling_method = uniform
 # max_samples should not be changed after the first set of ensemble.
 # So, when using Sobol sequence, max_samples might be set larger than
 # (end_run + 1) if you plan to add more samples to the ensemble later.
-max_samples = 4
+max_samples = 1024
```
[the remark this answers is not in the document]
matthewhoffman: This is not the number of samples that will be run, just the largest possible sample size in the Sobol sequence.  It's set this way to allow you to add more samples (up to 1024) later.  I think that if you were to change this number after you've set up the first phase of an ensemble that you would not get the same sequence for the lower numbers even with the same seed (but I'm not positive), so I set it to a number larger than I anticipate ever running an ensemble with.  My understanding is there is no problem with setting the max samples of a Sobol sequence larger than the number of samples you actually use (but not positive about that either).

compass/landice/tests/ensemble_generator/branch_ensemble/branch_run.py
matthewhoffman: I think the best solution is to create test-case-specific job script generator that uses a special job script name and then have the ensemble_manager use that.  I will save that for a different PR.

compass/landice/tests/ensemble_generator/branch_ensemble/streams.landice
```diff
@@ -0,0 +1,101 @@
...
+    <var name="ismip6shelfMelt_TFdraft"/>
+    <var name="allowableDtACFL"/>
+    <var name="allowableDtDCFL"/>
+    <var name="calvingCFLdt"/>
+    <var name="dtCalvingCFLratio"/>
+    <var name="totalRatebasedCalvedVolume"/>
+    <var name="totalRatebasedUncalvedVolume"/>
+</stream>
+
+<stream name="globalStats"
```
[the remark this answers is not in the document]
matthewhoffman: I agree, but for this PR I'd like to preserve the config I used for the Amery ensemble.  I'm thinking through designs for a future update that generalizes the handling of MALI input files to preserve each ensemble we work on without having to clobber the previous settings.  So I'd like to leave the namelist/streams/yaml settings in this PR as is, and make any fixes or generalizations in a follow on PR

compass/landice/tests/ensemble_generator/branch_ensemble/streams.landice
```diff
@@ -0,0 +1,101 @@
...
+        type="output"
+        filename_template="output/globalStats.nc"
+        filename_interval="none"
+        clobber_mode="overwrite"
+        output_interval="0000-00-00_01:00:00"
+        reference_time="2015-01-01_00:00:00"
```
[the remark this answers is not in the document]
matthewhoffman: Good catch.  Shouldn't have a practical impact but better to make consistent.  For this PR I'd like to leave as-is per previous comment.

compass/landice/tests/ensemble_generator/ensemble_generator.cfg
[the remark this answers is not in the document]
matthewhoffman: Yes, see previous comment.  I'll update the cfg in the user's guide and add a note how this is deliberately specific to the Amery ensemble but should be made more general in the future.

compass/landice/tests/ensemble_generator/ensemble_generator.cfg
```diff
@@ -18,7 +18,7 @@ end_run = 3
 # for a single parameter sensitivity study.  It will sample uniformly across
 # all dimensions simultaneously, thus sampling only a small fraction of
 # parameter space
-sampling_method = uniform
+sampling_method = sobol
 
 # maximum number of sample considered.
 # max_samples needs to be greater or equal to (end_run + 1)
```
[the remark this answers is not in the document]
matthewhoffman: Thanks for pointing that out - I've updated the comment to state that.

compass/landice/tests/ensemble_generator/ensemble_generator.cfg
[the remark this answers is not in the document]
matthewhoffman: I've reworded the comment slightly to hopefully better reflect this

compass/landice/tests/ensemble_generator/ensemble_manager.py
```diff
@@ -63,8 +65,24 @@ def run(self):
             # Get step object from 'steps' dictionary
             runStep = self.test_case.steps[run]
             os.chdir(runStep.work_dir)
-            # TODO: assess if this run is unrun, partially run, or complete,
-            # and adjust accordingly
+
+            # Skip run if an error is detected
+            err_files = glob.glob('log.landice.*.err')
```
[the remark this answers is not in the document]
matthewhoffman: I agree their is some fragility here but I wanted to be conservative so that if there are any issues with a run it doesn't get automatically relaunched.  A user could inspect that run and deal with it.

In general one shouldn't change code versions in an ensemble.  I've added a note to the users guide pointing that out.

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: Good catch.  This was an exploration of considering larger speed uncertainty.  I've commented it out with a note.

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: I've added comments explaining this section

matthewhoffman: w is the width of a boxcar filter applied to smooth this time series.  I've added a comment explaining that.

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: added

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: Yeah - I think this just happens when the the number of time levels in a run is less than `w`.  This smoothing operation was a bit of hack to get decent looking GL flux time series plotted.  It could use further improvement, but I'd rather not dig into it in this PR.

We could have `w` shorten to the length of the time-series if needed, but then different runs might have different levels of smoothing applied.  I've more comments about this issue in the script, so if this error occurs and a user looks at the script where the error occurred, the comments should guide them how to resolve it.

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: That was screening out some no data values.  Improving on this robustly would best include handling no data values in netcdf files and in the code explicitly.

compass/landice/tests/ensemble_generator/plot_ensemble.py
matthewhoffman: I liked having the annotation with the corresponding chunk of code making a plot.  There are only a couple of these run labeling events so that seems not too messy.

matthewhoffman:
@trhille and @alexolinhager , thanks for the careful review of this PR,  including wading through the messy plotting script.  I've responded to review comments and pushed a commit with fixes.  I requested a re-review from Trevor to confirm revisions are adequate, but Alex, feel free to look over again if you'd like.

trhille (changes requested):
Thanks for addressing and answering my comments, @matthewhoffman. Approved based on your responses to the comments, but I haven't tested.

trhille (approved):
Whoops, hit the wrong button last time. Approved.

PR #617, opened 2023-04-22
Add land-ice floating variables to `ocean_mesh` step

xylar:
This likely got missed because `ocean_mesh` got added to `files_for_e3sm` since the floating variable branch was created (several months ago).

Checklist


xylar:
## Testing

I tested these changes (not strictly from this branch but the same changes) on the `files_for_e3sm` step applied to the existing `oQU240wLI` mesh from E3SM.  The variables were added in the `ocean_mesh` step as expected.

PR #619, opened 2023-04-25
Fixup namelists for single layer RK4 ocean tests

cbegeman:
This PR sets `config_use_debugTracers=.false.` for all `dam_break` and `drying_slope` tests. This isn't strictly necessary for the multi-layer cases but there is no need to use `debugTracers` in any of these cases.


cbegeman:
### Testing

I have tested `ocean/dam_break/120cm/default` and `ocean/drying_slope/1km/single_layer/default` on compy with intel, impi.

xylar (approved):
Definitely looks fine to me!

cbegeman:
@sbrus89 I'm going to go ahead and merge so you can use https://github.com/MPAS-Dev/compass/pull/620 for testing https://github.com/E3SM-Project/E3SM/pull/5590 with this commit included.

PR #610, opened 2023-04-20
Add WOA23 initial condition

xylar:
This merge adds the WOA 2023 climatology from 1991-2020 as the default initial condition in the global ocean test group.

The WOA23 data was processed using #604.

Checklist
* [x] User's Guide has been updated
* [x] Developer's Guide has been updated
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes
* [x] New tests have been added to a test suite

PR #621, opened 2023-04-26
Update global ocean init cache files

xylar:
These need updating following #610 

Checklist


xylar:
## Testing

I ran the `pr` suite on Chrysalis with Intel using the cache files this points to.  The suite was successful, whereas I wasn't able to set it up without these changes.

PR #551, opened 2023-03-06
Cleanup ocean bottom drag namelist options

cbegeman:
This PR updates compass in accordance with https://github.com/E3SM-Ocean-Discussion/E3SM/pull/41. It results in different behavior for the QU240 test cases: explicit drag is no longer used and constant-coefficient implicit drag is retained (the default). 

Checklist
* [X] The `E3SM-Project` submodule has been updated with relevant E3SM changes
* [X] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #626, opened 2023-05-02
Add double quotes to flag

mark-petersen:
In python, string flag specifications require both single and double quotes. This fixes the recent `config_implicit_bottom_drag_type` addition. This broke all the `dynamic_adjustment` cases, and specifically `ocean/global_ocean/QU240/WOA23/dynamic_adjustment` in the `pr` suite. This works now.


mark-petersen:
@xylar do you prefer to be the person to merge all the compass PRs? Or do you want me to merge little ones like this after they are approved?

xylar:
@mark-petersen, you can merge once @cbegeman has approved.  

Sorry I missed this in the testing I did on #551.  There were expected non-bit-for-bit but I guess the actual test failures snuck past me in the process.

cbegeman (approved):
@mark-petersen Thank you for catching this and fixing it!

PR #628, opened 2023-05-09
Add more graph partition sizes

### LINKED ISSUE #627: Task counts for partition files that the current algorithm is missing
This is an issue to keep track of what is missing (and needs to be added manually) so we can eventually include them in the automatically generated partitions.

xylar:
These include multiples of the node sizes up to 20 nodes for all node sizes supported by E3SM.  They also include any core count that is an exact multiple of a supported node size that is within 2 nodes of a set of desirable set of core counts for the ne30 atmosphere/land mesh (factors of 5400).

This merge also takes care of some issues exposed while generating new partition files:
* some existing ocean meshes with ice-shelf cavities don't have the `landIceDraft` variable, so we simply skip it when generating an ocean mesh if it is not present.
* some meshes *do* contain a `cullCell` variable, which causes trouble for the sea-ice partitioning tools, so we remove that variable before passing the mesh to the sea-ice partition tools.

Checklist

closes #627

### DIFF
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/graph_partition.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/graph_partition.py
@@ -27,7 +27,15 @@ def get_core_list(ncells, max_cells_per_core=30000, min_cells_per_core=2):
     min_graph_size = max(2, int(ncells / max_cells_per_core))
     max_graph_size = int(ncells / min_cells_per_core)
 
-    cores = []
+    node_sizes = [16, 24, 30, 32, 36, 44, 56, 64, 84, 96, 112, 128, 256]
+    max_nodes = 20
+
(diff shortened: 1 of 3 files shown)


xylar:
I won't actually be able to test and use this branch until LCRC comes back up.

xylar:
## Testing

I generated additional partition files for the 10 meshes being actively supported in E3SM production runs or testing:
* EC30to60E2r2
* EC30to60E2r3
* ECwISC30to60E2r1
* oEC60to30v3
* oQU120
* oQU240
* oQU240wLI
* oQU480
* WC14to60E2r3
* WC14to60E2r5

I am still working on generating partition files for 2 more, but there is no sign of trouble with the branch as it is:
* oRRS18to6v3
* SOwISC12to60E2r4

PR #630, opened 2023-05-10
Explicitly specify nodes in srun calls

### LINKED ISSUE #629: Hanging on `pr` suite tests with 3 Perlmutter nodes
When I run the `pr` suite with 2 pm-cpu nodes (the current default), it runs fine in 15-20 minutes.  When I run on 3 nodes, certain steps hang for long periods of time:
* `ocean/baroclinic_channel/10km/decomp_test/8proc`
* `ocean/global_convergence/qu/cosine_bell/QU90/forward`
* ...

I suspect but have not been able to prove that this is related to the number of PIO tasks vs. the number of nodes the job is trying to run on.

I tried adding `--nodes` to the `srun` call that launches the step, matching the node count to the PIO task count, but this didn't seem to fix the issue.

xylar:
We want to explicitly specify how many nodes to use in `srun` calls for slurm steps because otherwise we may end up with a job running on more nodes than needed and an incompatible PIO task count.

Checklist

closes #629

### DIFF
--- a/compass/parallel.py
+++ b/compass/parallel.py
@@ -3,6 +3,7 @@ import os
 import subprocess
 import warnings
 
+import numpy as np
 from mpas_tools.logging import check_call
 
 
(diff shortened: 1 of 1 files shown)


xylar:
@matthewhoffman, it would be good to make sure the `full_integration` suite works fine on Perlmutter with these changes.  It might be worth testing on 3 nodes even if you don't have any tests that big, because that seemed to be the point at which the problem emerged.

xylar:
## Testing

The `pr` test suite on Perlmutter with 3 nodes passes with this branch and is BFB with a baseline using `main` (using 2 nodes).  Without this fix, the `pr` suite hangs on 3 nodes as reported in #629.

xylar:
With further testing, this problem persists even with the fix in this branch. I'm putting it in draft mode.  @matthewhoffman, please don't review or test just yet.

xylar:
Okay, things are now working for me in #631, so I think this is ready for you to test when you can, @matthewhoffman.

xylar:
@matthewhoffman, I ran the `full_integration` suite on Chrysalis and everything passed so I think we're good.  I'm going to go ahead with this one since it's the first in a chain that's getting pretty long.

matthewhoffman:
@xylar , sorry to not have noticed this until now.  Glad you were able to take care of it without me holding you up.

PR #631, opened 2023-05-10
Get global ocean resources and time steps based on mesh size

xylar:
This merge switches many global ocean steps to getting the core count from either an estimate or exact number of cells in the mesh.  Similarly, it switches to computing the time steps (both barotropic and baroclinic) from the minimum resolution of the mesh.

This work paves the way for generalized QU and icosahedral global ocean meshes.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/forward.py
+++ b/compass/ocean/tests/global_ocean/forward.py
@@ -1,10 +1,12 @@
 import os
+import time
 from importlib.resources import contents
 
 from compass.model import run_model
 from compass.ocean.tests.global_ocean.metadata import (
     add_mesh_and_init_metadata,
 )
+from compass.ocean.tests.global_ocean.tasks import get_ntasks_from_cell_count
 from compass.step import Step
 from compass.testcase import TestCase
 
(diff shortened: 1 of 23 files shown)


xylar:
I'm still seeing hanging in my tests of this branch, similar to #629.  In this case, it's for `ocean/global_ocean/QU240/WOA23/performance_test/forward`.

xylar:
## Testing

I ran the `pr` suite on Perlmutter, using the current `main` as a baseline.  Many RK4 tests (perf, decomp, threads) fail the BFB comparison, as expected because the time step has changed (from 5 to 6 minutes). Similarly, EC(wISC)30to60 perf. tests are non-BFB because the time step changed from 5 to 15 minutes and btr_dt changed from 10 to 45 seconds.

All other tests are passing.

xylar:
Retested on Chrysalis after merging #630 and rebasing.  I'm seeing non-BFB results for the cosine bell on Chrysalis in addition to the 5 tests mentioned above.  Differences are only for the QU60 test and are at roundoff.

PR #632, opened 2023-05-11
Add general QU and Icos meshes

xylar:
A user can set the resolution at setup or runtime with the `qu_resolution` config option in the `global_ocean` section.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/__init__.py
+++ b/compass/ocean/tests/global_ocean/__init__.py
@@ -16,6 +16,9 @@ from compass.ocean.tests.global_ocean.mesh.kuroshio.dynamic_adjustment import (
 from compass.ocean.tests.global_ocean.mesh.qu240.dynamic_adjustment import (
     QU240DynamicAdjustment,
 )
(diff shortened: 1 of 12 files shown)


xylar:
This builds off of #631 and #630, and can't really be tested until the LCRC server comes back online.

xylar:
## Testing

I ran QU and IcoswISC test cases (didn't try QUwISC and Icos, but they should work fine if the other 2 do) including `mesh`, `init` and `performance` on Chrysalis with Intel and OpenMPI.

Results are here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230513/qu*_2/
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/test_20230513/icoswisc*_2/
```

xylar:
The QU and Icos meshes are too time consuming and take too many codes to build (because of the topography remapping step) to be feasible to add to any test suites.  It doesn't seem useful to cache meshes, since the whole point of these meshes is that they are user configurable.

PR #566, opened 2023-03-16
Add bathymetry remapping

xylar:
This merge adds a step for remapping bathymetry and ice-shelf topography to the ocean framework.  It also adds this step to the mesh test cases in `global_ocean`.  It updates the `cull_mesh` step to use the remapped bathymetry to determine the land mask if topography remapping was performed.  It updates most meshes to use the this new approach, leaving the QU240 and Icos240 meshes as they were for more efficient testing.  Finally, this merge modifies the topography output with the variable names expected in MPAS-Ocean init mode and creates a stream for reading them in.

An alteration is needed to MPAS-Ocean to bypass reading in the land-ice topography from a lon-lat file.  Such a bypass exists for the bathymetry but not for land-ice topography.

Needs: https://github.com/E3SM-Project/E3SM/pull/5553

Checklist
* [x] User's Guide has been updated
* [x] Developer's Guide has been updated
* [x] API documentation in the Developer's Guide (`api.rst`) has any new or modified class, method and/or functions listed
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [x] The `E3SM-Project` submodule has been updated with relevant E3SM changes - merged in #585 
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #633, opened 2023-05-13
Remove unused namelist option

xylar:
The namelist option `config_global_ocean_land_ice_topo_source` doesn't exist.  Instead, `config_global_ocean_topography_source` is used both for bathymetry and land-ice topography.

This seems to have got missed in #566.

Checklist


xylar:
## Testing

I tested this along with #632.  Without this fix, I was seeing warnings about the unused config option.  After this change, I no longer see the warning.  The initial state gets created successfully.

PR #636, opened 2023-06-01
Add monthly output tests for all meshes

xylar:
We would like to run some test simulations with `timeSeriesStatsMonthlyOutput` to see how the AMOC behaves in MPAS standalone.  For this purpose, it seems useful to define a monthly output test for each mesh.  Other changes include:
* setting the number of cores based on the mesh size
* turning on surface restoring
* turning on global stats

Checklist


xylar:
## Testing

I ran the `ocean/global_ocean/EC30to60/WOA23/monthly_output_test` test case on Chrysalis for one month and verified that it completed successfully and produced the expected `analysis_members/mpaso.hist.am.timeSeriesStatsMonthly.0001-01-01.nc` file.

xylar:
@dengwirda, let me know how things go with this branch. I'll merge it in if it works okay for you.

xylar:
It seems like this wasn't needed.

PR #639, opened 2023-06-04
Always remap topogrpahy, even for low res meshes

xylar:
We use an older, lower resolution data set for the QU240 meshes.

This work is a preliminary step that will allow us to move toward initializing the global ocean in python, rather than Fortran.  

Checklist


xylar:
## Testing

I was able to run the `mesh`, `WOA23/init` and `WOA23/perforance_tests` test cases for both the QU240 and EC30to60 meshes with this branch. The QU240 `mesh` step takes 20-30 seconds longer than before, but this seems worth the cost because of the more consistent workflow and the future ability to eliminate MPAS-Ocean init mode.

PR #640, opened 2023-06-05
Renormalize topography after remapping in `global_ocean`

xylar:
A bug in #566 left the remapped topography multiplied by the remapped ocean mask.  This means that the topography falls back to zero at coastlines and grounding lines under ice shelves.  This merge fixes that issue by renormalizing topography as long as the remapped ocean mask is over a given threshold (0.01 by default).

This merge also turns off smoothing of topography in `global_ocean` init mode because it should no longer be needed with the conservative remapping approach.

Checklist


xylar:
## Testing

I ran the ECwISC mesh through `mesh`, `init` and `performance_test` test cases.  The topography is now renormalized as expected (values remain deep near grounding lines):
draft
bathymetry

PR #642, opened 2023-06-06
Update cached QU240 and QUwISC240 mesh and init files

xylar:
These are needed following #639 (and I forgot to add them as part of that PR).

Checklist


xylar:
## Testing

I successfully ran the `pr` suite on Chrysalis with Intel and OpenMPI with this branch, whereas it was failing to even set up following #639 because of missing cached files.

PR #643, opened 2023-06-06
Bound graph partition sizes

xylar:
The algorithm for computing graph partition sizes was previously producing partitions that were larger than the number of cells in the mesh.  With this fix, there will not be fewer than about 2 cells per task.

Checklist


xylar:
## Testing

I created a small test script to just test `get_core_list()`.  Whereas it previously could produced partitions that were larger than the number of cells, I tried this fix with `ncells=1791` (the size of the oQU480 mesh) and the partition file with the most tasks that I get is 880, (roughly 2 cells per task) as expected.

PR #644, opened 2023-06-06
For smaller meshes, create partitions for a single task

xylar:
This applies to meshes with fewer than 30,000 cells.  Although MPAS components run in standalone don't require partition files for a single core, E3SM does need them.

Checklist


xylar:
## Testing

With a test script similar to the one in #643, I was able to show that 1 task is included in the possible partition counts when I call `get_core_list()` with `ncells=1791` (i.e. < 30,000) but not when I call it with `ncells=179100` (i.e. > 30,000).

xylar:
I still need to test if `gpmetis` can actually make such a file.

xylar:
Unfortunately, no, this won't work:
```
$ gpmetis mp.graph.info.230422 1
The number of partitions should be greater than 1!
```

PR #653, opened 2023-07-20
Fix use of depricated np.float in hurricane analysis step

### LINKED ISSUE #652: Hurricane analysis step fails
This is due to the use of the depricated `np.float` type

sbrus89:
This PR also fix some isort and PEP8 issues in the analysis step caught by the pre-commit hooks.

This PR resolves #652.


sbrus89:
@xylar, I've tested this in reviewing https://github.com/E3SM-Project/E3SM/pull/5820. It should be fine to approve based on inspection.

PR #670, opened 2023-08-15
Add parabolic bowl test case

sbrus89:
This PR adds a parabolic bowl test case for wetting and drying. This case has an exact solution for the nonlinear shallow water equations, (see [Thacker, 1981](https://doi.org/10.1017/S0022112081001882 )). Here tests are added for the current wetting and drying implementation with and without the wetting velocity factor ramp. 

The E3SM branch required to run the parabolic bowl cases is here: https://github.com/sbrus89/E3SM/tree/ocn/parabolic_bowl

Checklist
* [x] New tests have been added to a test suite


compass/ocean/tests/parabolic_bowl/default/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        wetdry : str
+            The type of wetting and drying used (``standard``, ``subgrid``)
+        """
+        name = f'{wetdry}_{ramp_type}'
+
+        subdir = f'{wetdry}/{ramp_type}'
+        super().__init__(test_group=test_group, name=name,
+                         subdir=subdir)
+
+        resolutions = [5, 10, 20]
```
cbegeman, line 39: We have been making the list of resolutions a config option for other test cases so I would recommend moving this to parabolic_bowl.cfg.

compass/ocean/tests/parabolic_bowl/default/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+                                  wetdry=wetdry))
+        self.add_step(Viz(test_case=self, resolutions=resolutions))
+
+#    def validate(self):
+#        """
+#        Validate variables against a baseline
+#        """
+#        variables = ['layerThickness', 'normalVelocity']
+#        compare_variables(test_case=self, variables=variables,
+#                          filename1='forward/output.nc')
```
cbegeman, line 61: Is there a reason this is commented out?

compass/ocean/tests/parabolic_bowl/forward.py
```diff
@@ -0,0 +1,98 @@
...
+        self.resolution = resolution
+
+        super().__init__(test_case=test_case, name=name)
+
+        self.add_namelist_file('compass.ocean.tests.parabolic_bowl',
+                               'namelist.forward')
+
+        res_name = f'{resolution}km'
+        self.add_namelist_file('compass.ocean.tests.parabolic_bowl',
+                               f'namelist.{res_name}.forward')
```
cbegeman, line 46: When it's a matter of a single namelist option change, like time step, we've been making those changes with `add_namelist_options`. In this case, we've also been using a config option `dt_per_km` to determine how the time step should change as a function of resolution.

xylar (commented):
A few comments while you work on this.  One addition one is that it would be good to rebase and take out the E3SM-Project changes from 4fb836e42ea458e38e3f892deeddaba7182f92a1.  We want changes to E3SM-Project to happen in their own commit and it may be easier to do them as a separate PR.

compass/ocean/tests/parabolic_bowl/parabolic_bowl.cfg
```diff
@@ -0,0 +1,54 @@
...
+
+# Number of tasks for resolutions
+20km_ntasks = 72
+20km_min_tasks = 4
+
+10km_ntasks = 72
+10km_min_tasks = 4
+
+5km_ntasks = 128
+5km_min_tasks = 4
```
cbegeman, line 36: We could probably leave this out since in polaris we will be using these cfg options to determine the number of tasks:
```
# the number of cells per core to aim for
goal_cells_per_core = 200

# the approximate maximum number of cells per core (the test will fail if too
# few cores are available)
max_cells_per_core = 2000

**Comment by sbrus89** [comment]
The corresponding E3SM PR can be found here: https://github.com/E3SM-Project/E3SM/pull/5889

**Review by cbegeman — APPROVED** [review]
@sbrus89 It looks like you've addressed my comments and I've run this case successfully (just the nonramp one). Let me know if you would like further testing. Thanks for implementing this!

**Comment by sbrus89** [comment]
@cbegeman, thanks for the feedback! I added another of your suggestions you made on a call, which was to have plan-view plots. The viz step now makes a series of plots like this:

image

If you would be able to test the viz step to make sure this works for you, I'd really appreciate it.

File: compass/ocean/tests/parabolic_bowl/viz/__init__.py [hunk 5]
```diff
@@ -0,0 +1,324 @@
...
+                ds['layerThickness_exact'] = layerThickness_exact
+                ds.ssh_exact.encoding['_FillValue'] = None
+                ds.layerThickness_exact.encoding['_FillValue'] = None
+                ds.to_netcdf(f'output_{res}km.nc',
+                             format="NETCDF3_64BIT_OFFSET", mode='a')
+            ds.close()
+
+    def contour_plots(self, points):
+        """
+        Plot contour plots at a specified output interval for each resolution
```
cbegeman, line 131:
```suggestion
        Plot contour plots at a specified output interval for each resolution
        and show where the points used in `points.png` are located.
```

compass/ocean/tests/parabolic_bowl/namelist.forward
```diff
@@ -0,0 +1,13 @@
...
+config_vert_coord_movement='impermeable_interfaces'
+config_ALE_thickness_proportionality='weights_only'
+config_use_wetting_drying=.true.
+config_prevent_drying=.true.
+config_drying_min_cell_height=2.0e-2
+config_zero_drying_velocity=.true.
+config_verify_not_dry=.true.
+config_thickness_flux_type='upwind'
+config_use_cvmix=.false.
+config_use_implicit_bottom_drag=.false.
```
cbegeman, line 12:
```suggestion
config_bottom_drag_mode='explicit'
```

cbegeman: Is this right? Or is drag supposed to be entirely off?

xylar:
@sbrus89, please rebase to remove changes to the `MALI-Dev` submodule in https://github.com/MPAS-Dev/compass/pull/670/commits/1b4a82a349f75ea5a8ef8a649ec9cf1e48df7c5b

xylar:
I don't know why CI isn't running.  We need this to be fixed before I can approve and merge.

cbegeman:
@sbrus89 I love these plots. They are totally mesmerizing :)

The low res no ramp case looks like an iris!
image

compass/ocean/tests/parabolic_bowl/namelist.forward
sbrus89: Thanks for catching this. Bottom drag should be off for this case, so I deleted this and added `config_implicit_constant_bottom_drag_coeff = 0.0`

mark-petersen:
I was able to run this on LANL chicoma with gnu and chrysalis with intel, and tested both optimized and debug. I can see the same plots that you show above. Very nice! I would prefer that the order of execution of resolution goes from 20km to 5km, so the user can see that things are working quickly:
```
diff --git a/compass/ocean/tests/parabolic_bowl/default/__init__.py b/compass/ocean/tests/parabolic_bowl/default/__init__.py
index fe130ef087..ccb7fd437f 100644
--- a/compass/ocean/tests/parabolic_bowl/default/__init__.py
+++ b/compass/ocean/tests/parabolic_bowl/default/__init__.py
@@ -100,7 +100,7 @@ class Default(TestCase):
-        default_resolutions = '5, 10, 20'
+        default_resolutions = '20, 10, 5'
diff --git a/compass/ocean/tests/parabolic_bowl/parabolic_bowl.cfg b/compass/ocean/tests/parabolic_bowl/parabolic_bowl.cfg
index 3bd4f3f2ed..89e7ae2ca2 100644
--- a/compass/ocean/tests/parabolic_bowl/parabolic_bowl.cfg
+++ b/compass/ocean/tests/parabolic_bowl/parabolic_bowl.cfg
@@ -26,7 +26,7 @@ omega = 1.4544e-4

-resolutions = 5, 10, 20
+resolutions = 20, 10, 5
```

mark-petersen (approved):
Looks good! The documentation is beautiful! Just one misspelling, search for `propagtes`. Thanks!

sbrus89:
@xylar, this is ready for the E3SM submodule update now that https://github.com/E3SM-Project/E3SM/pull/5889 is merged. Did you still want to handle that?

xylar:
I'm running the `e3sm_udate` script right now.  I'll make a PR to update the submodule, and then this can be merged after that PR.

xylar:
The E3SM-Project submodule was updated in #692.  I will test this on a test merge with `main` and then merge if everything goes as expected.

xylar (changes requested):

xylar (approved):

xylar:
My tests passed!

sbrus89:
Thanks @xylar!

PR #671, opened 2023-08-16
Various fixes to the RRM tutorial

xylar:
Checklist

### DIFF
--- a/docs/tutorials/dev_add_rrm.rst
+++ b/docs/tutorials/dev_add_rrm.rst
@@ -177,7 +177,7 @@ Next, we need to add this mesh to the list of known meshes:
 
     ...
 
-        def __init__(self, test_group, mesh_name, remap_topography):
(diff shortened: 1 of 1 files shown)


xylar:
@irenavankova, can you take a look at this and recommend any further fixes?  We could also add a broader introduction to the structure of the global ocean test cases if you like, or you could make your own PR for that.

xylar:
## To fix

We need to switch to:
```
        self._add_tests(mesh_names=['YAM', 'YAMwISC'],
                        DynamicAdjustment=WC14DynamicAdjustment)
```
This means the step-by-step instructions somehow need to work in combination with `_add_tests()`.

It might be worth mentioning that you can do only the mesh with or without ice-shelf cavities if that makes more sense for your mesh.

It would be worth pointing to the Kuroshio meshes as an example of mostly sharing the same configuration and only having a few config options, etc. that are explicitly defined for each mesh.

PR #673, opened 2023-08-18
Minor changes to lts forward nmlist and streams

gcapodag:
Made minor changes to the streams and namelist of the hurricane lts test case.
The change to the streams is to print pointwise stats and output with the same frequency as the standard hurricane test. The change to namelist is to run for the same duration as the standard hurricane test.

Checklist


xylar (approved):
Approving by inspection.

xylar:
@gcapodag, can you make a comment here about whatever testing you've done with these changes?

gcapodag:
Tested on perlmutter, all hurricane lts steps are successful.

PR #675, opened 2023-08-18
Update to `mpas_tools` 0.23.0

xylar:
This brings in a bug fix where mask variables are cast to type int32 before they are passed to the `cull()` method, a requirement of the new `MpasCellCuller.x`, which does not support int64 or float mask.

Checklist

### DIFF
--- a/conda/compass_env/spec-file.template
+++ b/conda/compass_env/spec-file.template
@@ -22,7 +22,7 @@ mache=1.16.0
 {% endif %}
 matplotlib-base
 metis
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I ran the `pr` suite with the default configuration on Chrysalis.  All tests passed.  I didn't test against a baseline.

I reran the QUwISC240 mesh test case with output in NETCDF4 format that had been failing previously.  This time, it succeeded.

PR #676, opened 2023-08-21
Add a step for remapping sea surface salinity restoring

xylar:
This is used in E3SM runs with data atmospheric forcing.

Checklist


xylar:
## Testing

I ran the IcoswISC240 test cases through `files_for_e3sm` with this branch.  I successfully produced the salinity restoring file:
```
sss.PHC2_monthlyClimatology.IcoswISC240E3r1.20230821.nc
```
with the same dimensions, variables and variable attributes as attributes as files produced with @maltrud's csh script at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/sss_restoring/interpSSS
```
I visually verified that the salinities from the csh script and this branch look identical for the January data in paraview.  (I don't have a quick way to verify that they are bit-for-bit and that seems like overkill anyway.)

PR #678, opened 2023-08-23
Two fixes to data ice shelf melt fluxes

xylar:
This merge:
* adds a missing `Time` dimension to the ice shelf melt flux variables if one is not already present.
* fixes as issue where the step for remapping data ice shelf melt fluxes was being skipped in `files_for_e3sm` for existing meshes

It also removes the `landIceFluxMask` variable from the DISMF output file because this variable is not used and could cause confusion.

Checklist


xylar:
## Testing

I ran `files_for_e3sm` on 4 existing E3SM v3 meshes and was able to show that the `landIceFluxMask` variable was not in the output file and that the `Time` dimension was present.

I ran `mesh` and `data_ice_shelf_melt` steps for the `IcoswISC240` and now see the `Time` dimension in the output as expected.

PR #687, opened 2023-08-29
Update to geometric_features 1.3.0

xylar:
This brings in 2 new critical land blockages to make sure the interior of Greenland does not become part of the ocean (because of the GEBCO topography that we use).

Checklist


xylar:
## Testing

The `pr` suite ran fine on Chrysalis with this update.

PR #688, opened 2023-08-31
Update to GEBCO 2023 topography

xylar:
This version uses ice surface elevation, rather than bedrock below ice in Greenland.

Checklist


xylar:
## Testing

I was able to create an ECwISC30to60 with this new data set, and Greenland's ocean fraction looks as expected:
ocean fraction

Bathymetry also looks reasonable:

bathy

Issue #686, opened 2023-08-29
non-BFB results for ocean/drying_slope/1km/single_layer/ramp when using more than 9 procs


gcapodag:
The results for the `normalVelocity` are not BFB when comparing a serial run with a parallel run using 10 or more processes (I tried 10,11,12,13). Using 2,3,...,9 procs, the results for the `normalVelocity` are instead BFB.

xylar:
@gcapodag, I think this should also be an E3SM issue, since it is presumably a problem with MPAS-Ocean standalone, not with compass?

xylar:
But there's no problem with having issues in both places.

PR #695, opened 2023-09-05
Add decomp test case to the drying slope test group

cbegeman:
This PR adds a decomp test case to the drying slope test group.

Checklist
* [X] Developer's Guide has been updated
* [X] API documentation in the Developer's Guide (`api.rst`) has any new or modified class, method and/or functions listed
* [X] New tests have been added to a test suite

This test case helps evaluate https://github.com/MPAS-Dev/compass/issues/686.

### DIFF
--- a/compass/ocean/suites/wetdry.txt
+++ b/compass/ocean/suites/wetdry.txt
@@ -4,8 +4,10 @@ ocean/drying_slope/250m/single_layer/default
 ocean/drying_slope/250m/single_layer/ramp
 ocean/drying_slope/1km/sigma/default
 ocean/drying_slope/1km/sigma/ramp
+ocean/drying_slope/1km/sigma/decomp
 ocean/drying_slope/1km/single_layer/default
 ocean/drying_slope/1km/single_layer/ramp
+ocean/drying_slope/1km/single_layer/decomp
 ocean/dam_break/40cm/default
 ocean/dam_break/40cm/ramp
 ocean/dam_break/120cm/default
--- a/compass/ocean/tests/drying_slope/__init__.py
+++ b/compass/ocean/tests/drying_slope/__init__.py
@@ -1,3 +1,4 @@
+from compass.ocean.tests.drying_slope.decomp import Decomp
 from compass.ocean.tests.drying_slope.default import Default
 from compass.ocean.tests.drying_slope.loglaw import LogLaw
 from compass.ocean.tests.drying_slope.ramp import Ramp
@@ -21,6 +22,9 @@ class DryingSlope(TestGroup):
                 self.add_test_case(
                     Default(test_group=self, resolution=resolution,
                             coord_type=coord_type))
+                self.add_test_case(
+                    Decomp(test_group=self, resolution=resolution,
+                           coord_type=coord_type))
                 self.add_test_case(
                     Ramp(test_group=self, resolution=resolution,
                          coord_type=coord_type))
--- a/compass/ocean/tests/drying_slope/decomp/__init__.py
+++ b/compass/ocean/tests/drying_slope/decomp/__init__.py
@@ -1,0 +1,84 @@
+from compass.ocean.tests import drying_slope
+from compass.ocean.tests.drying_slope.forward import Forward
+from compass.ocean.tests.drying_slope.initial_state import InitialState
+from compass.testcase import TestCase
+from compass.validate import compare_variables
+
+
+class Decomp(TestCase):
+    """
+    A decomposition test case for the baroclinic channel test group, which
+    makes sure the model produces identical results on 1 and 12 cores.
+
+    Attributes
+    ----------
+    resolution : str
+        The resolution of the test case
+    """
+
+    def __init__(self, test_group, resolution, coord_type):
+        """
+        Create the test case
+
+        Parameters
+        ----------
+        test_group : compass.ocean.tests.baroclinic_channel.BaroclinicChannel
+            The test group that this test case belongs to
+
+        resolution : str
+            The resolution of the test case
+        """
+        name = 'decomp'
+        self.resolution = resolution
+        self.coord_type = coord_type
+        if resolution < 1.:
+            res_name = f'{int(resolution*1e3)}m'
+        else:
+            res_name = f'{int(resolution)}km'
+        subdir = f'{res_name}/{coord_type}/{name}'
+        super().__init__(test_group=test_group, name=name,
+                         subdir=subdir)
+
+        self.add_step(InitialState(test_case=self, coord_type=coord_type))
+        if coord_type == 'single_layer':
+            damping_coeff = None
+        else:
+            damping_coeff = 0.01
+        for procs in [1, 12]:
+            name = '{}proc'.format(procs)
+            forward_step = Forward(test_case=self, name=name, subdir=name,
+                                   resolution=resolution,
+                                   ntasks=procs, openmp_threads=1,
+                                   damping_coeff=damping_coeff,
+                                   coord_type=coord_type)
+            self.add_step(forward_step)
+
+    def configure(self):
+        """
+        Modify the configuration options for this test case.
+        """
+
+        resolution = self.resolution
+        config = self.config
+        ny = round(28 / resolution)
+        if resolution < 1.:
+            ny += 2
+        dc = 1e3 * resolution
+
+        config.set('drying_slope', 'ny', f'{ny}', comment='the number of '
+                   'mesh cells in the y direction')
+        config.set('drying_slope', 'dc', f'{dc}', comment='the distance '
+                   'between adjacent cell centers')
+
+    # no run() method is needed
+
+    def validate(self):
+        """
+        Test cases can override this method to perform validation of variables
+        and timers
+        """
+        variables = ['temperature', 'salinity', 'layerThickness',
+                     'normalVelocity']
+        compare_variables(test_case=self, variables=variables,
+                          filename1='1proc/output.nc',
+                          filename2='12proc/output.nc')
--- a/compass/ocean/tests/drying_slope/streams.forward
+++ b/compass/ocean/tests/drying_slope/streams.forward
@@ -30,6 +30,7 @@
 
     <stream name="mesh"/>
     <var name="layerThickness"/>
+    <var_struct name="tracers"/>
     <var name="ssh"/>
     <var name="normalVelocity"/>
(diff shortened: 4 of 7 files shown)


cbegeman:
### Testing

I ran both `ocean/drying_slope/1km/sigma/decomp` and `ocean/drying_slope/1km/single_layer/decomp` on chrysalis with intel, openmpi. They fail as expected due to the issue pointed out by @gcapodag https://github.com/E3SM-Project/E3SM/issues/5902. I am still waiting on the results of the `wetdry` test suite and will update this comment when complete.

cbegeman:
@gcapodag This is the branch I used to replicate your issue. I figured I'd add it to compass. It would be great if you had a chance to check it out and run (maybe both with master and your fix).

gcapodag:
Thank you @cbegeman . I tested on Perlmutter with gnu. I ran `ocean/drying_slope/1km/sigma/decomp` and `ocean/drying_slope/1km/single_layer/decomp` with master and the tests fail. With the fix I came up with in https://github.com/E3SM-Project/E3SM/issues/5902 they both pass. It is sufficient to just add this halo update call
``` 
call mpas_dmpar_field_halo_exch(domain, 'layerThicknessEdgeFlux')
```
 right before this code in RK4:
```
        block => domain % blocklist
        do while (associated(block))
           call ocn_time_integrator_rk4_compute_vel_tends(domain, block, dt, rk_substep_weights(rk_step), domain % dminfo, err )

           call ocn_time_integrator_rk4_compute_thick_tends( block, dt, rk_substep_weights(rk_step), err )
           block => block % next
        end do
```
 I wonder if the reason why we need this update should be investigated further. As I mentioned to @xylar , the `layerThickEdgeFlux` variable is incorrect at the beginning of the RK4 routine, when that diagnostic variable is computed with the old solution, so it does not seem an issue related to RK4 itself. I understand that with `split_explicit`  problems that might be related to this have not shown, though we should probably still look into the computation of `layerThickEdgeFlux`, at least to understand why problems are not showing with `split_explicit` but they are in the present case.

gcapodag:
Looking at the computation of `layerThickEdgeFlux` right now, considering the case of `upwind` which seems what currently is set on the `namelist` for the tests mentioned above:
```
         do iEdge = 1, nEdgesAll
            kmin = minLevelEdgeBot(iEdge)
            kmax = maxLevelEdgeTop(iEdge)
            cell1 = cellsOnEdge(1,iEdge)
            cell2 = cellsOnEdge(2,iEdge)
            do k=1,nVertLevels
               ! initialize layerThicknessEdgeFlux to avoid divide by
               ! zero and NaN problems.
               layerThickEdgeFlux(k,iEdge) = -1.0e34_RKIND
            end do
            do k = kmin,kmax
               if (normalVelocity(k,iEdge) > 0.0_RKIND) then
                  layerThickEdgeFlux(k,iEdge) = layerThickness(k,cell1)
               elseif (normalVelocity(k,iEdge) < 0.0_RKIND) then
                  layerThickEdgeFlux(k,iEdge) = layerThickness(k,cell2)
               else
                  layerThickEdgeFlux(k,iEdge) = &
                                      max(layerThickness(k,cell1), &
                                          layerThickness(k,cell2))
               end if
            end do
         end do
```
I am a little confused by the above computation: why is a halo update not needed at present? If we are looping over `nEdgesAll` that includes the outermost edges the process has access to, but then we use `cellsOnEdge`, so would that not return a cell that is even outside of the halo cells of the process? Even if `layerThickness` has been updated before this computation, doesn't that only apply to the owned+halo cells? If we try to access a cell that is even outside the owned+halo would `layerThickness` be reliable?

cbegeman:
@gcapodag Thanks for doing that testing. I'm happy to take a look at the RK4 routine to try to understand why the line is needed. 

Upwinding with split-explicit does pass a 1 proc vs 12 proc decomp test when applied to the baroclinic channel test case. It cannot be tested with the drying slope test case because I have not yet completed the W&D implementation for split-explicit.

cbegeman:
> I am a little confused by the above computation: why is a halo update not needed at present? If we are looping over nEdgesAll that includes the outermost edges the process has access to, but then we use cellsOnEdge, so would that not return a cell that is even outside of the halo cells of the process? Even if layerThickness has been updated before this computation, doesn't that only apply to the owned+halo cells? If we try to access a cell that is even outside the owned+halo would layerThickness be reliable?

@gcapodag With respect to this question, I don't think we need to worry about looping over `nEdgesAll` because any cell that is outside the domain will have a `maxLevelEdgeTop` of 0 and thus the `layerThickEdgeFlux` will stay the initialized value of -1e34.

gcapodag:
> > I am a little confused by the above computation: why is a halo update not needed at present? If we are looping over nEdgesAll that includes the outermost edges the process has access to, but then we use cellsOnEdge, so would that not return a cell that is even outside of the halo cells of the process? Even if layerThickness has been updated before this computation, doesn't that only apply to the owned+halo cells? If we try to access a cell that is even outside the owned+halo would layerThickness be reliable?
> 
> @gcapodag With respect to this question, I don't think we need to worry about looping over `nEdgesAll` because any cell that is outside the domain will have a `maxLevelEdgeTop` of 0 and thus the `layerThickEdgeFlux` will stay the initialized value of -1e34.

Hi @cbegeman , by outside the domain you mean outside the actual global domain or outside the partition handled by a given process?

cbegeman:
> Hi @cbegeman , by outside the domain you mean outside the actual global domain or outside the partition handled by a given process?

@gcapodag Oh, I misunderstood. I was talking about the whole domain. You're concerned with the partition. I believe I was motivated by trying to define `layerThickEdgeFlux` so that it would be defined for loops like this, where config_num_halos is 3 by default. 

https://github.com/E3SM-Project/E3SM/blob/82c581e987a397ffd601971b108e788af9a15615/components/mpas-ocean/src/mode_forward/mpas_ocn_time_integration_split.F#L801-L817

            bottomDepthEdge(:) = 0.0_RKIND
            !$omp parallel
            !$omp do schedule(runtime) &
            !$omp private(cell1, cell2, k, thicknessSum)
            do iEdge = 1, nEdgesHalo(config_num_halos+1)
               cell1 = cellsOnEdge(1,iEdge)
               cell2 = cellsOnEdge(2,iEdge)
               thicknessSum = layerThickEdgeFlux(minLevelEdgeBot(iEdge),iEdge)
               do k = minLevelEdgeBot(iEdge)+1, maxLevelEdgeTop(iEdge)
                  thicknessSum = thicknessSum + &
                                 layerThickEdgeFlux(k,iEdge)
               enddo
               bottomDepthEdge(iEdge) = thicknessSum &
                  - 0.5_RKIND*(sshNew(cell1) + sshNew(cell2))
            enddo ! iEdge
            !$omp end do
            !$omp end parallel

I suppose you could try replacing `nEdgesAll` with `nEdgesHalo(config_num_halos+1)` in the loop where we compute `layerThickEdgeFlux` and see if the solution is identical.

Am I getting warmer in terms of addressing your concerns? I'm not the most knowledgeable about halos in MPAS-O.

xylar (approved):
Sorry I lost track of the review request here.  I haven't run the test myself but the code looks great!  I'm happy to have you merge when you're ready.

cbegeman:
@xylar Thank you for reviewing! Would you rather I merge with the 1 vs. 12 proc. comparison (as written here), which will fail because we don't yet have @gcapodag 's bugfix, or replace it with the 4 vs. 8 proc. comparison, which will pass?

xylar:
I think it's better if it fails to remind us to merge the fix.  I guess we could wait and merge the test after the fix is in but I think it's fine if the test fails as long as it's not in the `pr` suite.

cbegeman:
@gcapodag Did you get a chance to check out this branch and try running it? If so, could you approve the PR? You're welcome to provide other feedback if you have any. If you don't have time to review, let me know.

gcapodag (approved):
Hi @cbegeman, I tested again on Perlmutter and I think we can merge. I took some time to look further into why the tests are failing in the current setup, and found out that the problem is with `layerThickness` and not with `layerThickEdgeFlux`. There is a freaky situation happening in which for some not yet identified reason, the `layerThickness` is not BFB anymore at some point during the RK4 time stepping and that affects the computation of `layerThickEdgeFlux`. If I add a halo update in RK4 right before the final computation of the diagnostic variables, the BFB issue is gone (so we do not need a halo update on `layerThickEdgeFlux`). The interesting/weird/troubling thing is that the issue with `layerThickEdgeFlux` is actually present also with 8 procs for instance, but due to the process configuration I suppose, the "bad" value is actually never used in the computation of the `hadv_coriolis` tendency contribution. This is a print from inside the loop on edge on edge in `hadv_coriolis` for 8 procs (focusing on `eoe` such that `indexToEdgeID(eoe)==14`):
```
 from tendency:    1.3944039866367177               11          18
 from tendency:    1.3944039866367177               11          18
 from tendency:    1.3944039866367177               11          18
 from tendency:    1.3944039866367177               11          18
 from tendency:    1.3944039866367177               11          18
 from tendency:    1.3944039866367177               11          18
 from tendency:    1.3944039866367177               11          18
 from tendency:    1.3944039866367177               11          18
```
the first number is `layerThickEdgeFlux(k,eoe)` and the other two are the two cells in local numbering that share that edge. Note that the edge `eoe` that is picked up is always the same, which happens to be the one with the correct value of `layerThickEdge`. For 10 procs, one time the loop picks up the edge with the wrong (i.e. not halo updated value):
```
 from tendency:    1.3944039866367177                7          14
 from tendency:    1.3944039866367177                7          14
 from tendency:    1.3944039866367177                7          14
 from tendency:    1.3944039866367177                7          14
 from tendency:    **10.045894683899487**               31          45
 from tendency:    1.3944039866367177                7          14
 from tendency:    1.3944039866367177                7          14
 from tendency:    1.3944039866367177                7          14
 ```
So now, there are two options in my opinion: 1. add a halo update on the layer thickness before the final computation of the diagnostics in RK4. 2. spend time to understand why the layer thickness is not BFB in the first place.  Obviousuly one route is less painful than the other.

cbegeman:
@gcapodag Thanks for your testing and this additional context! I'm moving away from using RK4 configurations on my projects, and since RK4 is also not used in E3SM configurations, this may really be a decision for ICoM. I think that given ICoM's priorities the former option would be sufficient. However, given the shifting priorities, you could raise this with Rob Hetland and see what he thinks.

PR #668, opened 2023-08-15
New Mesh: RRSwISC6to18E3r1

xylar:
Long name: RRSwISC6to18L80E3SMv3r1

This RRS (Rossby-radius scaled) mesh has:
* 6 km resolution near the poles
* 18 km resolution at the equator

Mesh, initial condition, dynamic adjustment and files for E3SM are on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/rrswisc6to18e3r1
```

Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #697, opened 2023-09-12
Allow culled mesh to be converted to CDF-5 format

xylar:
We will do this for large meshes like RRS6to18 for better performance.

This merge also fixes an `ncks` call for converting to CDF5 format but overwriting the file if it already exists.

Checklist


xylar:
## Testing

This has been tested in #668

PR #691, opened 2023-09-01
New Mesh: IcoswISC30E3r2

xylar:
Long name: IcoswISC30L64E3SMv3r2

This Icosahedral (Icos) mesh is made by subdividing an icosahedron and has nearly uniform 30 km resolution.

Mesh, initial condition, dynamic adjustment and files for E3SM are on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/icoswisc30e3r2
```

Checklist
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #669, opened 2023-08-15
New Mesh: SOwISC12to60E3r1

xylar:
Long name: SOwISC12to60kmL64E3SMv3r1

This version of the Southern Ocean Regionally Refined Mesh (SORRM) includes:
* 12 km resolution around Antarctica, about 1000 km farther north than in revision 3 of the mesh
* 45 km resolution at southern mid-latitudes
* 30 km resolution at the equator and the north Atlantic
* 60 km resolution in the north Pacific
* 35 km resolution in the Arctic

It matches the EC30to60 mesh except in the Southern Ocean and north Atlantic.

Mesh, initial condition, dynamic adjustment and files for E3SM are on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/sowisc12to60e3r1
```

Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #698, opened 2023-09-18
Update to pyremap 1.1.0

xylar:
Checklist

### DIFF
--- a/compass/ocean/mesh/remap_topography.py
+++ b/compass/ocean/mesh/remap_topography.py
@@ -2,7 +2,7 @@ import os
 
 import xarray as xr
 from mpas_tools.io import write_netcdf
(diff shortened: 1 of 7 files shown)


xylar:
## Testing

This is non-bit-for-bit for the `pr` suite, specifically for the QU240 `init` test, but the changes are small -- ~1e-9.  This is likely due to changing to SCRIP files from ESMF grid files when remapping topography.

I also tested that `files_for_e3sm` runs for the QU240 mesh with these changes.

xylar:
Because of the non-bit-for-bit changes, this shouldn't be merged until after #691, #668 and #669.

xylar:
I will tag #668 before the merge commit (i.e. excluding this merge) so we can get this in.  The new meshe in #699, #700 and #701 should use this new version of pyremap.

PR #699, opened 2023-09-20
New Mesh: WCwISC14to60E3r1

xylar:
Long name: WCwISC14to60kmL64E3SMv3r1

This version of the Water Cycle (WC) Regionally Refined Mesh (RRM) has the same distribution of resolution as in [WC14to60E2r3](https://github.com/MPAS-Dev/MPAS-Model/pull/628), including:
* 14 km resolution around North America and in the Arctic
* 35 km resolution in the Southern Ocean

It matches the EC30to60 mesh except in the Southern Ocean and north Atlantic.  This mesh is with Ice Shelf Cavities (wIsC) and has 64 vertical levels, with surface levels distributed similarly to those in the E3SM v2 WC14 initial condition.

```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/wcwisc14to60e3r1
```

Checklist


xylar:
Needs `landIceMask` fix from #732

PR #689, opened 2023-08-31
New Mesh: ECwISC30to60E3r2

xylar:
Long name: ECwISC30to60L64E3SMv3r2

As with previous versions of the mesh, this Eddy Closure (EC) mesh has:
* 30 km resolution at the equator
* 60 km resolution at mid latitudes
* 35 km resolution near the poles

Mesh, initial condition, dynamic adjustment and files for E3SM are on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/ecwisc30to60e3r2
```

Checklist
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

Replaces #666

PR #700, opened 2023-09-20
New Mesh: EC30to60E3r3

xylar:
Long name: EC30to60L64E3SMv3r3

This is identical to #689 but without ice-shelf cavities.

```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/ec30to60e3r3
```

Checklist


xylar:
Will not use

PR #706, opened 2023-09-28
New Mesh: RRS6to18E3r2

xylar:
Long name: RRS6to18L80E3SMv3r2

Mesh, initial condition, dynamic adjustment and files for E3SM will soon on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/rrs6to18e3r2
```

Checklist


xylar:
Will follow up once an RRSwISC mesh has been created and tested

PR #714, opened 2023-10-05
Temporary fix to reinstall mamba=1.5.1

xylar:
There is an issue where the latest mamba isn't compatible with the latest conda but locally cached versions of the mamba package don't know this.  The mamba package has been patched to a reinstall will fix teh issue. This should be fixed more permanently in mamba 1.5.2, which should be released soon.

Checklist


xylar:
## Testing

This worked fine for me on Chrysalis (using a very new install of mambaforge).  It it proving to be extraordinarily slow on my laptop, perhaps because I have some other dependencies in my base environment for building packages that are giving the very slow conda solver a lot of trouble.  Can't wait to have this fixed "properly"!

PR #715, opened 2023-10-08
New Mesh: ECwISC30to60E3r4

xylar:
Long name: ECwISC30to60L64E3SMv3r4

This mesh differs from ECwISC30to60E3r2 (#689) only in having topography smoothing of approximately 6 grid cells (180-360 km), see comment below.

The `mesh` test case is a symlink to that of the `ECwISC30to60E3r2` to guarantee that they are identical.

Initial condition, dynamic adjustment and files for E3SM will be on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/ecwisc30to60e3r4
```

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/global_ocean.cfg
+++ b/compass/ocean/tests/global_ocean/global_ocean.cfg
@@ -47,6 +47,11 @@ btr_dt_per_km = 1.5
 
 # Maximum allowed Haney number for configurations with ice-shelf cavities
 rx1_max = 20
+# the number of iterations of topography smoothing
+topo_smooth_iterations = 0
+# the weight given to the central cell during each iteration of smoothing,
+# where the n cellsOnCell (neighbors) are given a weight (1-weight)/n.
+topo_smooth_weight = 0.9
 
 # number of cores to use
 init_ntasks = 36
--- a/compass/ocean/tests/global_ocean/init/initial_state.py
+++ b/compass/ocean/tests/global_ocean/init/initial_state.py
@@ -139,9 +139,21 @@ class InitialState(Step):
         Get resources at setup from config options
         """
         self._get_resources()
-        rx1_max = self.config.getfloat('global_ocean', 'rx1_max')
-        self.add_namelist_options({'config_rx1_max': f'{rx1_max}'},
-                                  mode='init')
+        section = self.config['global_ocean']
+        rx1_max = section.getfloat('rx1_max')
+        topo_smooth_iterations = section.getint('topo_smooth_iterations')
+        topo_smooth_weight = section.getfloat('topo_smooth_weight')
+
+        options = {
+            'config_rx1_max':
+                f'{rx1_max}',
+            'config_global_ocean_topography_smooth_iterations':
+                f'{topo_smooth_iterations}',
+            'config_global_ocean_topography_smooth_weight':
+                f'{topo_smooth_weight}',
+        }
+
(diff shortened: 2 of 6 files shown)


xylar:
To determine the weight and number of iterations of smoothing, I used the following script:
```python
#!/usr/bin/env python3

import numpy as np
from matplotlib import pyplot as plt


def main():
    res = 30.
    x = res * np.arange(-40, 41)
    f = np.where(x >= 0, 1., 0.)

    gaussian_iter = 3
    sigma = 100.
    x_max = 200.
    f_gaussian = f
    for index in range(gaussian_iter):
        f_gaussian = gaussian_smooth(x, f_gaussian, sigma, x_max)

    mpas_init_iter = 65
    weight = 0.6
    f_mpas_init = f
    for index in range(mpas_init_iter):
        f_mpas_init = mpas_init_mode_smooth(f_mpas_init, weight)

    plt.plot(x, f, 'b', label='original')
    plt.plot(x, f_gaussian, 'r',
             label=f'Mat\'s Gaussian (iter={gaussian_iter}, sigma={sigma})')
    plt.plot(x, f_mpas_init, 'g',
             label=f'MPAS-Ocean init (iter={mpas_init_iter}, weight={weight})')
    plt.legend(loc='best')
    plt.show()


def gaussian_smooth(x, f, sigma, x_max):
    mask = np.logical_and(x >= -x_max, x <= x_max)
    x_kernel = x[mask]
    kernel = np.exp(-x_kernel**2/(2. * sigma**2))
    kernel = kernel / np.sum(kernel)

    f_gaussian = np.convolve(f, kernel, mode='same')
    ones_gaussian = np.convolve(np.ones_like(f), kernel, mode='same')

    mask = ones_gaussian > 0.
    f_gaussian[mask] = f_gaussian[mask] / ones_gaussian[mask]
    return f_gaussian


def mpas_init_mode_smooth(f, weight):
    kernel = np.array([0.5*(1. - weight), weight, 0.5*(1. - weight)])

    f_mpas_init = np.convolve(f, kernel, mode='same')
    ones_mpas_init = np.convolve(np.ones_like(f), kernel, mode='same')

    mask = ones_mpas_init > 0.
    f_mpas_init[mask] = f_mpas_init[mask] / ones_mpas_init[mask]
    return f_mpas_init


if __name__ == '__main__':
    main()
```

It shows nearly identical smoothing of a 1D step function using @maltrud's 3 iterations of a gaussian with sigma = 100 km and MPAS-Ocean's init mode approach with a central weight of 0.6 and 65 iterations:

smoothing

I used a characteristic resolution of 30 km.  In both cases, the smoothed version of the step function has a value of 0.25 and 0.75 at about -90 km and 90 km, respectively.  That is almost exactly 3 grid cells.  Whereas @maltrud's approach is independent of MPAS mesh resolution, the MPAS-Ocean init mode approach will scale with the size of grid cells, so the 2 will agree at 30-km resolution but the init mode approach will produce more smoothing at 60 km resolution.

xylar:
## Bathymetry 

### ECwISC30to60E3r2

bottom_depth

### ECwISC30to60E3r4

bottom_depth

xylar:
@vanroekel, I think we agreed that we don't want to pursue this version of the mesh, so I will close this for now.

vanroekel:
Yes agreed.  Thanks @xylar

PR #719, opened 2023-10-16
Fix conversion to absolute paths in config files

xylar:
We don't want to do this for config options that start with other config options.

Backport of https://github.com/E3SM-Project/polaris/pull/136

Checklist


xylar:
## Testing

An identical fix in Polaris (https://github.com/E3SM-Project/polaris/pull/136) fixed an issue when `-p` was not provided.  I'm going to take a chance that the same bug will be fixed here and spare myself some redundant testing.

PR #723, opened 2023-10-25
Fixup ISOMIP+ land ice pressure

cbegeman:
This PR changes the calculation of land ice pressure for ISOMIP+ thin-film test cases so that it is a function of the land ice thickness rather than land ice draft. The land ice pressure is not perfectly identical between `Ocean0` and `thin_film_Ocean0` test cases, mostly because in the former case we smooth the draft and in the latter we smooth ice thickness. This PR also fixes a bug wherein the thin film region was culled from the domain.

Checklist
* [X] API documentation in the Developer's Guide (`api.rst`) has any new or modified class, method and/or functions listed

### DIFF
--- a/compass/ocean/iceshelf.py
+++ b/compass/ocean/iceshelf.py
@@ -11,6 +11,69 @@ from compass.io import symlink
 from compass.model import partition, run_model
 
 
+def compute_land_ice_pressure_from_thickness(land_ice_thickness, modify_mask,
+                                             land_ice_density=None):
+    """
+    Compute the pressure from and overlying ice shelf
+
+    Parameters
+    ----------
+    land_ice_thickness: xarray.DataArray
+        The ice thickness
+
+    modify_mask : xarray.DataArray
+        A mask that is 1 where ``landIcePressure`` can be deviate from 0
+
+    land_ice_density : float, optional
+        A reference density for land ice
+
+    Returns
+    -------
+    land_ice_pressure : xarray.DataArray
+        The pressure from the overlying land ice on the ocean
+    """
+    gravity = constants['SHR_CONST_G']
+    if land_ice_density is None:
+        land_ice_density = constants['SHR_CONST_RHOICE']
+    land_ice_pressure = modify_mask * \
+        numpy.maximum(land_ice_density * gravity * land_ice_thickness, 0.)
+    return land_ice_pressure
+
+
+def compute_land_ice_density_from_draft(land_ice_draft, land_ice_thickness,
+                                        floating_mask, ref_density=None):
+    """
+    Compute the spatially-averaged ice density needed to match the ice draft
+
+    Parameters
+    ----------
+    land_ice_draft : xarray.DataArray
+        The ice draft (sea surface height)
+
+    land_ice_thickness: xarray.DataArray
+        The ice thickness
+
+    floating_mask : xarray.DataArray
+        A mask that is 1 where the ice is assumed in hydrostatic equilibrium
+
+    ref_density : float, optional
+        A reference density for seawater displaced by the ice shelf
+
+    Returns
+    -------
+    land_ice_density: float
+        The ice density
+    """
+    if ref_density is None:
+        ref_density = constants['SHR_CONST_RHOSW']
+    land_ice_draft = numpy.where(floating_mask, land_ice_draft, numpy.nan)
+    land_ice_thickness = numpy.where(floating_mask, land_ice_thickness,
+                                     numpy.nan)
+    land_ice_density = \
+        numpy.nanmean(-ref_density * land_ice_draft / land_ice_thickness)
+    return land_ice_density
+
+
 def compute_land_ice_pressure_from_draft(land_ice_draft, modify_mask,
                                          ref_density=None):
     """
@@ -39,6 +102,36 @@ def compute_land_ice_pressure_from_draft(land_ice_draft, modify_mask,
         modify_mask * numpy.maximum(-ref_density * gravity * land_ice_draft,
                                     0.)
     return land_ice_pressure
+
+
+def compute_land_ice_draft_from_pressure(land_ice_pressure, modify_mask,
+                                         ref_density=None):
+    """
+    Compute the ice-shelf draft associated with the pressure from an overlying
+    ice shelf
+
+    Parameters
+    ----------
+    land_ice_pressure : xarray.DataArray
+        The pressure from the overlying land ice on the ocean
+
+    modify_mask : xarray.DataArray
+        A mask that is 1 where ``landIcePressure`` can be deviate from 0
+
+    ref_density : float, optional
+        A reference density for seawater displaced by the ice shelf
+
+    Returns
+    -------
+    land_ice_draft : xarray.DataArray
+        The ice draft
+    """
+    gravity = constants['SHR_CONST_G']
+    if ref_density is None:
+        ref_density = constants['SHR_CONST_RHOSW']
+    land_ice_draft = \
+        - (modify_mask * land_ice_pressure / (ref_density * gravity))
+    return land_ice_draft
 
 
 def adjust_ssh(variable, iteration_count, step, update_pio=True,
--- a/compass/ocean/iceshelf.py
+++ b/compass/ocean/iceshelf.py
@@ -40,40 +40,6 @@ def compute_land_ice_pressure_from_thickness(land_ice_thickness, modify_mask,
     return land_ice_pressure
 
 
-def compute_land_ice_density_from_draft(land_ice_draft, land_ice_thickness,
-                                        floating_mask, ref_density=None):
-    """
-    Compute the spatially-averaged ice density needed to match the ice draft
-
-    Parameters
-    ----------
-    land_ice_draft : xarray.DataArray
-        The ice draft (sea surface height)
-
-    land_ice_thickness: xarray.DataArray
-        The ice thickness
-
-    floating_mask : xarray.DataArray
-        A mask that is 1 where the ice is assumed in hydrostatic equilibrium
-
-    ref_density : float, optional
-        A reference density for seawater displaced by the ice shelf
-
-    Returns
-    -------
-    land_ice_density: float
-        The ice density
-    """
-    if ref_density is None:
-        ref_density = constants['SHR_CONST_RHOSW']
-    land_ice_draft = numpy.where(floating_mask, land_ice_draft, numpy.nan)
-    land_ice_thickness = numpy.where(floating_mask, land_ice_thickness,
-                                     numpy.nan)
-    land_ice_density = \
-        numpy.nanmean(-ref_density * land_ice_draft / land_ice_thickness)
-    return land_ice_density
-
-
 def compute_land_ice_pressure_from_draft(land_ice_draft, modify_mask,
                                          ref_density=None):
     """
--- a/compass/ocean/iceshelf.py
+++ b/compass/ocean/iceshelf.py
@@ -14,7 +14,7 @@ from compass.model import partition, run_model
 def compute_land_ice_pressure_from_thickness(land_ice_thickness, modify_mask,
                                              land_ice_density=None):
     """
-    Compute the pressure from and overlying ice shelf
+    Compute the pressure from an overlying ice shelf from ice thickness
 
     Parameters
     ----------
@@ -43,7 +43,7 @@ def compute_land_ice_pressure_from_thickness(land_ice_thickness, modify_mask,
 def compute_land_ice_pressure_from_draft(land_ice_draft, modify_mask,
                                          ref_density=None):
     """
-    Compute the pressure from and overlying ice shelf
+    Compute the pressure from an overlying ice shelf from ice draft
 
     Parameters
     ----------
--- a/compass/ocean/iceshelf.py
+++ b/compass/ocean/iceshelf.py
@@ -11,34 +11,34 @@ from compass.io import symlink
 from compass.model import partition, run_model
 
 
-def compute_land_ice_pressure_and_draft(ssh, modify_mask, ref_density):
+def compute_land_ice_pressure_from_draft(land_ice_draft, modify_mask,
+                                         ref_density=None):
     """
-    Compute the pressure from and overlying ice shelf and the ice-shelf draft
+    Compute the pressure from and overlying ice shelf
 
     Parameters
     ----------
-    ssh : xarray.DataArray
-        The sea surface height (the ice draft)
+    land_ice_draft : xarray.DataArray
+        The ice draft (sea surface height)
 
     modify_mask : xarray.DataArray
         A mask that is 1 where ``landIcePressure`` can be deviate from 0
 
-    ref_density : float
+    ref_density : float, optional
         A reference density for seawater displaced by the ice shelf
 
     Returns
     -------
-    landIcePressure : xarray.DataArray
+    land_ice_pressure : xarray.DataArray
         The pressure from the overlying land ice on the ocean
-
-    landIceDraft : xarray.DataArray
-        The ice draft, equal to the initial ``ssh``
     """
     gravity = constants['SHR_CONST_G']
-    landIcePressure = \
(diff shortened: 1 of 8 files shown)


cbegeman:
### Testing

Ocean0 test case has been run on chrys with intel, openmpi. thin_film_Ocean0 has been run but fails due to state validation issues at ssh_adjustment. This is not believed to be related to this PR.

cbegeman:
@xylar Can you let me know what you think? I ended up doing something a little different from you to make the `Ocean0` and `thin_film_Ocean0` land ice pressure fields more similar. Open to other ideas though.

compass/ocean/iceshelf.py
```diff
@@ -11,34 +11,127 @@
...
+        A reference density for land ice
+
+    Returns
+    -------
+    land_ice_pressure : xarray.DataArray
+        The pressure from the overlying land ice on the ocean
+    """
+    gravity = constants['SHR_CONST_G']
+    if land_ice_density is None:
+        land_ice_density = constants['SHR_CONST_RHOICE']
```
xylar, line 37: I think this constant is the density of sea-ice but I could be wrong.

compass/ocean/tests/isomip_plus/initial_state.py
```diff
@@ -125,12 +129,20 @@ def _compute_initial_condition(self):
...
 
-        ref_density = constants['SHR_CONST_RHOSW']
-        landIcePressure, landIceDraft = compute_land_ice_pressure_and_draft(
-            ssh=ds.ssh, modify_mask=ds.ssh < 0., ref_density=ref_density)
+        land_ice_draft = ds.ssh
+        if thin_film_present:
+            land_ice_thickness = ds.landIceThickness
+            land_ice_density = compute_land_ice_density_from_draft(
+                land_ice_draft, land_ice_thickness,
+                floating_mask=ds.landIceFloatingMask)
```
xylar, line 137: I think we should just include this in the config options, like I am doing for Polaris:
https://github.com/xylar/polaris/blob/add-isomip-plus-init/polaris/ocean/tasks/isomip_plus/isomip_plus.cfg#L27
We know what BISICLES used in their simulation, as it's also the value prescribed in the paper (Table 1 in [Asay-Davis et al. 2016](https://gmd.copernicus.org/articles/9/2471/2016/gmd-9-2471-2016.html)).

compass/ocean/tests/isomip_plus/initial_state.py
```diff
@@ -125,12 +129,20 @@ def _compute_initial_condition(self):
 
         ds['landIceFraction'] = xr.where(mask, ds.landIceFraction, 0.)
 
-        ref_density = constants['SHR_CONST_RHOSW']
-        landIcePressure, landIceDraft = compute_land_ice_pressure_and_draft(
-            ssh=ds.ssh, modify_mask=ds.ssh < 0., ref_density=ref_density)
+        land_ice_draft = ds.ssh
```
xylar, line 132: The reason I don't do this for the thin film case in Polaris is because the ocean won't want to have the SSH at the flotation height that BISICLES wants to give it.  This is at least partly because BISICLES used a different reference density for the ocean (1028 instead of 1026) and partly because it has been my experience that computing the ice draft (SSH) from the land-ice pressure and the reference ocean density tends to produce the most reliable starting point for the SSH adjustment step.

xylar: I'm not sure if this will help with the state validation issues during SSH adjustment for the thin-film case but I think it might.

xylar: I discovered that I was missing a check in my Polaris computation of land_ice_draft:
```python
            land_ice_draft = np.maximum(ds_topo.bedrockTopography,
                                        land_ice_draft)
```
We obviously don't what to allow the ice draft to go deeper than the bedrock in the thin film region.

compass/ocean/iceshelf.py
cbegeman: It's 917 which seems fine for land ice https://github.com/MPAS-Dev/MPAS-Tools/blob/960ace8d4b00ba19c1fb726fb0e62386b3580635/conda_package/mpas_tools/cime/constants.py#L23

xylar: Ah, great!

compass/ocean/tests/isomip_plus/initial_state.py
cbegeman: If that's the case, then wouldn't we want to assign the ssh to the recomputed land ice draft for both thin film and non-thin film cases?

xylar: We could certainly do that.  It would dovetail better with the coupling strategy.

The reason I had been working with SSH/draft as the known quantity and land-ice pressure as the one to adjust was that this is more how we operate with static cavities with realistic topography (e.g. from  BedMachine).  Without wetting and drying, you can end up with too thin a water column if you are using land-ice pressure to determine SSH.  But we could either adjust the grounding line if this happens.  We can't adjust the minimum column thickness unless we also change the land-ice pressure because otherwise the pressure will just slam the column back down.

cbegeman:
@xylar Does the latest commit look like what you have in mind? It results in these water column thicknesses:
Ocean0:
image
thin_film_Ocean0:
image

compass/ocean/tests/isomip_plus/initial_state.py
cbegeman: I modify the bottom depth to maintain min water column thickness here: https://github.com/MPAS-Dev/compass/blob/b977252d360f52d9e8dc7c2af2d7c5efefe143ec/compass/ocean/tests/isomip_plus/initial_state.py#L156-L167 But I only do this for the initial state so it wouldn't guarantee for a time-varying pressure that the water column thickness would stay above the threshold without wetting and drying.

xylar: Right.  In my experience, the only to have things behave robustly without wetting-and-drying has been to treat the water-column thickness as the thing you know and the land-ice pressure as the thing you can adjust (in the poorly named "ssh adjustment" step).  Even if you thicken the water column in the initial condition, if the land-ice pressure is too high, it will just slam right back down and try to evacuate the water column.

xylar:
> Does the latest commit look like what you have in mind?

Yes exactly, that looks great to me!  Are you happy with it thought?  I definitely want to make sure we're on the same page.

cbegeman:
@xylar I'm fine with that. I can add the same thing to the time-varying forcing. I assume we keep the floating fraction the same, to mimic a case where the GL is different in MALI from MPAS-O. Do you agree?

xylar:
Right, I agree.  We want to treat the floating fraction as something that is prescribed or supplied by MALI, so it doesn't get updated based on whether MPAS-Ocean thinks the cell is wet or dry.

cbegeman:
@xylar Following up on your question earlier: Wetting and drying is on for thin-film cases, forward and ssh-adjustment steps, so that is not the issue with ssh_adjustment failures.

cbegeman:
@xylar Just wanted to make sure this was still on your radar to review. No particular rush.

xylar:
Yes, I think this was where we were discussing going back to one strategy where we trust `landIcePressure` and dig when necessary to prevent drying (for non-W&D cases).  Did you want to have this reviewed as it is or for me to make that change here?

cbegeman:
> Yes, I think this was where we were discussing going back to one strategy where we trust landIcePressure and dig when necessary to prevent drying (for non-W&D cases). Did you want to have this reviewed as it is or for me to make that change here?

@xylar Right, that's worth chatting about. I was assuming we merge this PR and tackle the digging strategy in a separate PR.
If you want to tackle it in this PR, that's fine by me. It would be good to have you review at least the thin film approach here because that's what I'll use in my W&D testing in the very near future.

xylar:
Okay, let's just focus on the thin film here and I'll make a follow-up PR to change the strategy for the non-thin-film cases.  I could do a rush review tonight but it would be better if it's on my list for tomorrow.

compass/ocean/tests/isomip_plus/initial_state.py
```diff
@@ -125,17 +132,21 @@ def _compute_initial_condition(self):
...
+                land_ice_pressure=land_ice_pressure,
+                modify_mask=ds.bottomDepth > 0.)
+            ds['ssh'] = np.maximum(land_ice_draft, -ds.bottomDepth)
+        else:
+            land_ice_draft = ds.ssh
+            land_ice_pressure = compute_land_ice_pressure_from_draft(
+                land_ice_draft=land_ice_draft, modify_mask=land_ice_draft < 0.)
 
-        ds['bottomDepth'] = -ds.bottomDepthObserved
+        ds['landIcePressure'] = land_ice_pressure
```
xylar, line 149:
```suggestion
        ds['landIcePressure'] = land_ice_pressure
        ds['landIceDraft'] = land_ice_draft
```
This seems to be needed in `_write_time_varying_forcing()` and it is also used in MPAS-Ocean forward mode.

xylar: @cbegeman, you marked this as resolved but it seems like the code hasn't changed.  Are there changes you have locally that you maybe need to push?

cbegeman: Done. Thanks for catching that.

compass/ocean/tests/isomip_plus/initial_state.py
```diff
@@ -383,10 +407,27 @@ def _write_time_varying_forcing(self, ds_init):
...
+            land_ice_thickness = ds_init.landIceThickness
+            land_ice_pressure = compute_land_ice_pressure_from_thickness(
+                land_ice_thickness=land_ice_thickness,
+                modify_mask=ds_init.landIceFraction > 0.,
+                land_ice_density=ice_density)
+            land_ice_draft = compute_land_ice_draft_from_pressure(
+                land_ice_pressure=land_ice_pressure,
+                modify_mask=ds_init.bottomDepth > 0.)
+            land_ice_draft = np.maximum(land_ice_draft, -ds_init.bottomDepth)
+            land_ice_draft = land_ice_draft.transpose()
```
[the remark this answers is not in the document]
cbegeman: Done. Good point.

compass/ocean/tests/isomip_plus/initial_state.py
```diff
@@ -229,8 +250,8 @@ def _plot(self, ds):
 
         ds['oceanFracObserved'] = \
             ds['oceanFracObserved'].expand_dims(dim='Time', axis=0)
-        ds['landIcePressure'] = \
-            ds['landIcePressure'].expand_dims(dim='Time', axis=0)
+        ds['landIceThickness'] = \
+            ds['landIceThickness'].expand_dims(dim='Time', axis=0)
         ds['landIceGroundedFraction'] = \
             ds['landIceGroundedFraction'].expand_dims(dim='Time', axis=0)
         ds['bottomDepth'] = ds['bottomDepth'].expand_dims(dim='Time', axis=0)
```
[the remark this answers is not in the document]
cbegeman: Done. It passes with the fix I added in 73e20b94f8

cbegeman:
@xylar FYI: I'm finding that the run hangs at the performance step of `time_varying_Ocean0`. I'll ping you when I figure out why that is.

xylar:
@cbegeman, I think the hanging itself isn't related to this PR.  It seems to happen sometimes when MPAS-Ocean fails with state validation errors but I don't understand why this leads to hanging as opposed to MPAS-Ocean returning an error.

But it's worth trying to fix the state validation issues so the hanging doesn't happen.

cbegeman:
Closed in favor of https://github.com/E3SM-Project/E3SM/pull/6501

Thanks for reviewing @jonbob @darincomeau @xylar!

cbegeman:
Oops! comment posted on the wrong PR

cbegeman:
@xylar I think we can get this in now with the non-thin film cases passing. I'm going to open subsequent PRs to (1) remove land ice draft forcing along with E3SM PR and (2) support thin film cases with [different E3SM PR](https://github.com/E3SM-Ocean-Discussion/E3SM/pull/119)

xylar (commented):
@cbegeman, this is a lot of work you've put in here!  For the most part, I'm going to trust your testing, though I'd like to verify that I can run all the tests myself.

I have one small change I'd suggest.  Let me know what you think.

There are some hard-coded pieces here (like freezing-temperature coefficients) that we'll want to do a better job of generalizing in Polaris but I'm going to treat the Compass version of ISOMIP+ as being in triage mode -- whatever works!

I'll let you know how my testing goes.

xylar:
If there's a chance next week, let's chat about how we want to handle porting this over to Polaris.  I'd like to get moving on that soon (particularly with https://github.com/E3SM-Project/polaris/pull/151, which is well over a year old).

compass/ocean/tests/isomip_plus/initial_state.py
```diff
@@ -149,44 +189,52 @@ def _compute_initial_condition(self):
...
-        _, thin_film_temp = np.meshgrid(ds.refZMid, freezing_temp)
-        _, thin_film_mask = np.meshgrid(ds.refZMid, thin_film_mask)
-        thin_film_temp = np.expand_dims(thin_film_temp, axis=0)
+
         if self.vertical_coordinate == 'single_layer':
-            ds['temperature'] = init_bot_temp * xr.ones_like(frac)
-            ds['salinity'] = init_bot_sal * xr.ones_like(frac)
+            # Initialize constant T,S
+            ds['temperature'] = init_bot_temp * xr.ones_like(ds.zmid)
+            ds['salinity'] = init_bot_sal * xr.ones_like(ds.zmid)
```
xylar, line 208: I'm seeing errors in my testing.  This needs to be:
```suggestion
            ds['temperature'] = init_bot_temp * xr.ones_like(ds.zMid)
            ds['salinity'] = init_bot_sal * xr.ones_like(ds.zMid)
```

xylar: `frac` also needs to be defined here, since it's returned below.

xylar: I would change this whole if/else as follows:
``` python
        if self.vertical_coordinate == 'single_layer':
            # Initialize constant T,S with bottom values
            frac = xr.ones_like(ds.zMid)
        else:
            # Initialize T,S as linear functions with max depth
            max_bottom_depth = -config.getfloat('vertical_grid',
                                                'bottom_depth')
            frac = (0. - ds.zMid) / (0. - max_bottom_depth)
        ds['temperature'] = \
            (1.0 - frac) * init_top_temp + frac * init_bot_temp
        ds['salinity'] = \
            (1.0 - frac) * init_top_sal + frac * init_bot_sal
```

cbegeman:
@xylar I was thinking of leaving the thin film cases for fix-up in a later PR. And honestly I'm thinking of removing all the single layer cases because I don't need them anymore now that we support baroclinic W&D (with PRs that are not yet merged to E3SM)

xylar:
@cbegeman, okay, that's fine.  It's just hard for me to keep track of what is supposed to work and what is expected not to work, which is also making porting over to Polaris difficult for me.  Let's chat about it this week.

cbegeman:
@xylar Sorry for the confusion. I went ahead and removed the single layer cases. They should all pass now but I need to retest; I'll let you know if I find one that doesn't.

compass/ocean/tests/isomip_plus/initial_state.py
cbegeman: Resolved with removal of single layer cases

xylar (approved):
@cbegeman, all the isomip_plus tests pass for me now, too.  Thanks for working so hard on this!

xylar:
I won't merge this yet, in case you wanted to wait on anything (you assigned it to yourself).

cbegeman:
@xylar Thank you for reviewing!

xylar:
Nice to get this in!

PR #728, opened 2023-11-11
New Mesh: ECwISC30to60E3r4

xylar:
Long name: ECwISC30to60L64E3SMv3r4

This mesh differs from ECwISC30to60E3r2 (#689) only in having a smaller maximum Haney Number (rx1) of 5 instead of 10.

The `mesh` test case is a symlink to that of the `ECwISC30to60E3r2` to guarantee that they are identical.

Initial condition, dynamic adjustment and files for E3SM will be on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/ecwisc30to60e3r4
```

This mesh name is being used for this instead of for topography smoothing (in #715) because we decided not to pursue any simulations with that version of the mesh.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/global_ocean.cfg
+++ b/compass/ocean/tests/global_ocean/global_ocean.cfg
@@ -47,6 +47,11 @@ btr_dt_per_km = 1.5
 
 # Maximum allowed Haney number for configurations with ice-shelf cavities
 rx1_max = 20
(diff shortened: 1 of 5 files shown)


xylar:
We do not seem to need to decrease the Haney Number from 10 to 5 after all.

Issue #729, opened 2023-11-13
Major bug in `landice/antarctica/mesh_gen`


trhille:
When running the `landice/antarctica/mesh_gen` case using default config settings, the mesh that is created is unusable:
![image](https://github.com/MPAS-Dev/compass/assets/17446278/a768cf44-6970-4a1a-9a7e-f8651142a593)

Thanks to @cwsmith for catching this!

xylar:
Looks fine to me!

trhille:
My initial thought was that that this was caused by the new cell culler, but the problem persists even when I checkout commit `51e65a5e342aa34bec97fb5478be7b75abc8be6e` and use `dev_compass_1.2.0-alpha.5`. In the [testing](https://github.com/MPAS-Dev/compass/pull/590#issuecomment-1503626739)section of the mesh generation refactor PR, I reported that the mesh was identical to the code before the refactor. However, there are numerous clean-up commits after that testing that could be responsible for the issue.

trhille:
Further investigation reveals that `antarctica_8km_2020_10_20_floodFillMask.nc` contains only isolated pockets of ice, which indicates an issue with the flood fill.

trhille:
Testing indicates that this was introduced by https://github.com/MPAS-Dev/compass/pull/590/commits/8b327bd224c27d3590e247002588469544701c51.

xylar:
@trhille, very sorry about that!  I don't understand why it's a problem but let's revert it!

PR #730, opened 2023-11-13
Fix `landice/antarctica/mesh_gen`

trhille:
This reverts commit 8b327bd224c27d3590e247002588469544701c51, which inadvertently broke the `landice/antarctica/mesh_gen` case and could lead to unexpected behavior elsewhere.

Checklist

This closes https://github.com/MPAS-Dev/compass/issues/729.

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -529,10 +529,10 @@ def build_cell_width(self, section_name, gridded_dataset,
     geom_points, geom_edges = set_rectangular_geom_points_and_edges(*bnds)
 
     # Remove ice not connected to the ice sheet.
-    flood_mask = gridded_flood_fill(thk) == 0
-    thk[flood_mask] = 0.0
-    vx[flood_mask] = 0.0
-    vy[flood_mask] = 0.0
+    flood_mask = gridded_flood_fill(thk)
+    thk[flood_mask == 0] = 0.0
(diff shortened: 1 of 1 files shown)


trhille:
Testing
-------
The mesh created using the current head of `origin/main` is just a collection of islands:
image

Using this branch results in the expected behavior:
image

compass/landice/mesh.py
```diff
@@ -529,10 +529,10 @@ def build_cell_width(self, section_name, gridded_dataset,
...
 
     # Remove ice not connected to the ice sheet.
-    flood_mask = gridded_flood_fill(thk) == 0
-    thk[flood_mask] = 0.0
-    vx[flood_mask] = 0.0
-    vy[flood_mask] = 0.0
+    flood_mask = gridded_flood_fill(thk)
+    thk[flood_mask == 0] = 0.0
+    vx[flood_mask == 0] = 0.0
+    vy[flood_mask == 0] = 0.0
```
xylar, line 535: I don't understand why the original change didn't work.  I would think it would be identical to the following, which would be identical to the code after reversion.

``` python
    flood_mask = gridded_flood_fill(thk)
    mask = (flood_mask == 0)
    thk[mask] = 0.0
    vx[mask] = 0.0
    vy[mask] = 0.0
```

But it's not worth exploring further.

xylar (approved):

PR #732, opened 2023-11-19
Fix a bug in how `landIceMask` is computed

xylar:
Before this fix, cells that were more than half land according to the remapped topography dataset but had been included in the ocean (through critical passages) were mistakenly being categorized as land ice simply because they were "not ocean".

With this fix, the `landIceMask` is computed based on where the `landIceFracObserved` from the remapped topography dataset is greater than 0.5.  A new function is needed to compute this mask because this definition is not the same as the mask used to cull the mesh when ice-shelf cavities are excluded from the ocean domain.

Checklist

### DIFF
--- a/compass/ocean/mesh/cull.py
+++ b/compass/ocean/mesh/cull.py
@@ -413,9 +413,8 @@ def _cull_mesh_with_logging(logger, with_cavities, with_critical_passages,
 
     if with_cavities:
         if has_remapped_topo:
-            _land_mask_from_topo(with_cavities=False,
-                                 topo_filename='topography_culled.nc',
-                                 mask_filename='ice_coverage.nc')
+            _land_ice_mask_from_topo(topo_filename='topography_culled.nc',
+                                     mask_filename='ice_coverage.nc')
         else:
             _land_mask_from_geojson(with_cavities=False,
                                     process_count=process_count,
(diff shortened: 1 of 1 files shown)


xylar:
In the process of debugging topography smoothing, I noticed strange temperature values in cells around the North Atlantic.  I traced this to an incorrect calculation of the `landIceMask` that incorporated land cells (that had been turned to ocean by critical passages) as well as land-ice cells.
land_ice_mask_natl

This is a serious bug that needs to be fixed in all E3SMv3 meshes with ice-shelf cavities!

xylar:
## Testing

With this bug fix, I'm no longer seeing this issue with the ECwISC mesh:

land_ice_mask_natl

proteanplanet:
@xylar Good catch.  Thanks for your careful checking - better to find a bug than have it unknown.

rljacob:
Were the temperature values strange but not unphysical?   (Does MPAS-O issue a warning for unphysical temps?)

xylar:
> Were the temperature values strange but not unphysical? (Does MPAS-O issue a warning for unphysical temps?)

The temperature values were different from the surrounding ocean but not unphysical (constant -1.8 C, which is the initial value we use under ice shelves).

MPAS-Ocean doesn't have any sanity checks that I'm aware of for unphysical temperature values short of NaNs.

PR #733, opened 2023-11-19
New Mesh: ECwISC30to60E3r4 

xylar:
Long name: ECwISC30to60L64E3SMv3r4

As with previous versions of the mesh, this Eddy Closure (EC) mesh has:
* 30 km resolution at the equator
* 60 km resolution at mid latitudes
* 35 km resolution near the poles

This mesh differs from ECwISC30to60E3r2 (#689) in having topography smoothing using a Gaussian filter with a characteristic length scale of 100 km.

Because of the bug fix in #732, the horizontal mesh has been recreated as well and is not identical to the `ECwISC30to60E3r2` mesh.

Initial condition, dynamic adjustment and files for E3SM will be on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/ecwisc30to60e3r4
```

Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #736, opened 2023-11-27
New Mesh: ECwISC30to60E3r5

xylar:
Long name: ECwISC30to60L64E3SMv3r5

This mesh differs from ECwISC30to60E3r2 (#689) primarily in having the the bug fix in #732, but also it has been produced with updated versions of the `mpas_tools` and `pyremap` packages, leading to small changes in the topography and coast line.

It differs from ECwISC30to60E3r4 (#733) in not having additional topographic smoothing.

```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/ecwisc30to60e3r5
```

Checklist


xylar:
Will not use

PR #737, opened 2023-11-27
Port the full MISMIP+ test group from the old vesrion of compass 

andrewdnolan:
__Description__: 
Currently the only test case within the MISMIP+ test group is a `SmokeTest`, which uses an archived "spun-up" initial condition. The goal of the PR is add more (if not all) of the experiments from the MISMIP+ experiments as test cases for MALI ([Asay-Davis et al. 2016](https://gmd.copernicus.org/articles/9/2471/2016/gmd-9-2471-2016.pdf)). 

----

__Checklist__:

### DIFF
--- a/compass/landice/tests/mismipplus/__init__.py
+++ b/compass/landice/tests/mismipplus/__init__.py
@@ -1,5 +1,6 @@
-from compass.testgroup import TestGroup
+from compass.landice.tests.mismipplus.makemeshtest import MakeMeshTest
 from compass.landice.tests.mismipplus.smoke_test import SmokeTest
+from compass.testgroup import TestGroup
 
 
 class MISMIPplus(TestGroup):
@@ -15,3 +16,6 @@ class MISMIPplus(TestGroup):
         super().__init__(mpas_core=mpas_core, name='mismipplus')
 
         self.add_test_case(SmokeTest(test_group=self))
+
+        # Dummy test case for testing the mesh stepup.
+        self.add_test_case(MakeMeshTest(test_group=self))
--- a/compass/landice/tests/mismipplus/makemeshtest/__init__.py
+++ b/compass/landice/tests/mismipplus/makemeshtest/__init__.py
@@ -1,0 +1,14 @@
+from compass.landice.tests.mismipplus.setup_mesh import SetupMesh
+from compass.testcase import TestCase
+
+
+class MakeMeshTest(TestCase):
+    """
+    """
+    def __init__(self, test_group):
+
+        name = "MakeMeshTest"
+
+        super().__init__(test_group=test_group, name=name)
+
+        self.add_step(SetupMesh(test_case=self))
--- a/compass/landice/tests/mismipplus/mismipplus.cfg
+++ b/compass/landice/tests/mismipplus/mismipplus.cfg
@@ -1,0 +1,20 @@
+# config options for the mismipplus test cases
+[mismipplus]
+
+resolution = 20
+# size (in cells) for the mesh
+nx = 284
+ny = 44
+
+# resolution
+dc = 2309.401077
+
+# wether the mesh should be periodic, probably don't even need this option
+# since I don't think it should ever be periodic
+nonperiodic = true
+
+# Number of vertical levels
+levels = 10
+
+# Initial ice thickness
+init_thickness = 100
--- a/compass/landice/tests/mismipplus/setup_mesh.py
+++ b/compass/landice/tests/mismipplus/setup_mesh.py
@@ -32,8 +32,8 @@ class SetupMesh(Step):
         """
         super().__init__(test_case=test_case, name='setup_mesh')
 
-        #
-        for filename in ['mesh.nc', 'graph.info', 'landice_grid.nc']:
+        # Files to be created as part of the this step
+        for filename in ['mpas_grid.nc', 'graph.info', 'landice_grid.nc']:
             self.add_output_file(filename=filename)
 
     def run(self):
@@ -63,27 +63,26 @@ class SetupMesh(Step):
         # per MISMIP+ converion
         ds_mesh = shift_origin_to_lower_left(ds_mesh)
 
-        write_netcdf(ds_mesh, 'mesh.nc')
+        write_netcdf(ds_mesh, 'mpas_grid.nc')
 
         # just use get, not getint since var need to be string
         # anyway for subprocess call
         levels = section.get('levels')
         args = ['create_landice_grid_from_generic_MPAS_grid.py',
-                '-i', 'mesh.nc',
-                '-o', 'landice_grid_template.nc',
+                '-i', 'mpas_grid.nc',
+                '-o', 'landice_grid.nc',
                 '-l', levels,
                 '--diri',
                 '--thermal']
 
         check_call(args, logger)
 
-        make_graph_file(mesh_filename='landice_grid_template.nc',
+        make_graph_file(mesh_filename='landice_grid.nc',
                         graph_filename='graph.info')
 
         # create the initial conditions for the MISMIP+ spinup expr
         _setup_MISMPPlus_IC(config, logger,
-                            input_file='landice_grid_template.nc',
-                            output_file='landice_grid.nc')
+                            filename='landice_grid.nc')
 
 
 def shift_origin_to_lower_left(ds_mesh):
@@ -102,10 +101,11 @@ def shift_origin_to_lower_left(ds_mesh):
     # because getting the first interior cell edge
     # from the global min/max wont' work
     unique_xs = np.unique(ds_mesh.xCell.data)
-    unique_ys = np.unique(ds_mesh.xCell.data)
+    # unique_ys = np.unique(ds_mesh.yCell.data)
 
     xShift = -1.0 * unique_xs.min()
-    yShift = -1.0 * unique_ys.min()
+    # yShift = -1.0 * unique_ys.min()*5.0
+    yShift = 40e3 - (ds_mesh.yCell.max() + ds_mesh.yCell.min()) / 2
 
     # Shift all spatial points by calculated shift
     for loc in ['Cell', 'Edge', 'Vertex']:
@@ -173,7 +173,7 @@ def __mismipplus_bed(x, y):
     return z_b
 
 
-def _setup_MISMPPlus_IC(config, logger, input_file, output_file):
+def _setup_MISMPPlus_IC(config, logger, filename):
     """
     Add the inital condition for the MISMIP+ spinup to the given MPAS mesh file
 
@@ -185,11 +185,9 @@ def _setup_MISMPPlus_IC(config, logger, input_file, output_file):
     logger : logging.Logger
         A logger for output from the step
 
-    input_file : str
-        NetCDF file to place the MISMIP+ ICs into
+    filename : str
+        NetCDF file to place the MISMIP+ ICs in
 
-    output_file : str
-        .....
     """
 
     # Hard code some parameters from Table 1. of Asay-Davis et al. 2016
@@ -207,7 +205,7 @@ def _setup_MISMPPlus_IC(config, logger, input_file, output_file):
     init_thickness = section.getfloat('init_thickness')
 
     # open the file
-    src = xr.open_dataset(input_file, engine='netcdf4')
+    src = xr.open_dataset(filename)
 
     # Set the bedTopography
     src['bedTopography'].loc[:] = __mismipplus_bed(src.xCell, src.yCell)
@@ -241,7 +239,6 @@ def _setup_MISMPPlus_IC(config, logger, input_file, output_file):
     src['layerThicknessFractions'].loc[:] = 1.0 / float(nVertLevels)
 
     # Write the dataset to disk
-    # NOTE: Do I need flags for the write mode?
-    write_netcdf(src, output_file, engine='netcdf4')
+    write_netcdf(src, filename)
 
-    print(f'Successfully added MISMIP+ initial conditions to: {output_file}')
+    print(f'Successfully added MISMIP+ initial conditions to: {filename}')
--- a/compass/landice/tests/mismipplus/setup_mesh.py
+++ b/compass/landice/tests/mismipplus/setup_mesh.py
@@ -70,18 +70,20 @@ class SetupMesh(Step):
         levels = section.get('levels')
         args = ['create_landice_grid_from_generic_MPAS_grid.py',
                 '-i', 'mesh.nc',
-                '-o', 'landice_grid.nc',
+                '-o', 'landice_grid_template.nc',
                 '-l', levels,
                 '--diri',
                 '--thermal']
 
         check_call(args, logger)
 
-        make_graph_file(mesh_filename='landice_grid.nc',
+        make_graph_file(mesh_filename='landice_grid_template.nc',
                         graph_filename='graph.info')
 
         # create the initial conditions for the MISMIP+ spinup expr
-        _setup_MISMPPlus_IC(config, logger, 'landice_grid.nc')
+        _setup_MISMPPlus_IC(config, logger,
+                            input_file='landice_grid_template.nc',
+                            output_file='landice_grid.nc')
 
 
 def shift_origin_to_lower_left(ds_mesh):
@@ -171,7 +173,7 @@ def __mismipplus_bed(x, y):
     return z_b
 
 
-def _setup_MISMPPlus_IC(config, logger, filename):
+def _setup_MISMPPlus_IC(config, logger, input_file, output_file):
     """
     Add the inital condition for the MISMIP+ spinup to the given MPAS mesh file
 
@@ -183,9 +185,11 @@ def _setup_MISMPPlus_IC(config, logger, filename):
     logger : logging.Logger
         A logger for output from the step
 
-    filename : str
+    input_file : str
         NetCDF file to place the MISMIP+ ICs into
 
+    output_file : str
+        .....
     """
 
     # Hard code some parameters from Table 1. of Asay-Davis et al. 2016
@@ -203,37 +207,41 @@ def _setup_MISMPPlus_IC(config, logger, filename):
     init_thickness = section.getfloat('init_thickness')
 
     # open the file
-    src = xr.open_dataset(filename)
+    src = xr.open_dataset(input_file, engine='netcdf4')
 
     # Set the bedTopography
-    src['bedTopography'] = __mismipplus_bed(src.xCell, src.yCell)
+    src['bedTopography'].loc[:] = __mismipplus_bed(src.xCell, src.yCell)
 
     # Set the ice thickness
-    src['thickness'] = xr.where(src.xCell < xcalve, init_thickness, 0.)
+    src['thickness'].loc[:] = xr.where(src.xCell < xcalve, init_thickness, 0.)
 
     # Convert SMB from m/yr to kg/m2/s
     accum *= rhoi / spy
     # Set the surface mass balance
-    src['thickness'] = xr.where(src.xCell > xcalve, accum, -100.)
+    src['sfcMassBal'].loc[:] = xr.where(src.xCell > xcalve, accum, -100.)
 
     # Boolean maks for indexes which correspond to the N/S boudnary of mesh
     mask = (src.yCell == src.yCell.min()) | (src.yCell == src.yCell.max())
     # Set the velocity boundary conditions
-    src['dirichletVelocityMask'] = xr.where(mask, 1, 0)
+    src['dirichletVelocityMask'].loc[:] = xr.where(mask, 1, 0)
 
     # SKIPPING setting the initial velocities because they are already zero.
 
     # convert to MPAS units
     C /= spy**(1.0 / 3.0)
 
-    # Set the effectivePressure using a Weertman power law.
-    src['effectivePressure'] = C
+    # Use `.loc[:]` for setting `effectivePressure` and
+    # `layerThicknessFractions` since we are setting the fields with scalar
+    # values. Needed to maintain the existing variables dimension/coords
 
-    # Set up the layerThicknessFractions, so that layers are evenly distributed
-    src['layerThicknessFractions'] = 1 / float(nVertLevels)
+    # Set the effectivePressure using a Weertman power law
+    src['effectivePressure'].loc[:] = C
+
+    # the layerThicknessFractions, so that layers are evenly distributed
+    src['layerThicknessFractions'].loc[:] = 1.0 / float(nVertLevels)
 
     # Write the dataset to disk
     # NOTE: Do I need flags for the write mode?
-    src.to_netcdf(filename)
+    write_netcdf(src, output_file, engine='netcdf4')
 
-    print(f'Successfully added MISMIP+ initial conditions to: {filename}')
+    print(f'Successfully added MISMIP+ initial conditions to: {output_file}')
--- a/compass/landice/tests/mismipplus/setup_mesh.py
+++ b/compass/landice/tests/mismipplus/setup_mesh.py
@@ -1,0 +1,239 @@
+import numpy as np
+import xarray as xr
+from mpas_tools.io import write_netcdf
+from mpas_tools.logging import check_call
+from mpas_tools.mesh.conversion import convert, cull
+from mpas_tools.planar_hex import make_planar_hex_mesh
+
+from compass.model import make_graph_file
+from compass.step import Step
+
+
+class SetupMesh(Step):
+    """
+    A step for creating a mesh and initial condition for MISMIP+ test cases
+
+    Parameters
+    ----------
+    resolution : int
+        The resolution of the test case
+    """
+    def __init__(self, test_case):
+        """
+        Create the step
+
+        Parameters
+        ----------
+        testcase : compass.TestCase
+            The test case this step belongs to
+
+        resolution : int
+            The resolution of the test case
+        """
+        super().__init__(test_case=test_case, name='setup_mesh')
+
+        #
+        for filename in ['mesh.nc', 'graph.info', 'landice_grid.nc']:
+            self.add_output_file(filename=filename)
+
+    def run(self):
+        """
+        Run this step of the test case
+        """
+        logger = self.logger
+        config = self.config
+
+        section = config['mismipplus']
+
+        nx = section.getint('nx')
+        ny = section.getint('ny')
+
+        dc = section.getfloat('dc')
+
+        nonperiodic = section.getboolean('nonperiodic')
+
+        ds_mesh = make_planar_hex_mesh(nx=nx, ny=ny, dc=dc,
+                                       nonperiodic_x=nonperiodic,
+                                       nonperiodic_y=nonperiodic)
+
+        ds_mesh = cull(ds_mesh, logger=logger)
+        ds_mesh = convert(ds_mesh, logger=logger)
+
+        # custom function to shift the orgin to the lower left
+        # per MISMIP+ converion
+        ds_mesh = shift_origin_to_lower_left(ds_mesh)
+
+        write_netcdf(ds_mesh, 'mesh.nc')
+
+        # just use get, not getint since var need to be string
+        # anyway for subprocess call
+        levels = section.get('levels')
+        args = ['create_landice_grid_from_generic_MPAS_grid.py',
+                '-i', 'mesh.nc',
+                '-o', 'landice_grid.nc',
+                '-l', levels,
+                '--diri',
+                '--thermal']
+
+        check_call(args, logger)
+
+        make_graph_file(mesh_filename='landice_grid.nc',
+                        graph_filename='graph.info')
+
+        # create the initial conditions for the MISMIP+ spinup expr
+        _setup_MISMPPlus_IC(config, logger, 'landice_grid.nc')
+
+
+def shift_origin_to_lower_left(ds_mesh):
+    """
+    Shift the orgin of the to lower left corner
+
+    Parameters
+    ----------
+    ds_mesh : xr.core.dataset.Datatset
+        The mesh to be shifted
+    """
+
+    # Why do you need unique grid points?
+    # that shouldn't alter the global min........
+    # I guess so that the masking method is consitent throughout,
+    # because getting the first interior cell edge
+    # from the global min/max wont' work
+    unique_xs = np.unique(ds_mesh.xCell.data)
+    unique_ys = np.unique(ds_mesh.xCell.data)
+
+    xShift = -1.0 * unique_xs.min()
+    yShift = -1.0 * unique_ys.min()
+
+    # Shift all spatial points by calculated shift
+    for loc in ['Cell', 'Edge', 'Vertex']:
+        ds_mesh[f'x{loc}'] = ds_mesh[f'x{loc}'] + xShift
+        ds_mesh[f'y{loc}'] = ds_mesh[f'y{loc}'] + yShift
+
+    # WHL :
+    #   Need to adjust geometry along top and bottom boundaries to get flux
+    #   correct there. Essentially, we only want to model the interior half of
+    #   those cells. Adding this here because we only want to do this if it
+    #   hasn't been done before. This method is assuming a periodic_hex mesh!
+
+    # Reduce the cell areas by half along the N/S boundary
+
+    # Boolean maks for indexes which correspond to the N/S boudnary of mesh
+    mask = ((ds_mesh.yCell == ds_mesh.yCell.min()) |
+            (ds_mesh.yCell == ds_mesh.yCell.max()))
+
+    ds_mesh['areaCell'] = xr.where(mask,
+                                   ds_mesh.areaCell * 0.5,
+                                   ds_mesh.areaCell)
+
+    # Reduce the edge lengths by half along the N/S boundary
+    unique_ys_edge = np.unique(ds_mesh.yEdge.data)
+
+    # Boolean maks for indexes edges on the N/S boudnary of mesh
+    mask = ((ds_mesh.yEdge == unique_ys_edge[0]) |
+            (ds_mesh.yEdge == unique_ys_edge[-1]))
+    # WHL: zero out the edges on the boundary
+    #      (not necessary because velocity will also be zero)
+    ds_mesh['dvEdge'] = xr.where(mask, 0.0, ds_mesh.dvEdge)
+
+    # Boolean mask for the indexed of edges between N/S boundary cells
+    mask = ((ds_mesh.yEdge == unique_ys_edge[1]) |
+            (ds_mesh.yEdge == unique_ys_edge[-2]))
+    # cut length in half for edges between boundary cells
+    ds_mesh['dvEdge'] = xr.where(mask, ds_mesh.dvEdge * 0.5, ds_mesh.dvEdge)
+
+    return ds_mesh
+
+
+def __mismipplus_bed(x, y):
+    x = x / 1.e3      # m to km
+    y = y / 1.e3      # m to km
+    B0 = -150.        # m
+    B2 = -728.8       # m
+    B4 = 343.91       # m
+    B6 = -50.57       # m
+    x_bar = 300.      # km
+    x_tilde = x / x_bar
+    dc = 500.         # m
+    fc = 4.           # km
+    wc = 24.          # km
+    Ly = 80.          # km
+    Bmax = -720.      # m
+
+    B_x = B0 + B2 * x_tilde**2 + B4 * x_tilde**4 + B6 * x_tilde**6
+    B_y = dc / (1 + np.exp(-2 * (y - Ly / 2 - wc) / fc)) +\
+        dc / (1 + np.exp(2 * (y - Ly / 2 + wc) / fc))
+
+    Bsum = B_x + B_y
+
+    z_b = np.maximum(Bsum, Bmax)  # Eqn 1. (Asay-Davis et al. 2016)
+
+    return z_b
+
+
+def _setup_MISMPPlus_IC(config, logger, filename):
+    """
+    Add the inital condition for the MISMIP+ spinup to the given MPAS mesh file
+
+    Parameters
+    ----------
+    config : comass.config.CompassConfigParser
+        Configuration options for this test case, ....
+
+    logger : logging.Logger
+        A logger for output from the step
+
+    filename : str
+        NetCDF file to place the MISMIP+ ICs into
+
+    """
+
+    # Hard code some parameters from Table 1. of Asay-Davis et al. 2016
+    accum = 0.3       # m^{-1}
+    C = 3.160e6       # Pa m^{-1/3} s^{1/3}
+    rhoi = 918        # kg m^{-3}
+    spy = 31556926    # s a^{-1}
+    xcalve = 640.e3   # m
+
+    # Read parameters from the .cfg file
+    section = config['mismipplus']
+    nVertLevels = section.getint('levels')
+
+    # this parameter might need to specififed in .cfg
+    init_thickness = section.getfloat('init_thickness')
+
+    # open the file
+    src = xr.open_dataset(filename)
+
+    # Set the bedTopography
+    src['bedTopography'] = __mismipplus_bed(src.xCell, src.yCell)
+
+    # Set the ice thickness
+    src['thickness'] = xr.where(src.xCell < xcalve, init_thickness, 0.)
+
+    # Convert SMB from m/yr to kg/m2/s
+    accum *= rhoi / spy
+    # Set the surface mass balance
+    src['thickness'] = xr.where(src.xCell > xcalve, accum, -100.)
+
+    # Boolean maks for indexes which correspond to the N/S boudnary of mesh
+    mask = (src.yCell == src.yCell.min()) | (src.yCell == src.yCell.max())
+    # Set the velocity boundary conditions
+    src['dirichletVelocityMask'] = xr.where(mask, 1, 0)
+
+    # SKIPPING setting the initial velocities because they are already zero.
+
+    # convert to MPAS units
+    C /= spy**(1.0 / 3.0)
+
+    # Set the effectivePressure using a Weertman power law.
+    src['effectivePressure'] = C
+
+    # Set up the layerThicknessFractions, so that layers are evenly distributed
+    src['layerThicknessFractions'] = 1 / float(nVertLevels)
+
+    # Write the dataset to disk
+    # NOTE: Do I need flags for the write mode?
+    src.to_netcdf(filename)
+
+    print(f'Successfully added MISMIP+ initial conditions to: {filename}')
--- a/compass/landice/tests/mismipplus/setup_mesh.py
+++ b/compass/landice/tests/mismipplus/setup_mesh.py
@@ -59,15 +59,15 @@ class SetupMesh(Step):
         ds_mesh = cull(ds_mesh, logger=logger)
         ds_mesh = convert(ds_mesh, logger=logger)
 
-        # custom function to shift the orgin to the lower left
-        # per MISMIP+ converion
-        ds_mesh = shift_origin_to_lower_left(ds_mesh)
+        # custom function to shift the orgin for the MISMIP+ experiments
+        ds_mesh = center_trough(ds_mesh)
 
         write_netcdf(ds_mesh, 'mpas_grid.nc')
 
-        # just use get, not getint since var need to be string
-        # anyway for subprocess call
+        # using `.get(...)`, instead of `.getint(...)` since variable needs
+        # to be string anyway for subprocess call
         levels = section.get('levels')
+
         args = ['create_landice_grid_from_generic_MPAS_grid.py',
                 '-i', 'mpas_grid.nc',
                 '-o', 'landice_grid.nc',
@@ -85,9 +85,10 @@ class SetupMesh(Step):
                             filename='landice_grid.nc')
 
 
-def shift_origin_to_lower_left(ds_mesh):
+def center_trough(ds_mesh):
     """
-    Shift the orgin of the to lower left corner
+    Shift the orgin so that the bed trough is center about the Y-axis and the
+    X-axis is shifted all the way to the left.
 
     Parameters
     ----------
@@ -101,11 +102,14 @@ def shift_origin_to_lower_left(ds_mesh):
     # because getting the first interior cell edge
     # from the global min/max wont' work
     unique_xs = np.unique(ds_mesh.xCell.data)
-    # unique_ys = np.unique(ds_mesh.yCell.data)
 
+    # Get the center of y-axis (i.e. half distance)
+    yCenter = (ds_mesh.yCell.max() + ds_mesh.yCell.min()) / 2
+
+    # Shift x-axis so that the x-origin is all the way to the left
     xShift = -1.0 * unique_xs.min()
-    # yShift = -1.0 * unique_ys.min()*5.0
-    yShift = 40e3 - (ds_mesh.yCell.max() + ds_mesh.yCell.min()) / 2
+    # Shift y-axis so that it's centered about the MISMIP+ bed trough
+    yShift = 40e3 - yCenter
 
     # Shift all spatial points by calculated shift
     for loc in ['Cell', 'Edge', 'Vertex']:
--- a/compass/landice/tests/mismipplus/smoke_test/__init__.py
+++ b/compass/landice/tests/mismipplus/smoke_test/__init__.py
@@ -1,6 +1,6 @@
-from compass.validate import compare_variables
-from compass.testcase import TestCase
 from compass.landice.tests.mismipplus.run_model import RunModel
+from compass.testcase import TestCase
+from compass.validate import compare_variables
 
 
 class SmokeTest(TestCase):


trhille (changes requested):
@andrewdnolan, this is looking good! I made some comments regarding compass conventions and whatnot, but this has come a long way in a short time.  However, I'm getting an error when running this on Perlmutter. Here's what I take to be the important points from the traceback:
```
  File "/global/cfs/cdirs/fanssie/users/trhille/compass/compass/landice/tests/mismipplus/setup_mesh.py", line 265, in _setup_MISMPPlus_IC
    src['dirichletVelocityMask'].loc[:] = xr.where(mask, 1, 0)
    ...
    ValueError: could not broadcast input array from shape (1,1,3456) into shape (1,3456,11)
    ```
It looks like it created the `landice_grid.nc` file successfully, but is throwing an error when trying to populate the mesh.

File: compass/landice/tests/mismipplus/setup_mesh.py [hunk 1]
```diff
@@ -0,0 +1,285 @@
...
+
+    # Convert SMB from m/yr to kg/m2/s
+    accum *= rhoi / spy
+    # Set the surface mass balance
+    src['sfcMassBal'].loc[:] = xr.where(src.xCell > xcalve, accum, -100.)
+
+    # Boolean maks for indexes which correspond to the N/S boudnary of mesh
+    mask = (src.yCell == src.yCell.min()) | (src.yCell == src.yCell.max())
+    # Set the velocity boundary conditions
+    src['dirichletVelocityMask'].loc[:] = xr.where(mask, 1, 0)
```
trhille, line 265: This is the line that's causing the error. It needs to be
```suggestion
    for lev in range(nVertLevels):
        src['dirichletVelocityMask'].loc[:,:,lev] = xr.where(mask, 1, 0)
```
or something similar.

andrewdnolan: Given that both data arrays in the offending line have the `nCells` coordinate, `xarray` _should_ automatically align the coordinates.  

It looks like this is a bug that was resolved as of xarray version `2023.9.0`. Here's the issue describing the bug and the PR which fixed it: 
  - **Issue**: https://github.com/pydata/xarray/issues/7030
  - **PR**: https://github.com/pydata/xarray/pull/8067 


From the discussion in the issue, it seems like: 
```
src['dirichletVelocityMask'] = xr.where(mask, 1, 0).variable
```
resolves the problem. Maybe we go that route, over the `for` loop since it's a little bit closer to what the code actually should be?

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,285 @@
...
+
+    filename : str
+        NetCDF file to place the MISMIP+ ICs in
+
+    """
+
+    # Hard code some parameters from Table 1. of Asay-Davis et al. 2016
+    accum = 0.3       # m^{-1}
+    C = 3.160e6       # Pa m^{-1/3} s^{1/3}
+    rhoi = 918        # kg m^{-3}
```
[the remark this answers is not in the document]
andrewdnolan: I'm happy to use 910, but Table 1 from [Asay-Davis et al. 2016](https://gmd.copernicus.org/articles/9/2471/2016/gmd-9-2471-2016.pdf) has 918. So I guess if we were strictly following MISMIP+ we'd use that. I guess the difference would be _really_ small in the actual simulation results. 

Since this test case is for other modelling purposes besides the actual MISMIP+ experiments, should I move it to the config file and set at 910, but add a comment that we are departing from MISMIP+ conventions? 

That way if at some point there was desire to actually run MISMIP+ there wouldn't be need to change any code?

compass/landice/tests/mismipplus/__init__.py
```diff
@@ -1,5 +1,6 @@
-from compass.testgroup import TestGroup
+from compass.landice.tests.mismipplus.makemeshtest import MakeMeshTest
```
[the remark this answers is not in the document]
andrewdnolan: Thanks for the help with the compass naming conventions! I've gone with `mesh_gen` for the test case name and renamed the testcase class to `MeshGen`. From looking at other examples that seems to be consistent with conventions, but if not let me know.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,285 @@
...
+    Shift the orgin so that the bed trough is center about the Y-axis and the
+    X-axis is shifted all the way to the left.
+
+    Parameters
+    ----------
+    ds_mesh : xr.core.dataset.Datatset
+        The mesh to be shifted
+    """
+
+    # Why do you need unique grid points?
```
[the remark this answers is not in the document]
andrewdnolan: @matthewhoffman  As you mentioned, for the translation `unique_xs.min()` is unnecessary since `xCell.min()` is equivalent, so I've just switched to the simpler option here. 

But, later on when we alter the `yEdge` length on the North/South boundaries  `unique_ys.min()` is still necessary because we need to alter both the boundary and the first interior edge on the North/South boundaries. `yCell.min()` can get us the boundary index but not the first interior, so I've left this approach there.

compass/landice/tests/mismipplus/makemeshtest/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Parameters
+        ----------
+        test_group : compass.landice.test.mismipplus.setup_mesh
+            The test group that this test case belongs to
+
+        resolution : int
+            The distance between horizontal (x-y) grid points
+        """
+        name = "MakeMeshTest"
```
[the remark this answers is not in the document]
andrewdnolan: Changed to `"mesh_gen"`.

compass/landice/tests/mismipplus/setup_mesh.py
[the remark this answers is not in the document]
matthewhoffman: I like the idea of making rhoi a cfg option for this particular test case.  It is important to be able to reproduce the prescribed MISMIP+ protocol if needed, but we otherwise prefer not to use pure ice density.  Putting it in a cfg where it can be used to automatically update both the initial condition calculations and the namelist is the safest, most flexible approach, I think.

compass/landice/tests/mismipplus/__init__.py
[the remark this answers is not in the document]
matthewhoffman: @andrewdnolan , I'm still seeing `MakeMeshTest` here.  Is the change to `mesh_gen` something you need to push yet?

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # Reduce the edge lengths by half along the N/S boundary
+    unique_ys_edge = np.unique(ds_mesh.yEdge.data)
+
+    # Boolean maks for indexes edges on the N/S boudnary of mesh
+    mask = ((ds_mesh.yEdge == unique_ys_edge[0]) |
+            (ds_mesh.yEdge == unique_ys_edge[-1]))
+    # WHL: zero out the edges on the boundary
+    #      (not necessary because velocity will also be zero)
+    ds_mesh['dvEdge'] = xr.where(mask, 0.0, ds_mesh.dvEdge)
```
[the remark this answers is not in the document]
matthewhoffman: I agree this particular manipulation seems extra unusual.  The idea was to effectively eliminate the northern half of the northernmost cells and similar for the southernmost row.  I'm skeptical this is actually affecting anything in the model, but given that we ran MISMIP+ successfully with this, my preference is to retain it.  If we run into complication stemming from it down the road, we can adjust.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+        ----------
+        testcase : compass.TestCase
+            The test case this step belongs to
+
+        resolution : int
+            The distance between horizontal grid points
+        """
+        # Defined resolutions, the value of `resolution` in the configuration
+        # file needs to a key in the dictionary, or added manually.
+        resolution_params = {'8km': {'nx': 73, 'ny': 14, 'dc': 9237.604307},
```
matthewhoffman, line 48: @trhille , for your usage of this domain for convergence testing with calving on, you will presumably want a gutter beyond the prescribed initial ice shelf calving front to allow for portions of the calving front to advance under prognostic calving without hitting the edge of the mesh.  My recollection is the `nx` values specified here give 2 extra cells (at all resolutions) beyond the MISMIP+ fixed calving front position.   

If allowing a gutter option sounds desirable now, I think that would be a new cfg option, either with units of number of cells or distance.  Do you want that included in this PR so you can use it out of the gate for your planned tests?

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+    # Read parameters from the .cfg file
+    section = config['mesh']
+    nVertLevels = section.getint('levels')
+    init_thickness = section.getfloat('init_thickness')
+
+    # open the file
+    src = xr.open_dataset(filename)
+
+    # Use `.loc[:]` for setting the initial conditions  since we are setting
+    # the fields with scalar values. This ensures values are properly broadcast
```
matthewhoffman, line 265: I'm not familiar with the `.loc` attribute, but I'm confused by this comment.  It looks the fields below are spatially varying so I don't follow the comment about scalar values here.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+    # against the fields coordinates
```
matthewhoffman, line 266:
```suggestion
    # against the field's coordinates
```

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+    # Set the bedTopography
+    src['bedTopography'].loc[:] = _mismipplus_bed(src.xCell, src.yCell)
+
+    # Set the ice thickness
+    src['thickness'].loc[:] = xr.where(src.xCell < xcalve, init_thickness, 0.)
+
+    # Convert SMB from m/yr to kg/m2/s
+    accum *= rhoi / spy
+    # Set the surface mass balance
+    src['sfcMassBal'].loc[:] = xr.where(src.xCell > xcalve, accum, -100.)
```
matthewhoffman, line 277: The original implementation here of using strong negative SMB to maintain calving position is fine for now.  @trhille , if you use these meshes for evolving calving fronts, you'll want to remove this.  We could replace this method with `calvingMask` instead (which didn't exist when the original MISMIP+ test case was written).  I don't think that needs to happen in this PR, unless you think it would be timesaving to have that implementation switched now.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+
+    # Boolean masks for indices which correspond to the N/S boundary of mesh
+    mask = (src.yCell == src.yCell.min()) | (src.yCell == src.yCell.max())
+    # Set the velocity boundary conditions
+    # NOTE: `.variable` is needed so that coordinates are properly broadcast
+    #        due to a bug in xarray, which was resolved as of 2023.9.0
+    src['dirichletVelocityMask'].loc[:] = xr.where(mask, 1, 0).variable
```
matthewhoffman, line 284: Add a comment that the prescribed north and south b.c. are free slip.  This is not a common b.c. we use in MALI and requires a special section in the albany_input.yaml file.  uReconstructX needs to be zero, but that is not explicitly set here because that's the default value when the field is created in the input file.
 
Note: the albany input file for MISMIP+ here:
https://github.com/MPAS-Dev/compass/blob/main/compass/landice/tests/mismipplus/albany_input.yaml
seems to have lost the free slip boundary condition information from the old repo:
[https://github.com/MPAS-Dev/MPAS-Model/blob/landice/develop/testing_and_setup/compass/landice/MISMIP%2B/albany_input.yaml and added to the test grou](https://github.com/MPAS-Dev/MPAS-Model/blob/landice/develop/testing_and_setup/compass/landice/MISMIP%2B/albany_input.yaml#L9-L10)
@mperego , can you advise us how to correct that?

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+    # since they are already zero.
+
+    # convert to MPAS units
+    C /= spy**(1.0 / 3.0)
+
+    # Set the effectivePressure using a Weertman power law
+    src['effectivePressure'].loc[:] = C
+
+    # Set the layerThicknessFractions so that layers are evenly distributed
+    src['layerThicknessFractions'].loc[:] = 1.0 / float(nVertLevels)
```
matthewhoffman, line 296: I don't think there is any reason why this needs to be case (I don't see a specification in the MISMIP+ paper), and if not I would prefer to stick with our default nonuniform vertical spacing.  I'm guessing this was specified in the old test case before we had implemented a standard nonuniform spacing option in our landice mesh creation script.  @trhille, what do you think? (if you interpolate from our archived MISMIP+ 2km state, this will be a difference to be aware of)

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+        The mesh to be shifted
+    """
+
+    # Get the center of y-axis (i.e. half distance)
+    yCenter = (ds_mesh.yCell.max() + ds_mesh.yCell.min()) / 2.
+
+    # Shift x-axis so that the x-origin is all the way to the left
+    xShift = -1.0 * ds_mesh.xCell.min()
+    # Shift y-axis so that it's centered about the MISMIP+ bed trough
+    yShift = 4.e4 - yCenter
```
matthewhoffman, line 152: @andrewdnolan , these shifts are slightly different from the old MISMIP+ test case, right?  If I followed things, these shifts are more sensible.  However, @trhille, be aware that if you interpolate the state from our archived MISMIP+ 2km state file, the coordinate system of these new meshes will be slightly different unless you do extra work to align them.

compass/landice/tests/mismipplus/smoke_test/__init__.py
```diff
@@ -1,11 +1,13 @@
...
 from compass.landice.tests.mismipplus.run_model import RunModel
+from compass.testcase import TestCase
+from compass.validate import compare_variables
 
 
 class SmokeTest(TestCase):
     """
     A test case for running a smoke test of the MISMIP+ configuration
+
+    This test case  uses a pre-made mesh file.
```
matthewhoffman, line 10:
```suggestion
    This test case performs a forward simulation for a short duration
    using a pre-made mesh file with a 2 km MISMIP+
    spin-up that has previously been run to steady state
```

compass/landice/tests/mismipplus/setup_mesh.py
trhille: Good catch. I think eventually we do want to replace this with `calvingMask`, but I don't care much whether that is done in this PR or in a short follow-up. I guess we might as well do it now, since I think it is an easy change.

compass/landice/tests/mismipplus/setup_mesh.py
trhille: I agree that we don't need to have uniform layer thickness. When interpolating from the archived state, I don't know if we will be interpolating 3D velocity, since the point of this (at least for now) is to use the velocity solver. We could also punt on this for now and have a config setting for 'uniform' or 'glimmer' layer spacing.

compass/landice/tests/mismipplus/mesh_gen/mesh_gen.cfg
```diff
@@ -0,0 +1,16 @@
+# config options for the mesh test cases
+[mesh]
+
+# resolution in km, valid options are 8, 4, 2, and 1 km
+resolution = 4
+
+# ice density (kg m^{-3}). MISMIP+ uses 918 (Table 1. Asay-Davis et al. 2016),
+# but MALI defaults to 910. The user can choose if they want to stricly follow
+# MISMIP+ or use the default MALI value.
+ice_density = 910.
```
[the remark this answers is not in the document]
trhille: Sure, that's fine with me if we are including the option in the config file.

compass/landice/tests/mismipplus/setup_mesh.py
trhille: Yes, good idea. I don't have a strong preference for whether it should be specified in number of cells or distance, but it should end up being ~the same distance for all resolutions, so maybe specifying distance is the simplest options.

compass/landice/tests/mismipplus/setup_mesh.py
mperego: I think you need to add back
```
    Dirichlet BCs: 
      SDBC on NS dirichlet for DOF U1 prescribe Field: dirichlet_field
```
otherwise Albany will prescribe that Dirichlet BCs to both component of the velocity.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+                         name=f'{key}_mesh_gen',
+                         subdir=f"{key}/mesh_gen")
+
+        # Files to be created as part of the this step
+        for filename in ['mpas_grid.nc', 'graph.info', 'landice_grid.nc']:
+            self.add_output_file(filename=filename)
+
+        # unpack the mesh parameters for the given resolution as set as
+        # attributes
+        self.nx, self.ny, self.dc = resolution_params[key].values()
```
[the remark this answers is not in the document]
andrewdnolan: Good catch, I didn't necessarily think of this. But it turns out that since python 3.6+ dictionaries are now ordered in "insertion order". [Here's a description](https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation) of the changes that were implemented: 

> The [dict](https://docs.python.org/3.6/library/stdtypes.html#typesmapping) type now uses a “compact” representation based on [a proposal by Raymond Hettinger](https://mail.python.org/pipermail/python-dev/2012-December/123028.html) which was [first implemented by PyPy](https://morepypy.blogspot.com/2015/01/faster-more-memory-efficient-and-more.html). The memory usage of the new [dict()](https://docs.python.org/3.6/library/stdtypes.html#dict) is between 20% and 25% smaller compared to Python 3.5.
> 
> The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon (this may change in the future, but it is desired to have this new dict implementation in the language for a few releases before changing the language spec to mandate order-preserving semantics for all current and future Python implementations; this also helps preserve backwards-compatibility with older versions of the language where random iteration order is still in effect, e.g. Python 3.5).

But given that it says the order preserving aspect shouldn't be relied upon, I'll switch to an `OrderedDict` so that I can keep the concise implementation.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+
+        section = config['mesh']
+
+        # read the resolution from the .cfg file at runtime
+        resolution = section.getint('resolution')
+
+        # check if the resolution has been changed since the `compass setup`
+        # command was run
+        if self.resolution != resolution:
+            warnings.warn(f'Resolution was set at {self.resolution:2d}km when'
```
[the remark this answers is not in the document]
andrewdnolan: Your right, it probably makes more sense to raise an error. This version of the code raises a warning and then ignores the changes to resolution made after the time `compass setup` was run. But I could definitely see someone getting confused as to why the their resolution change didn't work. The warning will go to the log file, and it would frustrating to have to dig through that to figure out why the resolution change didn't work.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,301 @@
...
+
+    # Shift x-axis so that the x-origin is all the way to the left
+    xShift = -1.0 * ds_mesh.xCell.min()
+    # Shift y-axis so that it's centered about the MISMIP+ bed trough
+    yShift = 4.e4 - yCenter
+
+    # Shift all spatial points by calculated shift
+    for loc in ['Cell', 'Edge', 'Vertex']:
+        ds_mesh[f'x{loc}'] = ds_mesh[f'x{loc}'] + xShift
+        ds_mesh[f'y{loc}'] = ds_mesh[f'y{loc}'] + yShift
```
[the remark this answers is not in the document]
andrewdnolan: Thanks @matthewhoffman!

compass/landice/tests/mismipplus/setup_mesh.py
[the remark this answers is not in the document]
andrewdnolan: Sounds good. Seems like the safest course of action to get these MISMIP+ like experiments running.

compass/landice/tests/mismipplus/setup_mesh.py
andrewdnolan: Yeah I need to update that comment to make it more accurate. I'm using it here to ensure broadcasting and alignment work correctly in variable assignment. When a scalar value is being assigned to a field `.loc[:]` will ensure that the scalar value is broadcast against all the dimensions of the field and that the broadcasting is aligned properly . For example, if we left off the `.loc[:]` and did something like `src.example_field = 910.` then the resulting example field would have all it's dimensions (e.g. `Time`, `nCells`, `nVertInterfaces`) dropped and just be a scalar value. 

But the need for `.loc[:]` in field assignment is more general. For example the `mask` variable in the `_setup_MISMPPlus_IC` function  only contains dimensions `nCells` (since it was created from `src.yCell` which itself only contains dimensions `nCells`). So when we use the `mask` variable to assign fields (e.g. `dirichletVelocityMask`) the inclusion of `.loc[:]` ensure that the value we are assigning are broadcast and aligned against all the dimensions of the field on the right hand side (`Time`, `nCells`, `nVertInterfaces`). 

So more generally, `.loc[:]` is needed anytime any time a variable is being used for assignment which does not contain all the dimensions of the field you're trying to set. Therefore, `.loc[:]` is needed in all the IC prescriptions because all variables in the mesh contain a `Time` dimension, but none of the fields used to set the IC contain that `Time` dimension. 

Hopefully that description is clear. I'll update the comment here to reflect the details as elaborated here more accurately.

compass/landice/tests/mismipplus/setup_mesh.py
andrewdnolan: Thanks for catching this @matthewhoffman! And thanks for the info @mperego. 

I'll update the `albany_input.yaml` file to reflect this.

compass/landice/tests/mismipplus/setup_mesh.py
andrewdnolan: Great, I'll add a config option. How does `vetical_layer_distribution` with the options `uniform` or `glimmer` sound? I'll leave the default as `glimmer` so the `SetupMesh` step doesn't alter the vertical layer distribution, as created by the `create_landice_grid_from_generic_MPAS_grid.py`, by default.

compass/landice/tests/mismipplus/setup_mesh.py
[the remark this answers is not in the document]
andrewdnolan: ### Option 1
Well maybe`OrderDict` isn't the easiest way. While it works, it makes it a little harder to read. Here's what would be needed: 
```python
resolution_params = {'8km': OrderedDict([('nx', 73),  ('ny', 14), ('dc', 9237.604307)]),
                     '4km': OrderedDict([('nx', 144), ('ny', 24), ('dc', 4618.802154)]),
                     '2km': OrderedDict([('nx', 284), ('ny', 44), ('dc', 2309.401077)]),
                     '1km': OrderedDict([('nx', 566), ('ny', 84), ('dc', 1154.700538)])}
```
then 
```python
self.nx, self.ny, self.dc = resolution_params[key].values()
```
should work version independent. 

### Option 2
An alternative and easier method as compared to option 1 is: 
```python
mesh_params_dict = resolution_params[key].values()

for resolution_param in mesh_params_dict: 
    setattr(self, resolution_param, mesh_params_dict[mesh_params_dict])
``` 
but still not super easy to read. 

### Option 3
There's always the brute force option, which while not super concise is very legible:
```
mesh_params_dict = resolution_params[key].values()

self.dx = mesh_params_dict['dx']
self.dy = mesh_params_dict['dy']
self.dc = mesh_params_dict['dc']
```
---
While this isn't super important for the PR, which do you prefer?

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+
+        ds_mesh = make_planar_hex_mesh(nx=nx, ny=ny, dc=dc,
+                                       nonperiodic_x=True,
+                                       nonperiodic_y=True)
+
+        ds_mesh = mark_cull_cells_for_MISMIP(ds_mesh)
+        ds_mesh = cull(ds_mesh, logger=logger)
+        ds_mesh = convert(ds_mesh, logger=logger)
+
+        # custom function to shift the orgin for the MISMIP+ experiments
```
matthewhoffman, line 111:
```suggestion
        # custom function to shift the origin for the MISMIP+ experiments
```

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+            Domain length in y direction [m]
+        """
+
+        # Find amplitude of dual mesh (i.e. `dy`) using the nomial resolution
+        nominal_dy = np.sqrt(3.) / 2. * nominal_resolution
+
+        # find the number of y rows (`ny`) from the `nominal_dy`
+        nominal_ny = Ly / nominal_dy
+
+        # find the acutal number of y rows (`ny`) by rounding the `nominal_ny`
```
matthewhoffman, line 181:
```suggestion
        # find the actual number of y rows (`ny`) by rounding the `nominal_ny`
```

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+        # to the nearest _even_ integer. `make_planar_hex_mesh` requires that
+        # `ny` be even
+        ny = np.ceil(nominal_ny / 2.) * 2
+
+        return int(ny)
+
+    def calculate_dc(Ly, ny):
+        """
+        Calcuate the edge length that conforms to the desired `ny`
```
matthewhoffman, line 190:
```suggestion
        Calculate the cell spacing that conforms to the desired `ny`
```

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+        gutterLength: float
+            Desired gutter length [m] on the eastern domain. A value of 0.0
+            will result in two extra cell for the gutter.
+        """
+
+        if gutterLength == 0.0:
+            nx = np.ceil(Lx / dc)
+            # The modulo condition below ensures there is exactly one cell
+            # past the the desired domain length. So, when no gutter
+            # infromation is provided, add an extra column to make
```
matthewhoffman, line 232:
```suggestion
            # information is provided, add an extra column to make
```

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+            # the default gutter length 2
+            nx += 1
+        else:
+            # ammend the domain length to account for the gutter
```
matthewhoffman, line 236:
```suggestion
            # amend the domain length to account for the gutter
```

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+        if Lx % dc > dc / 2.:
+            nx += 1
+
+        return int(nx)
+
+    ny = calculate_ny(nominal_resolution, Ly)
+    dc = calculate_dc(Ly, ny)
+    nx = calculate_nx(Lx, dc, gutterLength)
+
+    # add two to `ny` accomodate MISMIP+ specific culling requirments.
```
matthewhoffman, line 254:
```suggestion
    # add two to `ny` to accomodate MISMIP+ specific culling requirments.
```

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+    # This ensures, after the culling, that the cell center too cell center
```
matthewhoffman, line 255:
```suggestion
    # This ensures, after the culling, that the cell center to cell center
```

matthewhoffman (changes requested):
@andrewdnolan , I did a quick review now to get you quick feedback.  Overall, this looks fantastic.  All of my comments are nitpicky things.  I appreciate the level of detail that went in to getting a robust product here, and despite what I said initially about this mesh probably not getting used much, we keep finding new reasons we'll need it in the coming months, so I think this was worth doing right.  I'll review again once the docs are added, and I will then actually test it by running it.

compass/landice/tests/mismipplus/spin_up/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+        """
+        name = "spin_up"
+
+        super().__init__(test_group=test_group, name=name)
+
+        config = CompassConfigParser()
+        module = 'compass.landice.tests.mismipplus.spin_up'
+        # add from config
+        config.add_from_package(module, 'spin_up.cfg')
```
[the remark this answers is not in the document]
andrewdnolan: I just wanted to chime in here with a little context as to why this is here. I don't think this section is necessary in most cases, but this was my work around to get the resolution (as passed in the config file) in the directory structure. Adding the config here allows the resolution to be parsed before the step is initialized, and therefore the resolution value can be passed to the `subdir` argument in the `Step` initialization. 

Maybe there's a better way to do this, but this was just what I came up with from looking through the compass repo (with a particular focus on the mesh convergence tests which has resolution in the directory structure).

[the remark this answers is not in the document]
matthewhoffman: @andrewdnolan , thanks for the explanation - that's really helpful.  So what would happen if the user is specifying a cfg file in `compass setup` that has a different value from the default here?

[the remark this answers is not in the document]
andrewdnolan: Then an error would be raised. At the time of `compass setup` the `resolution` value is stored as an attribute of the step. Then at compass run, the `resolution` is read from the config again and if the value from the config differs from the stored value an error is raised: 

 https://github.com/andrewdnolan/compass/blob/d0b93c4f0b5fe7a7e701aac0964989bb8d3a6343/compass/landice/tests/mismipplus/setup_mesh.py#L85-L97

The error should only pertain to `resolution` since it's the only one stored as an attribute. Other parameters (e.g. gutter, density) should have no problem if changed at runtime.

compass/landice/tests/mismipplus/streams.landice
```diff
@@ -48,10 +44,6 @@
     <var name="xtime"/>
     <var name="daysSinceStart"/>
     <var name="deltat"/>
-    <var name="allowableDtACFL"/>
-    <var name="allowableDtDCFL"/>
-    <var name="calvingCFLdt"/>
-    <var name="dtCalvingCFLratio"/>
```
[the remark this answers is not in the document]
andrewdnolan: Okay! Works for me. Are they only needed in the `globalStatsOutput` stream? Or would you like them in the `restart` and/or `output` streams?

compass/landice/tests/mismipplus/smoke_test/__init__.py
```diff
@@ -21,11 +25,19 @@ def __init__(self, test_group):
         name = 'smoke_test'
         super().__init__(test_group=test_group, name=name)
 
-        ntasks = 36
-        min_tasks = 4
+        resolution = '2000m'
```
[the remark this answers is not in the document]
andrewdnolan: Fixed! Thanks for catching that.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,473 @@
...
+        if self.resolution != resolution:
+            raise Exception(f'Resolution was set at {self.resolution:2d}km'
+                            f' when `compass setup` was called. Since then,'
+                            f' the resolution in the configuration file has'
+                            f' been changed to {resolution:2d}km. Changing'
+                            f' resolution at runtime is not supported. Change'
+                            f' the resolution value in the configuration file'
+                            f' within the python module and rerun the `compass'
+                            f' setup` command in order to create a mesh at a'
+                            f' resolution of {resolution:2d}km')
```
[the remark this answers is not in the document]
andrewdnolan: Thanks for catching this! Fixed in commit `519189e`.

compass/landice/tests/mismipplus/albany_input.yaml
```diff
@@ -3,6 +3,10 @@
 ANONYMOUS:
   Build Type: Tpetra
 
+  Problem:
```
matthewhoffman, line 6: @mperego , MISMIP+ requires a specified rate factor of "A=6.338 × 10−25 Pa−3 s−1 Glen’s law coefficient = 2.0 × 10−17 Pa−3 a−1", and it looks like that setting has gotten lost at some point.  Can you instruct us the current way to specify that in the yaml file here?

matthewhoffman: @mperego , also do we need this bit?
https://github.com/MPAS-Dev/MPAS-Model/blob/landice/develop/testing_and_setup/compass/landice/MISMIP%2B/albany_input.yaml#L24-L27

mperego: > @mperego , MISMIP+ requires a specified rate factor of "A=6.338 × 10−25 Pa−3 s−1 Glen’s law coefficient = 2.0 × 10−17 Pa−3 a−1", and it looks like that setting has gotten lost at some point. Can you instruct us the current way to specify that in the yaml file here?

You should add this lines in the `Problem` section
  ```
Problem: 
    LandIce Viscosity:
      Type: 'Glen''s Law'
      Flow Rate Type: Uniform
      Glen's Law A: 6.338e-25
```
You can avoid the last line if you set the value in MPAS.

Comment (mperego) [reply compass/landice/tests/mismipplus/albany_input.yaml:6]: > @mperego , also do we need this bit? https://github.com/MPAS-Dev/MPAS-Model/blob/landice/develop/testing_and_setup/compass/landice/MISMIP%2B/albany_input.yaml#L24-L27

You don't really need it.

File: compass/landice/tests/mismipplus/mismipplus.cfg [hunk 31]
```diff
@@ -0,0 +1,13 @@
...
+
+# the number of cells per core to aim for
+goal_cells_per_core = 300
+
+# the approximate maximum number of cells per core (the test will fail if too
+# few cores are available)
+max_cells_per_core = 5000
+
+# number of cells at 1000m resolution (with no gutter present). This is used
+# as huerisitc to scale the number of cells with resolution, in order to
```
matthewhoffman, line 11:
```suggestion
# as heuristic to scale the number of cells with resolution, in order to
```

compass/landice/tests/mismipplus/spin_up/__init__.py
```diff
@@ -0,0 +1,61 @@
+from compass.config import CompassConfigParser
+from compass.landice.tests.mismipplus.run_model import RunModel
+from compass.landice.tests.mismipplus.setup_mesh import SetupMesh
+from compass.testcase import TestCase
+
+
+class SpinUp(TestCase):
+    """
+    Test case for create the MISMIP+ mesh, initial conditions,
```
matthewhoffman, line 9:
```suggestion
    Test case for creating the MISMIP+ mesh, initial conditions,
```

compass/landice/tests/mismipplus/spin_up/__init__.py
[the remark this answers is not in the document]
matthewhoffman: I guess my question is if as written it is even possible to change things at setup time - if a user supplied a modified cfg file with `-f`, whatever value in there is ignored in preference for the version of the cfg file that is actually sitting in the repo, right?

compass/landice/tests/mismipplus/streams.landice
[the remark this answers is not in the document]
matthewhoffman: Just in the globalStats file is good.

compass/landice/tests/mismipplus/tasks.py
```diff
@@ -0,0 +1,96 @@
...
+    ntasks = max(1, 4 * round(cell_count / (4 * goal_cells_per_core)))
+    # if ntasks exceeds the number of cores per node, round to the nearest
+    # multiple of `cores_per_node`.
+    if ntasks > cores_per_node:
+        ntasks = int(cores_per_node) * round(ntasks / cores_per_node)
+
+    return ntasks, min_tasks
+
+
+def _approx_cell_count(config):
```
matthewhoffman, line 63: Is this function called anywhere?  Maybe I'm missing it somehow, but I don't see a call to it.

compass/landice/tests/mismipplus/tasks.py
```diff
@@ -0,0 +1,96 @@
...
+    # approximate the number of cells for the desired resolution
+    cell_count = int(ncells_at_1km_res * (1000 / resolution)**2)
+
+    # Account for extra cells when gutter is requested
+    if gutterLength != 0.:
+        # calculate the area of the domain with the gutter
+        new_area = (Lx + gutterLength) * Ly
+        # scale by the approx cell count by the relative area increase
+        # due to presence of the gutter
+        cell_count *= ref_area / new_area
```
matthewhoffman, line 96: Related to previous question, why is there nothing being returned from this function?

matthewhoffman (changes requested):
Another quick pass with a few changes.  I'll follow up with actually running this in the next round!

compass/landice/tests/mismipplus/tasks.py
andrewdnolan: Oh, thanks for catching that Matt! Your totally right that the function is never called, but it definitely should be. 

The plan was to replace [Lines 34-40](https://github.com/andrewdnolan/compass/blob/bc691ee5a0ed5d1a235953e6db512ed6ea914b22/compass/landice/tests/mismipplus/tasks.py#L34-L40) of that same file with the function call. 

While I flushed out the function, I never actually deleted the old Lines 34-40 and replaced them with the function call. Related to your comment below, I guess that also why the function never returned anything although it definitely should. 

Both of the those things have been fixed and I'll push the related commit in just a second.

compass/landice/tests/mismipplus/spin_up/__init__.py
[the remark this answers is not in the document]
andrewdnolan: Hmm, I guess I'm not 100% clear by what you mean at "setup time."

For parameters in the config file, excluding `resolution`, it is possible to change values between running `compass setup` and `compass run`. I did some crude testing by printing to the log file parameter values changed between these two steps (e.g. `ice_density=200` from reference of `918`) and can confirm the values are updated accordingly. 

For resolution is it not. While that does make setting up various resolutions a little cumbersome, I think this limitation is needed since the directory structure is only created at the `compass setup` stage. If we didn't apply this restriction we could have a 2000m mesh sitting in the directory named 8000m, for example.

compass/landice/tests/mismipplus/spin_up/streams.spin_up
```diff
@@ -0,0 +1,25 @@
...
+<stream name="output"
+        type="output"
+        filename_template="output.$Y.nc"
+        filename_interval="20000-00-00_00:00:00"
+        output_interval="00100-00-00_00:00:00"
+        precision="double"
+        clobber_mode="overwrite">
+</stream>
+
+<stream name="globalStatsOutput"
```
trhille, line 21: It would be helpful to add `<var name="albanyVelocityError"/>` to the globalStats stream, so we can make sure the velocity solver is converging.

docs/developers_guide/landice/test_groups/mismipplus.rst
```diff
@@ -4,39 +4,98 @@ mismipplus
...
+setup_mesh
+~~~~~~~~~~
+The class :py:class:`compass.landice.tests.mismipplus.setup_mesh.SetupMesh` 
+defines a step for creating the mesh/initial conditions for a MISMIP+ spinup 
+simulation. This step is only used by the ``spin_up`` test case. An odd number 
+of rows are culled along the North/South boundaries such that the ragged 
+layout of mesh rows is symetric in the y direction. 
+
+The ``resolution`` parameter from the ``[mesh]`` section of the configuration
+file is the nominal resolution, where the actual resolution is itteratively
```
trhille, line 39:
```suggestion
file is the nominal resolution, where the actual resolution is iteratively 
```

docs/developers_guide/landice/test_groups/mismipplus.rst
```diff
@@ -4,39 +4,98 @@ mismipplus
...
+The function :py:func:`compass.landice.tests.mismipplus.tasks.get_ntasks_from_cell_count()`
+can be used to compute a good number of MPI tasks (both the target and the
+minimum) for MALI to use based on the ``goal_cells_per_core`` and
+``max_cells_per_core`` config options as well as the number of cells in a mesh.
+The idea is that we want to run MALI with about 300 cells per core
+(the default value of ``goal_cells_per_core``) but that we would be okay
+with as many as 5000 cells per core (the default ``max_cells_per_core``).
+
+A complication of using this function is that the number of cells in a mesh
+is not know unitl the :py:class:`compass.landice.tests.mismipplus.setup_mesh.SetupMesh`
```
trhille, line 69:
```suggestion
is not known until the :py:class:`compass.landice.tests.mismipplus.setup_mesh.SetupMesh`
```

docs/developers_guide/landice/test_groups/mismipplus.rst
```diff
@@ -4,39 +4,98 @@ mismipplus
...
+step has been run (i.e. at ``compass run``) for a given ``resolution``, but we 
+need to know how many cores and nodes we will use at setup 
+(i.e. at ``compass setup ...``) since thats when the job script is created.
```
trhille, line 72:
```suggestion
(i.e. at ``compass setup ...``) since that is when the job script is created. 
```

trhille (changes requested):
@andrewdnolan, this is excellent stuff! Thanks for your hard work. I have a handful of requested changes, but nothing major.

compass/landice/tests/mismipplus/spin_up/namelist.spin_up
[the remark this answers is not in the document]
matthewhoffman: I also like 0.8 - good catch

compass/landice/tests/mismipplus/mismipplus.cfg
```diff
@@ -0,0 +1,13 @@
+[mismipplus]
+
+# the number of cells per core to aim for
+goal_cells_per_core = 300
+
+# the approximate maximum number of cells per core (the test will fail if too
+# few cores are available)
+max_cells_per_core = 5000
+
+# number of cells at 1000m resolution (with no gutter present). This is used
```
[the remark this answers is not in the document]
matthewhoffman: Based on how I understand this is being used, I think this could be hardcoded and removed from the cfg file.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,479 @@
...
+                                   ds_mesh.areaCell * 0.5,
+                                   ds_mesh.areaCell)
+
+    # `mesh.dcEdge` is a vector. So, ensure that all values equal before we
+    # arbitrarily select a value from the array to use in the `de` calculation
+    if ds_mesh.dcEdge.all():
+        dc = float(ds_mesh.dcEdge[0])
+
+    # get the distance between edges
+    de = 0.5 * dc * np.sin(np.pi / 3)
```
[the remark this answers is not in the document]
matthewhoffman: yeah, this is using periodic_grid meshes only, so we'll always have perfect hexagons and this formula will hold

matthewhoffman:
@andrewdnolan , when I try to set up the mismip+ smoke test, I get a cfg error:
```
  File "/usr/projects/climate/mhoffman/mambaforge/envs/dev_compass_1.2.0-alpha.9/lib/python3.10/configparser.py", line 794, in get
    raise NoOptionError(option, section)
configparser.NoOptionError: No option 'gutter_length' in section: 'mesh'
```

Looks like we need tweak the cfg hierarchy a bit.

compass/landice/tests/mismipplus/spin_up/__init__.py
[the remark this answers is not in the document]
matthewhoffman: Compass supports a 'user' cfg file with `-f` that can contain any cfg sections they want, e.g. specifying the allocation to submit jobs against plus some custom test case settings.  I tried doing that and got incorrect behavior like this:
```
cd compass/landice/tests/mismipplus/spin_up
cp spin_up.cfg blah.cfg
edit blah.cfg: resolution = 4000.
compass setup -n 95 -w $scratch5/compass-mismip+-spinup-blahcfg -f blah.cfg
ls $scratch5/compass-mismip+-spinup-blahcfg/landice/mismipplus/spin_up
```
and I see:
```
8000m  job_script.sh  load_compass_env.sh  spin_up.cfg  test_case.pickle
```
In other words, hardcoding retrieving the cfg from the repo eliminates the ability to use a user-specified cfg file with `-f`.  (I realize now I might not have made the question clear.)  I think you just want to change this bit of code to use self.config, which will already be populated with the merged cfg from the defaults in the repo plus the user one.

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,479 @@
...
+        # ensure that the requested `gutterLength` is valid. Otherwise set
+        # the value to zero, such that the default `gutterLength` of two
+        # gridcells is used.
+        if (gutterLength < 2. * resolution) and (gutterLength != 0.):
+            gutterLength = 0.
+
+        # check if the resolution has been changed since the `compass setup`
+        # command was run
+        if self.resolution != resolution:
+            raise Exception(f'Resolution was set at {self.resolution:2d}m'
```
matthewhoffman, line 94: I got this exception to be raised with the `-f` config testing I mentioned above.  In so doing, I got this value error for this line:
```
  File "/usr/projects/climate/mhoffman/compass/mismip+_spinup/compass/landice/tests/mismipplus/setup_mesh.py", line 94, in run
    raise Exception(f'Resolution was set at {self.resolution:2d}m'
ValueError: Unknown format code 'd' for object of type 'float'
```

compass/landice/tests/mismipplus/spin_up/spin_up.cfg
```diff
@@ -0,0 +1,29 @@
...
+
+# ice density (kg m^{-3}). MISMIP+ uses 918 (Table 1. Asay-Davis et al. 2016),
+# but MALI defaults to 910. The user can choose if they want to stricly follow
+# MISMIP+ or use the default MALI value.
+ice_density = 918.
+
+# Number of vertical levels
+levels = 10
+
+# Initial ice thickness (m)
```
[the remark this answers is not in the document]
andrewdnolan: Added a bit more info to the comment in commit `7ef36a1`. Tried to explain where/why this is needed.

compass/landice/tests/mismipplus/spin_up/namelist.spin_up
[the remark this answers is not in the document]
andrewdnolan: Thanks for the heads up here! Added in [7ef36a1](https://github.com/MPAS-Dev/compass/pull/737/commits/7ef36a1d01d58112c6d06979deb4d13d8346b261).

compass/landice/tests/mismipplus/mismipplus.cfg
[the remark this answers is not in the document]
andrewdnolan: Your totally right. We actually really don't want the user playing with the number, since it would throw off the `ntasks` calculations. 

I've removed the option and comment, and moved both to the [`tasks.py` file](https://github.com/andrewdnolan/compass/blob/4b0f35a01e4606b80d1b86e37910645cfd780e92/compass/landice/tests/mismipplus/tasks.py#L6-L9).

compass/landice/tests/mismipplus/spin_up/streams.spin_up
andrewdnolan: Thanks for the suggestion here! I added this in [`7ef36a1`](https://github.com/MPAS-Dev/compass/pull/737/commits/7ef36a1d01d58112c6d06979deb4d13d8346b261).

compass/landice/tests/mismipplus/spin_up/__init__.py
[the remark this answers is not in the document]
andrewdnolan: Thanks for the clarification about the `-f` option to `compass setup`, I wasn't really familiar with it. 

I've revamp the way the directory structure is created, mainly by overloading the `compass.TestCase`'s `configure` method. Doing things this way ensures configuration passed over the command line (via `-f`) are properly resolved and accounted for. 

Would be great for you to experiment and corroborate that the fix works on your end. This overhaul was added in [4b0f35a](https://github.com/MPAS-Dev/compass/pull/737/commits/4b0f35a01e4606b80d1b86e37910645cfd780e92).

compass/landice/tests/mismipplus/streams.landice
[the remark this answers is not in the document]
andrewdnolan: Added!

compass/landice/tests/mismipplus/setup_mesh.py
andrewdnolan: Thanks for testing this! I wasn't really aware of the `-f` option. 

I've revamped the way the directory structure is created (see response to your original comment). With that revamp, this bug should be fixed and configuration options passed from the command line should be properly parsed.  (Also the error you got in the `SmokeTest` should be fixed). 

All that was pushed in [4b0f35a](https://github.com/MPAS-Dev/compass/pull/737/commits/4b0f35a01e4606b80d1b86e37910645cfd780e92).

compass/landice/tests/mismipplus/setup_mesh.py
```diff
@@ -0,0 +1,479 @@
...
+
+        Parameters
+        ----------
+        Lx : float
+            Length [m] of x domain
+
+        dc : float
+            Edge length [m]
+
+        gutterLength: float
```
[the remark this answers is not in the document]
andrewdnolan: I've gone to a consistent snake-case naming scheme as of [7025f22](https://github.com/MPAS-Dev/compass/pull/737/commits/7025f22a8b8540ecded117ac6e9295816dceb2dd). Thanks for the heads up about this!

compass/landice/tests/mismipplus/setup_mesh.py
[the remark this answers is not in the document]
andrewdnolan: I've added a comment which explains/justifies the equilateral assumption, per you request.

compass/landice/tests/mismipplus/spin_up/spin_up.cfg
```diff
@@ -0,0 +1,29 @@
+# config options for the mesh (i.e. initial condition) setup
+[mesh]
+
+# Nominal cell spacing (m). Resolution is "nominal" since the true cell spacing
+# will be determined such that the cell center to cell center length of the
+# entire domain in the y direction is exactly the required y domain length (80 km).
+resolution = 8000.
+
+# length (m) to extend the eastern domain boundary by. Needed for simulations
```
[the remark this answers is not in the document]
andrewdnolan: In the MISMIP+ protocol, the ice front and eastern boundary are both located at 640 km. (See $L_\rm{x}$ and $x_{\rm calve}$ in Table 1. of[ Asay-Davis et al. [2016]](https://gmd.copernicus.org/articles/9/2471/2016/gmd-9-2471-2016.pdf)). 

As for the default, it's out of view here, but it's included in the comment a little further down. The default gutter length is two gridcells downstream of the ice front. Any `gutter_length` less than 2*resolution is ignored in favor of the default value.

compass/landice/tests/mismipplus/run_model.py
```diff
@@ -56,17 +72,20 @@ def __init__(self, test_case, name='run_model', subdir=None,
         if suffixes is None:
             suffixes = ['landice']
         self.suffixes = suffixes
+
+        if resolution is None:
+            resolution = 0.0
```
trhille, line 77: How does this work?

compass/landice/tests/mismipplus/spin_up/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    def __init__(self, test_group):
+        """
+        Create the test case
+
+        Parameters
+        ----------
+        test_group : compass.landice.test.mismipplus
+            The test group that this test case belongs to
+
```
trhille, line 23:
```suggestion
```

compass/landice/tests/mismipplus/spin_up/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+        """
```
trhille, line 24:
```suggestion
        """

```

compass/landice/tests/mismipplus/spin_up/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+        step.add_streams_file(package, 'streams.spin_up')
+        step.add_namelist_file(package, 'namelist.spin_up')
+
+        self.add_step(step)
+
+    def configure(self):
+        """
+        Set up the directory structure, based on the requested resolution.
+        """
+        # get the config options from the TestCase, which
```
trhille, line 53:
```suggestion

         # get the config options from the TestCase, which
```

compass/landice/tests/mismipplus/spin_up/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+                                     step.subdir)
+
+            # NOTE: we do not set the `step.work_dir` attribute, since it
+            # will be set by `compass setup`` by joining the work dir
+            # provided through the command line interface and the
+            # `step.path` set above.
+
+            # store the resolution (at the time of `compass setup`) as an
+            # attribute. This is needed to prevent the changing of resolution
+            # between `compass setup` and `compas run`, which could result
```
trhille, line 81:
```suggestion
            # between `compass setup` and `compass run`, which could result
```

compass/landice/tests/mismipplus/tasks.py
```diff
@@ -0,0 +1,114 @@
+import xarray as xr
+
```
trhille, line 2:
```suggestion


```

compass/landice/tests/mismipplus/tasks.py
```diff
@@ -0,0 +1,114 @@
...
+    Computes `ntasks` and `min_tasks` based on mesh size (i.e. cell_count)
+
+    Parameters
+    ----------
+    config : compass.config.CompassConfigParser
+        Configuration options for the test case the step belongs to
+
+    cell_count : int
+        Number of horizontal cells in the mesh. The value of this parameter is
+        caclulated (or approximated) using the functions below.
```
trhille, line 23:
```suggestion
        calculated (or approximated) using the functions below.
```

compass/landice/tests/mismipplus/tasks.py
```diff
@@ -0,0 +1,114 @@
...
+    Returns
+    -------
+    ntasks : int
+        The target number of MPI tasks
+
+    min_tasks : int
+        The minimum number of tasks
+    """
+
+    # read MPI related parameters from configuration file
```
trhille, line 34: Can we just use `config.getint()` instead of getting floats and then converting to floats below?

compass/landice/tests/mismipplus/tasks.py
```diff
@@ -0,0 +1,114 @@
...
+    ----------
+    mesh_filename : str
+        A file containing the MPAS mesh (specifically `nCells`)
+
+    Returns
+    -------
+    cell_count : int
+        the number of cells in the mesh
+    """
+    # get cell count from mesh
```
trhille, line 65:
```suggestion

     # get cell count from mesh
```

compass/landice/tests/mismipplus/tasks.py
```diff
@@ -0,0 +1,114 @@
...
+
+    gutter_length: float
+        Desired gutter length [m] on the eastern domain.
+
+    Returns
+    -------
+    cell_count : int
+        the number of cells in the mesh
+    """
+    # Fixed domain lenghts [m] (without gutter)
```
trhille, line 91:
```suggestion

    # Fixed domain lenghts [m] (without gutter)
```

docs/users_guide/landice/test_groups/mismipplus.rst
```diff
@@ -3,26 +3,93 @@
...
-`MISMIP+ description paper <https://tc.copernicus.org/articles/14/2283/2020/>`_
-for more details.  Note that the 2km mesh is not the most accurate resolution
-used by MALI, but provides a convenient trade off between computational cost
-and accuracy for running tests.
-
-The test group only includes a single test - a smoke test.
-More tests may be added later.
-There are no config options.
+The ``mismipplus`` test group (:py:class:`compass/landice/tests/mismipplus`)
+implements a limited number of the expeirments from the Marine Ice Sheet
```
trhille, line 7:
```suggestion
implements a limited number of the experiments from the Marine Ice Sheet 
```

docs/users_guide/landice/test_groups/mismipplus.rst
```diff
@@ -3,26 +3,93 @@
...
-and accuracy for running tests.
-
-The test group only includes a single test - a smoke test.
-More tests may be added later.
-There are no config options.
+The ``mismipplus`` test group (:py:class:`compass/landice/tests/mismipplus`)
+implements a limited number of the expeirments from the Marine Ice Sheet 
+Model Intercomparison Project Plus (MISMIP+; 
+`Asay-Davis et al., 2016 <https://gmd.copernicus.org/articles/9/2471/2016/gmd-9-2471-2016.pdf>`_).
+To date, the spinup and ``Ice0`` are the only expeirments currently supported.
```
trhille, line 10:
```suggestion
To date, the spinup and ``Ice0`` are the only experiments currently supported. 
```

trhille (changes requested):
@andrewdnolan, I just have a handful of minor comments. I tested this by creating and running 1-, 2-, 4-, 8-, and 16-km meshes, and everything seems to be functioning as expected!

compass/landice/tests/mismipplus/run_model.py
andrewdnolan: Yeah this is definitely strange but has to do with the very different way `resolution` is set between `SmokeTest` and `SpinUp` test cases. 

## `SmokeTest` 
So, this last round of commits removes the option to specify the `SmokeTest` resolution in a config file. The problem with the former approach is a user could request a resolution not supported by test case (currently, any value other than `2000m`). With the current code, the supported resolutions are added as individual test cases at the test group level:
https://github.com/andrewdnolan/compass/blob/67b26525f73724c1d6ba0b2f5d7c724427a2c9e0/compass/landice/tests/mismipplus/__init__.py#L19-L21
(This also has the added benefit of the supported resolutions being shown in `compass list`). __So__, for the `SmokeTest` a resolution is always passed to the initialization and in the code above `resolution` will never be `None`. 

## `SpinUp` 
Okay, so now we get to why a "dummy" value of `0.0` is specified. For the `SpinUp` we want to support arbitrary resolution, which should be set in the users config file. The big hiccup here is that we want to use the requested resolution in the directory structure, but the config file which isn't fully parsed until after the step is initialized. Ideally we'd be able to pass `resolution` into the initialization of the `SpinUp` testcase, but that doesn't work. (We can parse the resolution value in the compass repo's config before initialization, but can not parse a user specified config file (e.g. `compass setup -f ....`) prior to the testcases initialization). So, instead we have this dummy value of `0.0` and instead set the value of the `resolution` attribute in the `SpinUp.configure` method so that we have access to all the config options: 
https://github.com/andrewdnolan/compass/blob/67b26525f73724c1d6ba0b2f5d7c724427a2c9e0/compass/landice/tests/mismipplus/spin_up/__init__.py#L49-L83

So, in short the `SmokeTest` knows the value of the `resolution` attribute at the time of initialization, whereas the `SpinUp` needs to parse `resolution` from the config options, so a temporary dummy value is used until the `resolution` attribute is set by the `SpinUp.configure` method. 

Now, you might ask: Why set the dummy value of `0.0` at all? Why not just leave the `resolution` attribute unset for the `SpinUp` testcase until the configure method is called? That's definitely an option, but in talking to Xylar that seems like poor python practice. I guess it's best to initialize all attributes within the `__init__` method. 

Happy to ditch this if need be in order to have a bit clearer but less "proper" python code.

compass/landice/tests/mismipplus/spin_up/namelist.spin_up
```diff
@@ -0,0 +1,5 @@
+config_run_duration = '20000-00-00_00:00:00'
```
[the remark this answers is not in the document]
andrewdnolan: Gotcha, I've switched to `config_stop_time` for both the `SmokeTest` and the `SpinUp`.

compass/landice/tests/mismipplus/run_model.py
trhille: Okay, this is fine. I just wanted to make sure it was actually doing something intentional.

andrewdnolan: I added a comment in [f916f60](https://github.com/MPAS-Dev/compass/pull/737/commits/f916f60ccc9d7b8e515ef6a8a775fd71979c5b40) which summarizes the message above to provide a brief explanation to new/future users.

compass/landice/tests/mismipplus/tasks.py
andrewdnolan: I've switched to parsing all the parameters as `int`s and then casting them as floats where need be. This was done as of [f916f60](https://github.com/MPAS-Dev/compass/pull/737/commits/f916f60ccc9d7b8e515ef6a8a775fd71979c5b40). Thanks for catching this!

andrewdnolan:
# Testing 
All testing done on `chicoma`
 
## `SmokeTest`

__Setup command__: 
`compass setup -n 95 -w /lustre/scratch5/anolan/mismip+_pr_testing`

__Then__:  
```
cd /lustre/scratch5/anolan/mismip+_pr_testing/landice/mismipplus/smoke_test/2000m
sbatch job_script.sh
```
__Output__: 
Screenshot 2024-02-21 at 12 43 16 PM


## `SpinUp`
Added some account/job related config, plus all resolution / parameter variations (e.g. `ice_density` and `gutter_length`) are done through a config file passed to `compass setup` with `-f`. 
```
[parallel]
account = t24_aishydro

[job]
qos = debug

wall_time = 0:10:00

# config options for the mesh (i.e. initial condition) setup
[mesh]

# Nominal cell spacing (m). Resolution is "nominal" since the true cell spacing
# will be determined such that the cell center to cell center length of the
# entire domain in the y direction is exactly the required y domain length (80 km).
resolution = 8000.

# length (m) to extend the eastern domain boundary by. Needed for simulations
# that use a dynamic calving law, where the calving front will be irregularly
# shaped. Any value less than `2*resolution` will be ignored, as the default
# gutter length is 2 gridcells.
gutter_length = 0.

# ice density (kg m^{-3}). MISMIP+ uses 918 (Table 1. Asay-Davis et al. 2016),
# but MALI defaults to 910. The user can choose if they want to stricly follow
# MISMIP+ or use the default MALI value.
ice_density = 918.
```
All variations of the test are setup with: 
```
compass setup -n 94 -w /lustre/scratch5/anolan/mismip+_pr_testing -f spin_up.cfg
```
And run by executing: 
```
cd /lustre/scratch5/anolan/mismip+_pr_testing/landice/mismipplus/spin_up
sbatch job_script.sh
```
Then once the test case is done, run: 
```
 mv *.log {resolution}m
 mv compass.o* {resolution}m
```
where `{resolution}` is the resolution of the testcase as specified in the config file (at the time of compass setup). 
The above command is need to prevent the log files being overwritten by successive tests at different resolutions. 
 
### (a): 1 km 
__Custom config options__: 
```
[mesh]
resolution = 1000.
```
__Output__: 
Screenshot 2024-02-21 at 12 58 24 PM

### (b):  4 km
__Custom config options__: 
```
[mesh]
resolution = 4000.
gutter_length = 12000.
ice_density = 218.
```
(_Note_: a density of 218. is totally arbitrary and just used to make sure the density in the namelist reflects the value requested in the config file). 

__Output__: 
Screenshot 2024-02-21 at 1 00 00 PM
From`/lustre/scratch5/anolan/mismip+_pr_testing/landice/mismipplus/spin_up` running `less 4000m/run_model/namelist.landice` shows: 
Screenshot 2024-02-21 at 1 00 43 PM
where density in the namelist properly reflects what's in the config file. 

### (c) 8000 m 
__Custom config options__: 
```
[mesh]
resolution = 1000.
ice_density = 910.
```
__Output__: 
Screenshot 2024-02-21 at 1 04 14 PM
From`/lustre/scratch5/anolan/mismip+_pr_testing/landice/mismipplus/spin_up` running `less 8000m/run_model/namelist.landice` shows: 
Screenshot 2024-02-21 at 1 05 09 PM
where density in the namelist properly reflects what's in the config file.

### (d) 16000 m 
 __Custom config options__: 
```
[mesh]
resolution = 16000.
```
__Output__: 
Screenshot 2024-02-21 at 1 06 31 PM

### Summary 
All variations of the `SpinUp` testcase generate the mesh/initial condition successfully and successfully execute the landice model until the wall clock time (10 minutes) is reached.

compass/landice/tests/mismipplus/spin_up/namelist.spin_up
```diff
@@ -0,0 +1,4 @@
+config_stop_time = '20000-01-01_00:00:00'
```
trhille, line 1:
```suggestion
config_stop_time = '20001-01-01_00:00:00'
```

compass/landice/tests/mismipplus/tasks.py
trhille: Do we need to cast these as floats at all? It does no harm, but it will give exactly the same answer as using ints.

trhille (approved):
@andrewdnolan, amazing work! I have two tiny comments, but this is ready to go.

matthewhoffman (approved):
@andrewdnolan , thanks for all the revisions on this PR!  I'm happy to merge it.  Before I do that, can you update the PR description describing the final product?  You're also welcome to make the small changes Trevor suggested if you like.  Once you're happy with things, let me know and I will merge.

compass/landice/tests/mismipplus/tasks.py
andrewdnolan: Your right, in python casting as floats is kind of overkill. Especially since all of the floating point operations with `int`s we are doing are division, where we are explicitly using the floating point operator (e.g. `/` vs. `//`). 

I was just trying to follow best practice, but seems a little unnecessary here and at the expense of readable code. I'll remove the casting as floats to make things a little more concise.

andrewdnolan:
@trhille @matthewhoffman Thanks for all the help reviewing / work shopping this PR! 

I've updated the PR description to reflect the final product. I've also implemented those two requested changes from Trevor. 

So, I'm happy with where things are if you guys are.

compass/landice/tests/mismipplus/tasks.py
trhille: Thanks. I wasn't sure what was considered best practice here, but I think it is more readable without casting them as floats.

PR #743, opened 2023-12-03
Use only compass release version is spack env name

xylar:
This means that we update the release version for new spack envs and we update the alpha, beta, etc. for dependency changes.

The docs for spack deployment have been updated accordingly.

Checklist


xylar:
This fixes an issue introduced by #742 where `-alpha_8` spack environments were not created but load scripts generated after that merge are looking for them.

xylar:
## Testing

With this change, I am able to run the `pr` suite on Chrysalis with Intel and OpenMPI, whereas I wasn't able to successfully source a load script without these changes.

PR #745, opened 2023-12-03
Update namelist options for recent changes to Redi and Adams-Bashforth 2 

xylar:
This merge sets `config_Redi_min_layers_diag_terms = 0`, consistent with https://github.com/E3SM-Project/E3SM/pull/6077.

It also changes the time integrator from `split_explicit` to `split_explicit_ab2` nearly everywhere.

Checklist


xylar:
@mark-petersen and @cbegeman, I will test this to make sure all affected tests still work.

Once that's done, I'd appreciate your feedback to make sure we don't want to keep `split_explicit` (as opposed to `split_explicit_ab2`) for any tests.

xylar:
## Testing

I ran the `pr` suite, the `gotm` test, the two `internal_waves` default tests, the `isomip_plus` z-star Ocean0 test, and the 4 `sphere_transport` tests on Chrysalis with Intel and OpenMPI.  I'll update when `sphere_transport` finishes but all other tests ran successfully.  I haven't looked at the output carefully or compared with baselines because the changes are expected to be larger than machine precision ("climate changing").

cbegeman:
@mark-petersen Could you weigh in on whether this internal_wave test result is satisfactory? I wouldn't have expected there to be negative values.

AB2
image

split_explicit
image

mark-petersen:
@xylar thanks for changing these tests. I agree that we should change to AB2 in all compass tests, since we made the switch in E3SM. 

@cbegeman thanks for noticing the negative relative RPE in the internal wave test case. Note that the vertical scale is a factor of 10 smaller, which means less spurious mixing over time. The negative values are concerning. It means that "anti-mixing" is going on, which would be gradient steepening or grid-scale noise. We should view some horizontal slices of temperature at those negative RPE points and compare the two time step schemes, to see what is going on. 

It might be that the timestep needs to be shortened a bit for AB2 relative to split explicit, because there are fewer iterations within each timestep, and we are seeing the result of a numerical instability.

xylar:
@mark-petersen, I think we're still waiting for your response here.

mark-petersen:
Sorry I missed this. I'm out for a week, so I think it would be better to remove the flag change from the `internal_wave` test case.

mark-petersen (approved):
Please merge after removing changes from `internal_wave` tests. The AB2 was tested at length for global simulations.

xylar:
@cbegeman, I switched the internal waves test back to `split_explicit`.  Would you be in a good position to test it again?  If things work out, I think we can hopefully get this branch merged.

xylar:
@cbegeman, sure that's fine.

xylar:
Thanks for the earlier testing and review!

PR #749, opened 2023-12-21
Add an option to setup coupled MALI-SLM runs in the ismip6 run testgroup

hollyhan:
This PR adds to the `/landice/ismip6_run/ismip_ais_proj2300` testcase the option to setup coupled MALI-Sea Level Model simulations using the ismip6-Antarctica-2300 experimental protocol. 

With these changes, inclusion of the SLM can be activated through a test case cfg option.  This will set up required namelist and input settings, as well as the namelist.sealevel file used by SLM.  It will also set up an additional test case step that generates the required mapping files between MALI and SLM.  This must be run before any of the experiments are run.

This PR requires addition of Pyproj string `projections['ais-bedmap2-sphere'] = '+proj=stere +lat_ts=-71.0 +lat_0=-90 +lon_0=0.0 +k_0=1.0 +x_0=0.0 +y_0=0.0 +ellps=sphere'` in the MPAS-Tools script `set_lat_lon_fields_in_planar_grid.py` to set the latitude and longitude values as if MALI was projected on sphere rather than ellipse. The requirement has been merged here (https://github.com/MPAS-Dev/MPAS-Tools/pull/559/files).

### DIFF
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300.cfg
@@ -44,3 +44,19 @@ von_mises_parameter_path = UNKNOWN
 
 # Whether facemelting should be included in the runs
 use_face_melting = False
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
+
+# NOTE: for the directories related to the sea-level model input/outputs, the slash '/' at the end of the directory name is necessary.
+# Path to the directory containing globally defined ice thickness field for the sea-level model
+slm_input_ice = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/icemodel/
+
+# Path to the directory containing earth model for the sea-level model
+slm_input_earth = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/earthmodel/
+
+# Path to the diectory containing other input files (present-day global topography and sea-level model grid files)
+slm_input_others = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/others/
+
+# number of gauss-legendre nodes in latitude (typically an integer multiple of 512)
+nglv = 2048
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300.cfg
@@ -46,7 +46,7 @@ von_mises_parameter_path = UNKNOWN
 use_face_melting = False
 
 # True if running coupled MALI-sea level model simulation
-sea_level_model = True
+sea_level_model = False
 
 # NOTE: for the directories related to the sea-level model input/outputs, the slash '/' at the end of the directory name is necessary.
 # Path to the directory containing globally defined ice thickness field for the sea-level model
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300.cfg
@@ -55,6 +55,15 @@ slm_input_ice = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/ice
 # Path to the directory containing earth model for the sea-level model
 slm_input_earth = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/earthmodel/
 
+# Earth structure profile
+# possible values: any file name string of earth model file.
+# Note that there is no single representative earth model for the globe or
+# any region of the globe (e.g., West Antarctica, East Antarctica, Greenland)
+# But for this test, we use 'prem_512.l60K2C.sum18p6.dum19p2.tz19p4.lm22' for West Antarctica
+# and 'prem_coll_512.l120C.ump5.lm10' for East Antarctica. Other earthmodels that can be used
+# should exist in the path defined in the config option 'slm_input_earth'.
+slm_earth_structure = prem_512.l60K2C.sum18p6.dum19p2.tz19p4.lm22
+
 # Path to the diectory containing other input files (present-day global topography and sea-level model grid files)
 slm_input_others = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/others/
 
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_chicoma.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_chicoma.cfg
@@ -54,6 +54,15 @@ slm_input_ice = /usr/projects/climate/hollyhan/SeaLevelModel_Inputs/icemodel/
 # Path to the directory containing earth model for the sea-level model
 slm_input_earth = /usr/projects/climate/hollyhan/SeaLevelModel_Inputs/earthmodel/
 
+# Earth structure profile
+# possible values: any file name string of earth model file.
+# Note that there is no single representative earth model for the globe or
+# any region of the globe (e.g., West Antarctica, East Antarctica, Greenland)
+# But for this test, we use 'prem_512.l60K2C.sum18p6.dum19p2.tz19p4.lm22' for West Antarctica
+# and 'prem_coll_512.l120C.ump5.lm10' for East Antarctica. Other earthmodels that can be used
+# should exist in the path defined in the config option 'slm_input_earth'.
+slm_earth_structure = prem_512.l60K2C.sum18p6.dum19p2.tz19p4.lm22
+
 # Path to the diectory containing other input files (present-day global topography and sea-level model grid files)
 slm_input_others = /usr/projects/climate/hollyhan/SeaLevelModel_Inputs/others/
 
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_chicoma.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_chicoma.cfg
@@ -3,7 +3,7 @@
 # Can be "tier1", "tier2", "all", or a comma-delimited list of runs
 # "tier1" is expAE01-expAE06 plus hist and ctrlAE
 # "tier2" is expAE07-expAE14
-exp_list = hist,ctrlAE,expAE05
+exp_list = hist,ctrlAE,expAE01,expAE02,expAE05
 
 # Resolution that will be set up, in km.  Should be one of 4, 8, as those are
 # the two resolutions currently supported.
@@ -43,3 +43,19 @@ von_mises_parameter_path = UNKNOWN
 
 # Whether facemelting should be included in the runs
 use_face_melting = True
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
+
+# NOTE: for the directories related to the sea-level model input/outputs, the slash '/' at the end of the directory name is necessary.
+# Path to the base directory containing globally defined ice thickness field for the sea-level model
+slm_input_ice = /usr/projects/climate/hollyhan/SeaLevelModel_Inputs/icemodel/
+
+# Path to the directory containing earth model for the sea-level model
+slm_input_earth = /usr/projects/climate/hollyhan/SeaLevelModel_Inputs/earthmodel/
+
+# Path to the diectory containing other input files (present-day global topography and sea-level model grid files)
+slm_input_others = /usr/projects/climate/hollyhan/SeaLevelModel_Inputs/others/
+
+# number of gauss-legendre nodes in latitude (typically an integer multiple of 512)
+nglv = 2048
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_chicoma.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_chicoma.cfg
@@ -45,7 +45,7 @@ von_mises_parameter_path = UNKNOWN
 use_face_melting = True
 
 # True if running coupled MALI-sea level model simulation
-sea_level_model = True
+sea_level_model = False
 
 # NOTE: for the directories related to the sea-level model input/outputs, the slash '/' at the end of the directory name is necessary.
 # Path to the base directory containing globally defined ice thickness field for the sea-level model
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_nersc.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_nersc.cfg
@@ -44,3 +44,19 @@ von_mises_parameter_path = UNKNOWN
 
 # Whether facemelting should be included in the runs
 use_face_melting = False
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
+
+# NOTE: for the directories related to the sea-level model input/outputs, the slash '/' at the end of the directory name is necessary.
+# Path to the base directory containing globally defined ice thickness field for the sea-level model
+slm_input_ice = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/icemodel/
+
+# Path to the directory containing earth model for the sea-level model
+slm_input_earth = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/earthmodel/
+
+# Path to the diectory containing other input files (present-day global topography and sea-level model grid files)
+slm_input_others = /global/cfs/cdirs/fanssie/MALI_projects/SeaLevelModel_Inputs/others/
+
+# number of gauss-legendre nodes in latitude (typically an integer multiple of 512)
+nglv = 2048
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_nersc.cfg
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_nersc.cfg
@@ -46,7 +46,7 @@ von_mises_parameter_path = UNKNOWN
 use_face_melting = False
 
 # True if running coupled MALI-sea level model simulation
-sea_level_model = True
+sea_level_model = False
 
 # NOTE: for the directories related to the sea-level model input/outputs, the slash '/' at the end of the directory name is necessary.
 # Path to the base directory containing globally defined ice thickness field for the sea-level model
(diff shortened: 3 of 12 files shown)


hollyhan:
Hi @xylar and @matthewhoffman, the PR is ready to be reviewed. 

The testing directory is in Chicoma (/lustre/scratch4/turquoise/hollyhan/folder_test_compass_ismip6run/landice/ismip6_run/ismip6_ais_proj2300/expAE01_04) where any `climate` group member can access.

Testing (done in Chicoma, with the compass conda environment `1.2.0-alpha.8` version):
In addition to editing and adding the `namelist.landice` and `namelist.sealevel` to the run folder setup by compass when sea-level model is coupled to MALI during `compass setup`, `compass run` in the work directory creates the mapping files `mapfile_mali_to_slm.nc` and `mapfile_slm_to_mali.nc`.  Below is the logger message:
```
Loading conda environment
Done.

Loading Spack environment...

Lmod is automatically replacing "cce/16.0.0" with "gcc/12.2.0".

Done.

compass calling: compass.run.serial._run_test()
  in /lustre/scratch4/turquoise/.mdt4/hollyhan/compass/compass/run/serial.py


compass calling: compass.landice.tests.ismip6_run.ismip6_ais_proj2300.set_up_experiment.SetUpExperiment.runtime_setup()
  inherited from: compass.step.Step.runtime_setup()
  in /lustre/scratch4/turquoise/.mdt4/hollyhan/compass/compass/step.py


compass calling: compass.landice.tests.ismip6_run.ismip6_ais_proj2300.set_up_experiment.SetUpExperiment.run()
  in /lustre/scratch4/turquoise/.mdt4/hollyhan/compass/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py

creating scripfile for the SLM grid with 2048 Gauss-Legendre points in latitude
Running: ncremap -g slm_nglv2048scripfile.nc -G latlon=2048,4096#lat_typ=gss#lat_drc=n2s
ncremap: WARNING slm_nglv2048scripfile.nc already exists and will be overwritten by newly created grid

Running: set_lat_lon_fields_in_planar_grid.py --file mali_meshfile.nc --proj ais-bedmap2-sphere
== Gathering information.  (Invoke with --help for more details. All arguments are optional)

Using ais-bedmap2-sphere projection, defined as: +proj=stere +lat_ts=-71.0 +lat_0=-90 +lon_0=0.0 +k_0=1.0 +x_0=0.0 +y_0=0.0 +ellps=sphere
Input file xCell min/max values: -2545830.9970442397 2785860.086302431
Input file yCell min/max values: -2189345.7664010245 2281377.989714217
Calculated latCell min/max values (radians): -1.5695348162256555 -1.094701146681904
Calculated lonCell min/max values (radians): 7.204743572800302e-06 6.283164667290495
Lat/lon calculations completed.  File has been written.

creating scrip file for the mali mesh
 -- Landice Masks are disabled

 -- WARNING: sphereRadius<0 so setting sphereRadius = 6371220.0
 -- WARNING: 'on_a_sphere' attribute is 'NO', which means that there may be some disagreement regarding area between the planar (source) and spherical (target) mesh
Input latCell min/max values (radians): -1.5695348162256555, -1.094701146681904
Input lonCell min/max values (radians): 7.204743572800302e-06, 6.283164667290495
Calculated grid_center_lat min/max values (radians): -1.5695348162256555, -1.094701146681904
Calculated grid_center_lon min/max values (radians): 7.204743572800302e-06, 6.283164667290495
Calculated grid_area min/max values (sq radians): 2.6159873831700514e-07, 1.0476949449951053e-05
Creation of SCRIP file is complete.
creating MALI -> SLM grid mapfile with bilinear method
Running: srun ESMF_RegridWeightGen -s mali_scripfile.nc -d slm_nglv2048scripfile.nc -w mapfile_mali_to_slm.nc -m conserve -i -64bit_offset --netcdf4 --src_regional
 Starting weight generation with these inputs:
   Source File: mali_scripfile.nc
   Destination File: slm_nglv2048scripfile.nc
   Weight File: mapfile_mali_to_slm.nc
   Source File is in SCRIP format
   Source Grid is a regional grid
   Source Grid is an unstructured grid
   Use the center coordinates of the source grid to do the regrid
   Destination File is in SCRIP format
   Destination Grid is a global grid
   Destination Grid is a logically rectangular grid
   Use the center coordinates of the destination grid to do the regrid
   Regrid Method: conserve
   Pole option: NONE
   Ignore unmapped destination points
   Output weight file in NetCDF4 file format
   Line Type: greatcircle
   Norm Type: dstarea
   Extrap. Method: none

 Completed weight generation successfully.
  ...
...
SLM -> MALI mesh mapfile with blinear method
Running: srun ESMF_RegridWeightGen -s slm_nglv2048scripfile.nc -d mali_scripfile.nc -w mapfile_slm_to_mali.nc -m bilinear -i -64bit_offset --netcdf4 --dst_regional
 Starting weight generation with these inputs:
   Source File: slm_nglv2048scripfile.nc
   Destination File: mali_scripfile.nc
   Weight File: mapfile_slm_to_mali.nc
   Source File is in SCRIP format
   Source Grid is a global grid
   Source Grid is a logically rectangular grid
   Use the center coordinates of the source grid to do the regrid
   Destination File is in SCRIP format
   Destination Grid is a regional grid
   Destination Grid is an unstructured grid
   Use the center coordinates of the destination grid to do the regrid
   Regrid Method: bilinear
   Pole option: ALL
   Ignore unmapped destination points
   Output weight file in NetCDF4 file format
   Line Type: cartesian
   Norm Type: dstarea
   Extrap. Method: none

 Completed weight generation successfully.
  ...
  ...
Removing the temporary mesh and scripfiles...
Warning: replacing namelist options in namelist.landice
config_pio_num_iotasks = 4
config_pio_stride = 128
Running: gpmetis graph.info 512
.
.
.
```


The contents in the work directory look as follows:

```
hollyhan@ch-fe1:/lustre/scratch4/turquoise/hollyhan/folder_test_compass_ismip6run/landice/ismip6_run/ismip6_ais_proj2300/expAE01_04> ls
AIS_4to20km_r01_20220907.regionMask_ismip6.nc				      
ICELOAD_SLM
namelist.sealevel
AIS_4to20km_r01_20220907_smb_NorESM1-M_RCP26-repeat_2300_minus1_bare_land.nc  
ismip6_ais_proj2300.cfg  
OUTPUT_SLM
AIS_4to20km_r01_20220907_TF_NorESM1-M_RCP26-repeat_2300.nc		     
job_script.sh	       
restart_timestamp
albany_input.yaml							      
landice_model	      
 rst.2015-01-01.nc
basin_and_coeff_gamma0_DeltaT_quadratic_non_local_after_relaxation.nc	      
load_compass_env.sh      
step.pickle
expAE01.o10070767		
mapfile_mali_to_slm.nc   
streams.landice
graph.info								      
mapfile_slm_to_mali.nc
graph.info.part.512
```
Note that in addition to `namelist.sealevel` and the mapping files, the folders needed for the SLM outputs `ICELOAD_SLM` and `OUTPUT_SLM` are created as well.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -219,6 +225,42 @@ def setup(self):  # noqa:C901
...
                                       out_name='namelist.landice')
 
+        if sea_level_model:
+            # get config options
+            slm_input_ice = section.get('slm_input_ice')
+            slm_input_earth = section.get('slm_input_earth')
+            slm_input_others = section.get('slm_input_others')
+            nglv = section.getint('nglv')
+
+            # complate the full paths to the SLM inputs
```
matthewhoffman, line 235:
```suggestion
            # complete the full paths to the SLM inputs
```

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300.cfg
```diff
@@ -44,3 +44,19 @@ von_mises_parameter_path = UNKNOWN
 
 # Whether facemelting should be included in the runs
 use_face_melting = False
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
```
matthewhoffman, line 49: Let's leave the default as False for now to keep the behavior from the submitted ISMIP6 experiments.
```suggestion
sea_level_model = False
```

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_chicoma.cfg
```diff
@@ -43,3 +43,19 @@ von_mises_parameter_path = UNKNOWN
 
 # Whether facemelting should be included in the runs
 use_face_melting = True
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
```
matthewhoffman, line 48:
```suggestion
sea_level_model = False
```

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_4km_nersc.cfg
```diff
@@ -44,3 +44,19 @@ von_mises_parameter_path = UNKNOWN
 
 # Whether facemelting should be included in the runs
 use_face_melting = False
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
```
matthewhoffman, line 49:
```suggestion
sea_level_model = False
```

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/ismip6_ais_proj2300_8km_nersc.cfg
```diff
@@ -44,3 +44,19 @@ von_mises_parameter_path = UNKNOWN
 
 # Whether facemelting should be included in the runs
 use_face_melting = False
+
+# True if running coupled MALI-sea level model simulation
+sea_level_model = True
```
matthewhoffman, line 49:
```suggestion
sea_level_model = False
```

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/namelist.sealevel.template
```diff
@@ -0,0 +1,66 @@
...
+
+/
+&file_format
+    ext =''
+    fType_in = 'text'
+    fType_out = 'both'
+
+/
+&file_name
+    planetmodel   = 'prem_512.l60K2C.sum18p6.dum19p2.tz19p4.lm22'
```
matthewhoffman, line 29: Can we add a cfg option to set up either WAIS or typical Earth structure?

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -268,5 +310,95 @@ def run(self):
...
+                     '-d', slm_scripfile,
+                     '-w', 'mapfile_mali_to_slm.nc',
+                     '-m', 'conserve',
+                     '-i', '-64bit_offset', '--netcdf4',
+                     '--src_regional'])
+
+        check_call(args, logger)
+
+        # create mapping file from SLM grid to MALI mesh
+        logger.info('SLM -> MALI mesh mapfile with blinear method')
```
matthewhoffman, line 388:
```suggestion
        logger.info('SLM -> MALI mesh mapfile with bilinear method')
```

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -268,5 +310,95 @@ def run(self):
...
+        config = self.config
+        section = config['ismip6_run_ais_2300']
+        init_cond_path = section.get('init_cond_path')
+        nglv = section.getint('nglv')
+
+        # create the mapping files
+        # first create scrip files
+        mali_scripfile = 'mali_scripfile.nc'
+        slm_scripfile = f'slm_nglv{nglv}scripfile.nc'
+        mali_meshfile = 'mali_meshfile.nc'
```
matthewhoffman, line 346: Let's change this to have a more descriptive filename, e.g.
```suggestion
        mali_meshfile = 'mali_meshfile_sphereLatLon.nc'
```
I know that this file gets deleted when this function completes, but that way if an error occurs, this file isn't confusing.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -268,5 +310,95 @@ def run(self):
...
+                f'latlon={nglv},{2*int(nglv)}#lat_typ=gss#lat_drc=n2s']
+
+        check_call(args, logger=logger)
+
+        # mali scripfile
+        # first have to adjust lat-lon values as if they are on sphere
+        shutil.copy(init_cond_path, mali_meshfile)
+        args = ['set_lat_lon_fields_in_planar_grid.py',
+                '--file', mali_meshfile,
+                '--proj', 'ais-bedmap2-sphere']
```
matthewhoffman, line 364: If we need to use a projection with a sphere, we should open a PR to MPAS-Tools that adds this option.  I feel like we had considered a solution that would allow the MALI mesh to keep its standard projection but couple to SLM on the sphere, but I'm having trouble remembering the details.  @hollyhan or @xylar , do you remember?

One concern I have here is that this change could complicate comparing MALI runs with and without SLM coupling.  I guess the only difference is the mapping and the MALI input files are unaffected by assuming the sphere, but it would be good to remember the details of this sphere/ellipsoid discrepancy and document that in the user's guide for the test.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
```diff
@@ -268,5 +310,95 @@ def run(self):
         """
         Run this step of the test case
         """
+
+        config = self.config
+        logger = self.logger
+        section = config['ismip6_run_ais_2300']
+        sea_level_model = section.getboolean('sea_level_model')
+        if sea_level_model:
+            self._build_mapping_files(config, logger)
```
matthewhoffman, line 319: I agree it makes sense to build the mapping files in the run method, because it will typically be more expensive than we want to do on a login node.  A problem with putting it here, though, is the mapping files would get regenerated if `compass run` was used on a restart, which is what happens when the generated job script is used.  This is kind of a symptom of using compass outside its intended usage.  One solution could be to have `_build_mapping_files` not do anything if the mapping file already exists (as would be the case on a restart).  That seems a little fragile, but less clunky than having to tell compass run if it's a restart or not.  @xylar, do you have any advice for this situation?

matthewhoffman: One option would be to move `_build_mapping_files` to a separate step.  This would have the advantage of only needing to be run once, instead of right now it will be rerun for every experiment.  The experiment steps could contain a symlink to the mapping files in the build_mapping step.  A user would have to know to execute `compass run` for the build_mapping step before starting any of the runs (this would have to be documented in the users guide).  If a user forgot that step, MALI would die when the nonexistent mapping file tried to be read, which would alert the user of the problem.  I think I prefer this solution, but I'm curious to hear if @hollyhan or @xylar can think of any objections.  (@hollyhan , if we agree to go this route, I'm willing to implement this change so you don't have to deal with that)

matthewhoffman (changes requested):
@hollyhan , this is fantastic to get the SLM coupled configuration automated in compass!  I have a number of requested changes, most of them small or straightforward, but there are a couple items that might require some further discussion to decide on the best solution.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
hollyhan: Thanks for making the note, @matthewhoffman!  I'm quite neutral about making the further changes. In one hand, I have no problem with re-running the build_mapping routine in the case of restart because it's relatively a cheap and short process, and it's likely that one would run the simulations by directly executing `srun ./landice_model` rather than `compass run` as that was the initial intention of this PR (following your previous PR that sets up standalone MALI simulations), which would avoid re-creating the mapping files. But I'm also happy if you want to make changes :) .

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
hollyhan: I'm not remembering the details either, even whether we talked about in detail or not.. @xylar - do you have any suggestions?

xylar: Sorry, I clearly lost track of these comments from January.

I'm fine with @hollyhan making an MPAS-Tools PR for these changes.  I agree that the use of the lat/lon on the sphere should only affect the coupling with the SLM so it shouldn't affect comparison with a MALI run without the SLM.

> but it would be good to remember the details of this sphere/ellipsoid discrepancy and document that in the user's guide for the test.

> I'm not remembering the details either, even whether we talked about in detail or not.

Yes, we talked about the sphere vs. ellipsoid issue extensively during my fall visit to LANL.  We found that we needed  to account for it to get the expected conservative mapping between MALI and the SLM, and I think that's what @matthewhoffman wants documented.  @hollyhan, hopefully you have notes from that time that can be turned into documentation, but we can revisit those discussions if necessary.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
xylar: Again, sorry for losing track of this comment.

I'm more inclined to @matthewhoffman's solution of having a separate step for making the mapping files.  Yes, it's cheap to run in this case and it's a little annoying for a user to have to run an extra step.  But it's more consistent with the philosophy of how Compass is supposed to work.

I also don't think we want users to get in the habit of circumventing `compass run`.  It's okay for a knowledgeable user to do that but it shouldn't be required by default because it requires more manual intervention (or at least it normally does -- I'm not sure I'm up on the details of this test group anymore).

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/namelist.sealevel.template
hollyhan: @matthewhoffman, given that no one earth model can be a representative of a region or the globe, I was hesitant to add a config option with names like 'WAIS' or 'typical'. Instead, I made a config option with which the users define the filename of the earth model they want use.

compass/landice/tests/ismip6_run/ismip6_ais_proj2300/set_up_experiment.py
hollyhan: Okay. I am convinced that we take the route that Matt suggested. @matthewhoffman - would you mind implementing the solution? My bandwidth is super tight for the next few weeks, it would be really helpful if you could take care of this part. :)

matthewhoffman: @hollyhan , yes, I can do that part!

matthewhoffman:
@hollyhan , I've moved the mapping file generation to a new step and confirmed that mapping files are created when it is run.  I still need to update the docs, but feel free to try it out and let me know if you run into any issues.

One thing I realized is that all the experiments are set up as restarts from the hist run, so we will need to handle the SLM restart files somehow (or modify the experiments to not be restarts from hist).  Let me know if you have thoughts about that.

matthewhoffman:
@hollyhan , thinking about it more, the easiest thing would be to add instructions that a user needs to manually copy the SLM output from the `hist` run into the projections runs prior to conducting them.

matthewhoffman (approved):
@hollyhan , I updated the docs with the requirement to manually copy the OUTPUT_SLM directory from the hist run for the projection runs.  With that and the changes you added today, I think all the details are taken care of.  Can you update the PR description and let me know if you have any feedback on the new step I added for creating the mapping files?  Once you sign off, I'm happy to have this merged.

@xylar , let me know if you wanted to take another look or are happy to consider this complete.

xylar:
@matthewhoffman and @hollyhan, I think you should proceed. I don't think I'll have time to take another look this week and I don't want to hold things up.

matthewhoffman:
Thanks, @xylar .  Holly said she'll take a final look today as she is setting up new runs on Chicoma.  Assuming no new issues come up, I'll merge this after her thumb's up.

hollyhan:
@matthewhoffman, thanks for updating the document! I confirmed that everything looks and works great. I think it's ready to be merged!

PR #752, opened 2024-01-03
Fix land-locked cells at calving fronts

### LINKED ISSUE #746: Sea ice trapped in isolated cells adjacent to ice shelves
In the SORRM 1950 control experiment, we're seeing 90,000 m of sea ice in one grid cell adjacent to the Ronne ice-shelf:
image
image

This completely corrupts the time series of sea-ice volume:
[iceVolume.pdf](https://github.com/MPAS-Dev/compass/files/13575154/iceVolume.pdf)

This appears to be because the cell has all land vertices, meaning it is completely isolated from a B-grid perspective.  I think this is because we widen channels where land is present but we don't modify `landIceMask` in the same way.  This needs to be fixed in the next SORRM and other Polar-focused meshes.

xylar:
If there are land-locked cells at calving fronts of ice shelves, they need to be added to the land-ice mask so sea-ice doesn't get trapped there.

This merge also gets the minimum latitude and number of sweeps used in finding land-locked cells from config options.

Checklist

closes #746

### DIFF
--- a/compass/ocean/mesh/cull.py
+++ b/compass/ocean/mesh/cull.py
@@ -166,12 +166,19 @@ class CullMeshStep(Step):
         convert_to_cdf5 = config.getboolean('spherical_mesh',
                                             'convert_culled_mesh_to_cdf5')
 
+        latitude_threshold = config.getfloat('spherical_mesh',
+                                             'latitude_threshold')
+
+        sweep_count = config.getint('spherical_mesh', 'sweep_count')
+
         cull_mesh(with_critical_passages=True, logger=logger,
                   use_progress_bar=use_progress_bar,
                   preserve_floodplain=preserve_floodplain,
                   with_cavities=with_ice_shelf_cavities,
                   process_count=self.cpus_per_task,
-                  convert_to_cdf5=convert_to_cdf5)
+                  convert_to_cdf5=convert_to_cdf5,
+                  latitude_threshold=latitude_threshold,
+                  sweep_count=sweep_count)
 
         if do_inject_bathymetry:
             inject_bathymetry(mesh_file='culled_mesh.nc')
@@ -180,7 +187,8 @@ class CullMeshStep(Step):
 def cull_mesh(with_cavities=False, with_critical_passages=False,
               custom_critical_passages=None, custom_land_blockages=None,
               preserve_floodplain=False, logger=None, use_progress_bar=True,
-              process_count=1, convert_to_cdf5=False):
+              process_count=1, convert_to_cdf5=False, latitude_threshold=43.,
+              sweep_count=20):
     """
     First step of initializing the global ocean:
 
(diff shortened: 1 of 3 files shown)


xylar:
## Testing

I created a version of the SORRM mesh using this branch.  Below are screenshots from ParaView before and after adding land-locked cells to the land-ice mask:

Before:
before_land_locked_cells

After:
after_land_locked_cells

Two cells are filled in this example (and 40 more around Antarctica).

xylar:
@mark-petersen, this would be nice to get merged. Would you be able to take a look at these 2 meshes:
```
/lcrc/group/e3sm/data/inputdata/ocn/mpas-o/RRSwISC6to18E3r4/mpaso.RRSwISC6to18E3r4.20240105.nc
/lcrc/group/e3sm/data/inputdata/ocn/mpas-o/SOwISC12to30E3r2/mpaso.SOwISC12to30E3r2.20240104.nc
```
and tell me if you think the land-locked cells at calving fronts are fixed (as in the plots I made above)?

Also, if you can follow the Compass code changes I made, it would be great to know if they look right to you.

mark-petersen:
Thanks. I can see why this would be a problem for sea ice getting caught in the calving front. It's great that the solution is a simple additional call to `add_land_locked_cells_to_mask` with the ice_coverage mask. It's also good to change the hard-coded numbers to config options. I'm looking at those files now. Sorry for the delay.

mark-petersen:
Very nice! I do not see any land-locked cells (landice-locked cells) in the sample I looked at, from your RRS mesh above:
image
image
image

Thanks, looks good!

mark-petersen:
Does it matter if there is a cell with `landIceMask = 1` but `landIcePressure=0` ? I guess it is land ice cover to the coupling fields and sea ice, but the ocean doesn't get pushed down there, as if it were a very thin layer of land ice. I think that should be fine.

mark-petersen (approved):

xylar:
Thanks so much @mark-petersen!  I lost track of this but I really appreciate your review.

> Does it matter if there is a cell with `landIceMask = 1` but `landIcePressure=0` ? 

No, that's perfectly fine, just a little odd.  It won't have melt fluxes because I'm only updating `landIceMask`, not `landIceFraction`, which will remain zero and which is what is used to scale melt fluxes.  In actual fact, `landIcePressure` can have non-zero values right now even where `landIceMask = 0` because we sometimes smooth the calving front over a small distance.  So there is currently no assumption about the correspondence between `landIceMask = 1` and `landIcePressure > 0`

> I guess it is land ice cover to the coupling fields and sea ice, but the ocean doesn't get pushed down there, as if it were a very thin layer of land ice. I think that should be fine.

Right, with the added factor that the land ice also won't melt there, the cell is just insulated from the atmosphere.  This is clearly much better than having trapped but thermodynamically active sea ice.

PR #753, opened 2024-01-03
New Mesh: SOwISC12to30E3r2

xylar:
Long name: SOwISC12to30kmL64E3SMv3r2

* 30 km elsewhere

```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/sowisc12to30e3r2
```

Checklist


xylar:
## Testing

This mesh has been run through `files_for_e3sm` and the output is in:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/sowisc12to30e3r2
```

I have verified that the land-ice mask includes "land-locked" cells that were formerly a problem for sea-ice, based on the fix in #752

xylar:
This will not be adopted because of issues discussed in https://acme-climate.atlassian.net/wiki/spaces/OO/pages/4059529268/Review+SOwISC12to30E3r2, notably a region of trapped sea ice near Stange Ice Shelf and only 2 vertical levels in some cells.

PR #747, opened 2023-12-09
New Mesh: RRSwISC6to18E3r3

xylar:
Long name: RRSwISC6to18L80E3SMv3r3

This RRS (Rossby-radius scaled) mesh has:
* 6 km resolution near the poles
* 18 km resolution at the equator

Mesh, initial condition, dynamic adjustment and files for E3SM will be on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/rrswisc6to18e3r3
```

This mesh is similar to RRSwISC6to18E3r1 #668 but with #732 and some recent updates to Compass dependencies (notably pyremap and mpas_tools).

Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #754, opened 2024-01-04
New Mesh: RRSwISC6to18E3r4

xylar:
Long name: RRSwISC6to18L80E3SMv3r4

```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/rrswisc6to18e3r4
```

This mesh is similar to RRSwISC6to18E3r3 #747 but with #752 and some fixes to the metadata describing the mesh.

Checklist


xylar:
This mesh will not be used for production runs due to the issues reported in https://github.com/E3SM-Project/E3SM/pull/6143#issuecomment-2014045573

PR #757, opened 2024-01-24
Several fixes to deployment

xylar:
This merge includes fixes to deployment, following installation of Jigsaw and Jigsaw-Python from a submodule.  These include:
* installing cmake into the deployment environment
* removing the `compass.egg-info` directory, which sometimes causes trouble
* including the `/usr/lib64` directory in `LD_LIBRARY_PATH` on Anvil, which seems to be needed to build MPAS-Ocean
* do not remove `jigsaw` and `jigsawpy` when updating an environment

Checklist

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -327,24 +327,24 @@ def build_conda_env(env_type, recreate, mpi, conda_mpi, version,
             print(f'{env_name} already exists')
 
     if env_type == 'dev':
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I was able to create, update and recreate a development environment, deploy a test spack environment, and build MPAS-Ocean with Intel and Intel-MPI on Anvil with these changes, whereas I had trouble at each step along the way without these modifications.

PR #760, opened 2024-01-25
Fix jigsaw build

### LINKED ISSUE #759: Move to Jigsaw-Python submodule seems to have broken MPAS-Ocean builds
It seems like https://github.com/MPAS-Dev/compass/pull/755 must have broken things on both Anvil and Perlmutter (maybe others).  I tested on Chrysalis but not elsewhere and that seems to have come back to bite me.

xylar:
This merge switches to installing `cmake` and `cxx-compiler` in the deployment environment (not the final compass environment) and use them to build Jigsaw.  This prevents build tools from being left in the compass conda environment that later interfere with system tools and/or the spack environment.

This merge also removes a temporary fix from #757 on Anvil that is not needed after this "real" fix.

Checklist

closes #759

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -409,11 +409,6 @@ def get_env_vars(machine, compiler, mpilib):
                    f'export I_MPI_F77=ifort\n' \
                    f'export I_MPI_F90=ifort\n'
 
-    if machine == 'anvil':
-        # Anvil seems to need libs from here to build successfully
-        env_vars = f'{env_vars}' \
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I was able to build a conda environment (including Jigsaw and Jigsaw-Python) on Perlmutter with this fix, and then build MPAS-Ocean, whereas I saw the errors in #759 without this fix.

I was also able to do the same on Anvil and Chrysalis, whereas I had seen similar errors to those in #759 on Anvil before the fix.

PR #769, opened 2024-02-01
Fix finding the NetCDF library in the Jigsaw build

xylar:
This merge explicitly points to the libnetcdf that should be used in the Jigsaw build.  It also breaks the jigsaw build into its own function.

Checklist

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -340,8 +340,8 @@ def build_conda_env(env_type, recreate, mpi, conda_mpi, version,
             check_call(commands, logger=logger)
 
         if recreate or update_jigsaw:
-
-            build_jigsaw(activate_env, source_path, env_path, logger)
+            build_jigsaw(activate_env, conda_base, source_path, env_path,
+                         logger)
 
         # install (or reinstall) compass in edit mode
         print('Installing compass\n')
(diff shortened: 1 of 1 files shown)


xylar:
The Jigsaw build in Polaris was finding `libnetcdf.so` in `/usr/lib64` on Compy, and that was subsequently causing a bunch of build failures.  Instead, we want to be using the conda-forge version of `libnetcdf`, which this branch does.  In the process, we have to call `cmake` for Jigsaw ourselves to pass the extra argument, rather than relying on Jigsaw-Python to do it for us.

dengwirda:
@xylar if you're being forced to include netcdf for jigsaw I expect there must be an issue / bug on my end --- it's intended to be an optional (and experimental) dependency that the cmake should handle...

xylar:
@dengwirda. I'm keen to look into this in the long term but for now I just need to get this working so it isn't preventing me from updating Polaris for reasons unrelated to Jigsaw.

I don't think it's necessarily that you're doing anything wrong in Jigsaw.  It's just that we would need some way for someone calling `python setup.py build_external` to pass hints about where to find NetCDF.  The issue for me is that I need to be building Jigsaw in one conda environment where it's safe to have compilers installed but installing it into another, where I have NetCDF.  But also `find_library()` seemed to choose the `libnetcdf.so` from `/usr/lib64` over the one in a conda environment even when that was available.  So, yeah, I basically need a way to tell cmake where to find NetCDF rather than having it guess.

dengwirda:
Okay, makes sense --- I've added an issue for me to look at going forward.

PR #770, opened 2024-02-08
Add `files_for_e3sm` step for remapping rms tidal velocity

xylar:
This field is needed for any meshes that will run in the `TMIX` configuration described here:
https://github.com/E3SM-Ocean-Discussion/E3SM/pull/76
The field gets remapped from a dataset that @irenavankova has preprocessed from the CATS tidal model,
see https://web.lcrc.anl.gov/public/e3sm/mpas_standalonedata/mpas-ocean/tidal_mixing/README.txt.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/__init__.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/__init__.py
@@ -28,6 +28,9 @@ from compass.ocean.tests.global_ocean.files_for_e3sm.remap_iceberg_climatology i
 from compass.ocean.tests.global_ocean.files_for_e3sm.remap_sea_surface_salinity_restoring import (  # noqa: E501
     RemapSeaSurfaceSalinityRestoring,
 )
+from compass.ocean.tests.global_ocean.files_for_e3sm.remap_tidal_mixing import (  # noqa: E501
+    RemapTidalMixing,
+)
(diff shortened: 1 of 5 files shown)


xylar:
## Testing

I ran the `files_for_e3sm` test case (the one for pre-existing meshes) and pointed it to the `IcoswISC30E3r5` mesh.

I was able to generate the RMS tidal velocity:
xylar

My version differs only slightly from @irenavankova's:
irena

The places where they differ are where there is very little overlap between the original grid and the MPAS mesh:
mask
and results from a different approach to renormalization when there is less than 10% overlap between the meshes.

xylar:
@irenavankova, this is ready for you to try out and review.

We should talk about the renormalization strategy.  What I chose to do was to renormalize if there is more than 1% overlap and to set the velocity to zero if there is less.

It seemed like you had a different approach to cells with less than 10% overlap where you always renormalized as if the overlap were 1%.  I think my approach is a little more standard and should lead to smoother results but I don't think it matters much because only a handful of cells are affected.

xylar:
@irenavankova, I sent you an invitation to collaborate here.  If you accept, I can add you as a reviewer but you can still comment without being an official reviewer.

irenavankova (approved):
All works for me

xylar:
@irenavankova, thanks for testing this out!

PR #774, opened 2024-02-13
Add FRIS meshes

irenavankova:
Meshes with high resolution in the southern Weddell Sea and beneath the Filchner Ronne Ice Shelf (FRIS). The meshes go down to 1, 2, 4, and 8 km beneath FRIS and on the continental shelf in front of it.


xylar (approved):
@irenavankova, this looks excellent to me!  I tested the 8 km mesh on Chrysalis with a recent version of this branch that included all but the last 3 commits.

I think these meshes will be great to have documented and available both as examples for others to work from and for future work focused on the Weddell Sea.

PR #783, opened 2024-02-27
New Mesh: NARRM14to30E3r2

xylar:
Long name: NARRM14to30kmL80E3SMv3r2

This version of the North American Regionally Refined Mesh (RRM) is the same as has the same distribution of resolution as in [NARRM14to30E3r1](https://github.com/MPAS-Dev/compass/pull/782), except that it has 80 instead of 64 vertical levels.

```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/narrm14to30e3r2
```

Checklist


xylar:
## Testing

I ran all Compass test cases (even the mesh) for NARRM14, with results here:
```
/lcrc/group/e3sm/ac.xylar/compass_1.2/chrysalis/e3smv3-meshes/narrm14to30e3r2
```
The `inputdata` files have been copied to `/lcrc/group/e3sm/data/inputdata/` for testing in E3SM.

PR #784, opened 2024-02-27
Fix typo culeed --> culled

xylar:
Checklist


xylar:
## Testing

I was able to run the `culled_mesh` step for #782 with this fix, whereas it failed without it.

PR #787, opened 2024-03-08
Fix bug in limiting cull distance

trhille:
This merge removes an unnecessary assertion about the value of `cull_distance`, since it is handled by other logic.

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -369,7 +369,7 @@ def set_cell_width(self, section_name, thk, bed=None, vx=None, vy=None,
     # cell size in the final mesh. There may be a more rigorous way to set
     # that distance.
     if dist_to_edge is not None:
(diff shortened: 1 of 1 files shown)


matthewhoffman:
@trhille , looking at this again, I think the proper thing to do is delete the assertion and its message.  What this bit of code was meant to do was stick coarse resolution out past the boundary of the culled mesh so that jigsaw runs faster by not having the highest resolution over a huge area that later gets thrown away.  This bit of code does that automatically by setting the coarsest resolution at a (hopefully) safe distance beyond the specified cull_distance.  That logic has nothing to do with the relation between cull_distance and the high_dist, so no check actually needs to happen related to that.  I'm not sure what I was thinking when I wrote the assertion, but I don't think it makes sense, and we should just remove it.

matthewhoffman (approved):

PR #790, opened 2024-03-12
Fixes to landlocked cells when culling global meshes 

xylar:
Since critical land blockages can cause new cells to be landlocked, we need to remove landlocked cells only after we have added land blockages to the land mask.

This merge also fixes a small issue where the `latitude_threshold` parameter for widening critical passages is now supplied from the appropriate config option, whereas it had previously been hard coded by mistake.

Checklist

### DIFF
--- a/compass/ocean/mesh/cull.py
+++ b/compass/ocean/mesh/cull.py
@@ -407,8 +407,9 @@ def _cull_mesh_with_logging(logger, with_cavities, with_critical_passages,
 
         # Alter critical passages to be at least two cells wide, to avoid sea
         # ice blockage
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the QUwISC240 `mesh` test case with this branch.  Previously, a single cell on the coast of the Eastern Pacific was marked as landlocked only in the second pass (which is meant to "detect landlocked" cells at ice-shelf calving fronts).  This meant that it was marked as an ice-shelf cell (`landIceMask = 1`) instead of a land cell (culled from the mesh).

before

With this branch, the cell is now correctly identified as land and culled from the mesh.

after

xylar:
I need to check if this problem has affected finer resolution meshes.

xylar:
Thanks for taking a look, @cbegeman.

PR #798, opened 2024-03-19
Only make tidal mixing files for meshes with cavities

xylar:
Since the tidal model used to produce the dataset only covers the Southern Ocean, there is not much value in having the reampped dataset for meshes without cavities.  If we want to make a more general mixing parameterization that uses tidal information, we would need to use data from a global tidal model for the input dataset in the future.

Checklist


xylar:
This addresses these comments from E3SM-Ocean-Discussion:
https://github.com/E3SM-Ocean-Discussion/E3SM/pull/85#issuecomment-2004545677
https://github.com/E3SM-Ocean-Discussion/E3SM/pull/85#issuecomment-2004574463
https://github.com/E3SM-Ocean-Discussion/E3SM/pull/85#issuecomment-2004820198

PR #800, opened 2024-03-26
Perform flood fills on the land-ice mask

xylar:
This merge makes sure the land-ice mask does not have holes (by flood filling an "ocean" mask that is the inverse of the land-ice mask) and then ensures that all land-ice cells are connected to each other (by flood filling the land-ice mask from the south pole). 

Checklist


xylar:
I also ran the `pr` suite and it completed successfully.

PR #802, opened 2024-03-27
Add a tides test case for the VR45to5 mesh

sbrus89:
This PR adds the workflow to generate the mesh and initial conditions for the VR45to5 barotropic tides mesh, run the simulation, and perform the global tidal constituent error analysis. It relies on the developments from https://github.com/E3SM-Project/E3SM/pull/6310. This case reproduces the MPAS-Ocean configuration that gives the best deep water M2 RMSE accuracy (3.3 cm compared to TPXO9) to date.

Checklist
* [x] Document (in a comment titled Testing in this PR) any testing that was used to verify the changes


sbrus89:
**Testing**

The mesh generated looks like:
mesh_EC

Land ice draft is included in the init step
ice_draft

Bathymetric slopes and buoyancy frequency climatology are calculated/interpolated in the init step for topographic wave drag parameterizations:
bathy_slope
Nb

The harmonic constituents are compared against TPXO9 in the analysis step:
M2_plot
N2_plot
S2_plot
O1_plot
K1_plot

compass/ocean/tests/tides/mesh/vr45to5/__init__.py
```diff
@@ -0,0 +1,284 @@
...
+from mpas_tools.logging import check_call
+from skimage.filters import gaussian, scharr
+
+from compass.mesh.spherical import SphericalBaseStep
+
+# from skimage.filters import farid
+# from skimage.filters.rank import median, percentile
+
+
+class VRTidesMesh(SphericalBaseStep):
```
sbrus89, line 16: This should probably inherit from `QuasiUniformSphericalMeshStep` then I can reuse the `setup` and `run` methods (if I refactor the return variables from `build_cell_width_lat_lon` and the input variables to `make_jigsaw_mesh`

compass/ocean/tests/tides/dem/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+
+        Parameters
+        ----------
+        test_case : compass.ocean.tests.tides.init.Init
+            The test case this step belongs to
+        """
+        super().__init__(test_case=test_case, name='pixel',
+                         ntasks=1, min_tasks=1, openmp_threads=1)
+
+        self.add_input_file(
```
dengwirda, line 27: Building these DEM datasets could be done once and cached somewhere, as it's pretty expensive and is typically only redone when GEBCO release a new bathymetry (once per year).

compass/ocean/tests/tides/dem/dem_pixel.py
dengwirda, line 1: I'll keep these tools updated in https://github.com/dengwirda/dem for the latest GEBCO etc.

compass/ocean/tests/tides/forward/namelist.ocean
```diff
@@ -4,10 +4,14 @@ config_calendar_type = 'noleap'
 config_pio_num_iotasks = 0
 config_pio_stride = 1
 config_time_integrator = 'RK4'
+config_use_mom_del2 = .true.
+config_mom_del2 = 3.333
+config_use_mom_del4 = .true.
```
dengwirda, line 9: I didn't do a lot of tuning on the del^k dissipation, so this could be looked at in the future if there's interest.

compass/ocean/tests/tides/init/initial_state.py
dengwirda, line 1: Since the model is BTR-only, the following is a simple non-iterative adjustment of the SSH due to ISC pressure.

compass/ocean/tests/tides/mesh/vr45to5/__init__.py
xylar: If possible, that does seem better.  But we would want to avoid updating `QuasiUniformSphericalMeshStep` if it can be avoided.

dengwirda (commented):
Thanks @sbrus89 this runs for me successfully on pm-cpu using https://github.com/E3SM-Project/E3SM/pull/6310. A few things I noticed:
- It looks like the plots show TPXO8 rather than TPXO9?
- It looks like the wave-drag step may be printing quite a bit of output to the logs which could be suppressed? Also some long python loops which may be expensive.
- You may see improved performance if jigsaw is upgraded to a more recent version that runs with multi-threading + accelerated steepest descent.

dengwirda:
Not relevant to this PR specifically, but since the errors in M2 are quite competitive here, I wonder if tuning for other constituents too (RMSE_total: https://gmd.copernicus.org/articles/14/1125/2021/) may make sense in the future, so as not to over fit on M2.

mark-petersen (approved):
Approving based on @sbrus89's testing above. Thank you! It is great to have this published test case in the repository!

xylar:
@sbrus89, as I just mentioned on slack, my attempt to update the submodule was not very successful.  https://github.com/E3SM-Project/E3SM/pull/6310 seems to be causing segfaults on Chrysalis with Intel and OpenMPI.

sbrus89:
@xylar, the segfault issues should be fixed via #825, which turns off topographic wave drag for the global ocean test cases. We should only test `config_use_topographic_wave_drag=.true.` for tides cases like the one in this PR.

xylar:
Almost there, just waiting on #827

xylar:
@sbrus89, now that #827 is in, can you rebase and test one last time?  Then, I'll merge.

sbrus89:
@xylar, yes I will do that. Thanks for helping with the submodule update!

xylar:
Following up from a month ago, a reminder to rebase and retest.

sbrus89:
@xylar, sorry for the big delay on this. Everything is working for me after the rebase now that I sorted out the issues with the ice shelf cavity culling step. I'm retesting the Icos7 test to make sure it is still working as expected. I'll let you know once I've verified that it is.

bmooremaley:
@xylar, reviving this PR. I confirmed the `Icos7` and `VR45to5` cases build for Compass 2.0.0 and reproduce the figures originally posted by @sbrus89. New commits add tempest parameters to `vr45to5.cfg`, correct the tidal constituents specified for comparison in `Icos7.cfg` and `vr45to5.cfg`, and fix a bug in the harmonic analysis step.

xylar:
@bmooremaley and @sbrus89, I don't have time to test this right now but I'm happy to merge it if you think it's ready to go in.

sbrus89:
@xylar, @bmooremaley has tested this thoroughly. I'm comforatble with you merging it.

PR #805, opened 2024-03-29
Cap fractions from topography to one

xylar:
Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

closes #804

PR #807, opened 2024-03-31
New Mesh: SOwISC12to30E3r3

xylar:
Long name: SOwISC12to30kmL80E3SMv3r3

* 30 km elsewhere

This mesh differs from r2 in #753 in that:
* holes in land-ice mask are filled and isolated "islands" of land ice are removed (#800, #844, #845, #846, #848)
* the `landIceFraction` is capped at 1 (#805)
* it has 80 (instead of 64) vertical levels with a highest vertical resolution of 2 m (instead of 10)
* it specifies a minimum of 5 layers with a minimum thickness of 2 m (usring #812)
* the sea surface height is iteratively adjusted for hydrostatic balance (following #813)
* Omega and pi have been corrected in https://github.com/E3SM-Project/E3SM/pull/6481, brought in through #839
* Bug fix to the haney-number-constrained vertical coordinate code in https://github.com/E3SM-Project/E3SM/pull/6535, updated in the E3SM-Project submodule in https://github.com/MPAS-Dev/compass/pull/849
* Bug fix related to masking of ice thickness: #854
* Bug fix related to constraining ice thickness to flotation thickness: #855 

```
/lcrc/group/e3sm/ac.xylar/compass_1.4/chrysalis/e3smv3-meshes/sowisc12to30e3r3
```

Checklist


xylar:
We will switch to 80 levels

xylar:
@cbegeman and I discussed switching to having `landIcePressure` be supplied from observations and `ssh` be adjusted to be dynamically consistent.  (At the moment, Compass does the opposite.)  This would be more consistent with plans for MALI coupling in the near future.

compass/ocean/tests/global_ocean/mesh/so12to30/namelist.init
```diff
@@ -0,0 +1,3 @@
+config_rx1_min_levels = 8
+config_rx1_min_layer_thickness = 2.5
+config_global_ocean_minimum_depth = 20.0
```
xylar, line 3: @cbegeman and I discussed that the 20 m minimum water-column thickness is probably overkill.

What feels more reasonable?  10 m?

xylar:
My preliminary testing shows that the new ssh adjustment technique from #813 works just fine and converges sufficiently in 4 iterations (change of about 10 cm).  So I think this is ready to include in the next version of the mesh.

compass/ocean/tests/global_ocean/mesh/so12to30/so12to30.cfg
```diff
@@ -2,28 +2,47 @@
...
+
+# minimum number of vertical levels, both in the open ocean and in ice-shelf
+# cavities
+min_levels = 5
+cavity_min_levels = ${min_levels}
+
+# minimum thickness of layers in ice-shelf cavities at the beginning and end
+# of iterative ssh init
+cavity_min_layer_thickness_initial = 2.0
+cavity_min_layer_thickness_final = 1.0
```
cbegeman, line 40: @xylar Assuming that we still have a max sea ice pressure-related depression of 5m, I just want to make sure that minimum column thickness > 5m where sea ice is allowed to form. Can you confirm this?

cbegeman: @xylar Is this the right file to be looking at? `/lcrc/group/e3sm/ac.xylar/compass_1.4/chrysalis/e3smv3-meshes/sowisc12to30e3r3/ocean/global_ocean/SOwISC12to30/WOA23/files_for_e3sm/ocean_initial_condition/mpaso.SOwISC12to30E3r3.20240705.nc`

xylar: I think the answer is yes but I would treat this as the "official" one for now:
```
/lcrc/group/e3sm/data/inputdata/ocn/mpas-o/SOwISC12to30E3r3/mpaso.SOwISC12to30E3r3.20240705.nc
```
I believe those 2 files are identical.

cbegeman: There is 1 cell outside ice shelf cavities (`landIceMask == 0`) with column thickness < 5.0m (4.4m).
There are 9 cells inside ice shelf cavities (`landIceMask == 1`) with column thickness < 5.0m (min column thickness of 2.1m, 5 active layers with min column thickness of 0.36m).

We definitely need to set the minimum thickness outside cavities with enough of a buffer that it is >5m, since we can apply ~5m worth of sea ice pressure. I'd propose a min column thickness of 10m everywhere. We don't strictly need this in ice shelf cavities since sea ice can't form there but it does seem that we're getting greater ssh perturbations there than elsewhere.

xylar: @cbegeman, I agree.  Let's stick with a 10-m minimum water-column thickness both inside and outside the cavities.

I'm adding a histogram (and a min/max bit of text) to the initial-condition plots to at least partly cover this sanity check.

xylar:
Here are the figures from the `initial_state` step, showing the water-column thickness > 10 m in both the open ocean and in cavities:

initial_state
vertical_grid

compass/ocean/tests/global_ocean/mesh/so12to30/so12to30.cfg
xylar: @cbegeman, I synced over the new version of the file.  Could you give it another look and make sure you see in the E3SM initial condition what I was seeing in the MPAS-Ocean initial condition, namely that the minimum column thickness isn't too much thinner than 10 m?

```
/lcrc/group/e3sm/data/inputdata/ocn/mpas-o/SOwISC12to30E3r3/mpaso.SOwISC12to30E3r3.20240711.nc
```
(Same as before but with today's date.)

cbegeman: @xylar This is looking good to me! Thanks

Number of cells with column thickness less than 10m inside ice shelf cavities	61 (min column thickness 5.5m)
Number of cells with column thickness less than 10m outside ice shelf cavities 	10 (min column thickness 8.3m)
Number of cells with column thickness less than 5m     0

xylar: Okay, it's interesting how much some water columns still thin during dynamic adjustment!  I hadn't realized it would be that much.  It's certainly better to have more of a safety margin!

xylar: At first, I was worried I messed something up with MPAS-Ocean the initial condition (before dynamic adjustment) but that doesn't seem to be the reason there were such thin water columns in the E3SM initial condition.

xylar:
Needs https://github.com/E3SM-Project/E3SM/pull/6535

PR #812, opened 2024-04-06
Add config options for min levels

xylar:
Previously, we specified the minimum thickness of the global ocean, which was set to 10 m.

This is not a good approach with different vertical coordinates.  If we want to use the same settings across various vertical coordinates, it is more useful to require a minimum number of levels.

This merge changes the global ocean `initial_state` step to use config options to set the number of vertical levels both in the open ocean and in ice-shelf cavities.  Relatedly, we also add a config option to set the minimum layer thickness in cavities.

This change will affect the vertical 3D vertical cooridinate in all meshes (hopefully for the better).

For meshes that use variants of the 80-layer vertical coordinate, we set the minimum number of layers that sets the column thickness to either 10 or 20 m.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/global_ocean.cfg
+++ b/compass/ocean/tests/global_ocean/global_ocean.cfg
@@ -49,8 +49,17 @@ btr_dt_per_km = 1.5
 
 ## config options related to the initial_state step
 
+# minimum number of vertical levels, both in the open ocean and in ice-shelf
+# cavities
+min_levels = 3
+cavity_min_levels = ${min_levels}
+
+# minimum thickness of layers in ice-shelf cavities
+cavity_min_layer_thickness = 1.0
+
 # Maximum allowed Haney number for configurations with ice-shelf cavities
-rx1_max = 20
+rx1_max = 20.0
+
 # the number of iterations of topography smoothing (0 means no smoothing)
 topo_smooth_num_passes = 0
 # The distance in km over which the Gaussian filter is applied
--- a/compass/ocean/tests/global_ocean/init/initial_state.py
+++ b/compass/ocean/tests/global_ocean/init/initial_state.py
@@ -176,6 +176,25 @@ class InitialState(Step):
         plot_vertical_grid(grid_filename='vertical_grid.nc', config=config,
                            out_filename='vertical_grid.png')
 
+        min_levels = config.getint('global_ocean', 'min_levels')
+        # minimum depth will be the depth of the middle of the minimum layer
+        # so that MPAS-Ocean init mode will convert it back to the same
+        # minimm number of levels
+        min_depth = 0.5 * (interfaces[min_levels - 1] +
+                           interfaces[min_levels])
+        namelist = {'config_global_ocean_minimum_depth': f'{min_depth}'}
+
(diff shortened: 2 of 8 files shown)


xylar:
## Testing

I ran the `pr` suite successfully on Chrysalis.  I didn't update the cached file for Icos and IcoswISC because I don't think these changes are critical for testing forward runs.

I also successfully ran `mesh`, `init` and `performance_test` test cases for all Icos and QU setups (240 and 30 km resolution, with and without cavities).  I verified that the resulting initial conditions had the expected minimum `maxLevelCell` and minimum `bottomDepth`.

I decided not to test other meshes here because new meshes always get their own PR anyway, so these properties should be verified as part of typical mesh review.

xylar:
@cbegeman, could you take a quick look at these changes and see if they make sense to you?  I think I've done sufficient testing already, I just want a quick sanity check on my work.

My thinking was essentially that it's weird to set something like:
```
config_rx1_min_levels = 8
config_rx1_min_layer_thickness = 2.5
config_global_ocean_minimum_depth = 20.0
```
to get similar behavior in the open ocean and in cavities. Also, it's a little weird that we set `config_global_ocean_minimum_depth` but then internally in init mode it converts that back to a number of levels (the first level with its bottom deeper than the min).  So I figure it's better to have the min for both inside and outside of cavities (which are the same by default).  And then you can set the min layer thickness in cavities as well.  Maybe so you set it to get roughly the same minimum water-column thickness but that's still not a requirement.

compass/ocean/tests/global_ocean/global_ocean.cfg
```diff
@@ -49,8 +49,17 @@ btr_dt_per_km = 1.5
 
 ## config options related to the initial_state step
 
+# minimum number of vertical levels, both in the open ocean and in ice-shelf
+# cavities
+min_levels = 3
+cavity_min_levels = ${min_levels}
+
+# minimum thickness of layers in ice-shelf cavities
+cavity_min_layer_thickness = 1.0
```
xylar, line 58: Before I merge this, I'm going to determine what this actually needs to be to prevent ssh adjustment from failing.  This may take a little experimentation.

cbegeman (approved):
@xylar Fantastic! I agree; this is much clearer/cleaner. I'm curious to know what you learn from ssh_adjustment. We could also consider changing `min_levels` to 2 in `global_ocean.cfg` to reflect the real minimum, esp if it isn't actually used in any of the cases.

xylar:
> We could also consider changing min_levels to 2 in global_ocean.cfg to reflect the real minimum, esp if it isn't actually used in any of the cases.

I don't think we want to be running with 2 levels on purpose.  I think that was a oversight in the Icos mesh.  I don't think there will be detrimental effects but I also don't think we carefully considered what physics gets disabled in those shallow layers.  So I would think we want a minimum of 3 levels in any future Icos30 meshes we might make.

xylar:
@cbegeman, thanks so much for reviewing!

xylar:
I'm going to go ahead with this.  I can make special branches for the various meshes if necessary in order to create tags that can recreate them.

PR #814, opened 2024-04-09
Update to mache v1.22.0

xylar:
This version of mache uses newer OpenMPI libraries for Intel, needed to prevent jobs from hanging during cancellation, blocking nodes.

### DIFF
--- a/compass/version.py
+++ b/compass/version.py
@@ -1,1 +1,1 @@
-__version__ = '1.3.0-alpha.2'
+__version__ = '1.3.0-alpha.3'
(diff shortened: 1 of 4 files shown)


xylar:
The new `mache` package isn't yet available so I'll test and merge this in the morning.

xylar:
## Testing

On Chrysalis, I successfully ran the `pr` suite with this version of `mache` after I updated the shared spack build.  I verified that it is using the new OpenMPI required for Intel on Chrysalis.

I also tested with Gnu for good measure (mostly to make sure jobs didn't hang with CG status) and everything was okay.

PR #813, opened 2024-04-09
For initialization with ice-shelf cavities, adjust SSH not land ice pressure

xylar:
Checklist
* [x] User's Guide has been updated
* [x] Developer's Guide has been updated
* [x] API documentation in the Developer's Guide (`api.rst`) has any new or modified class, method and/or functions listed
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [x] The `E3SM-Project` submodule has been updated with relevant E3SM changes
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes
* [x] Update cached files needed for the `pr` suite to run

PR #810, opened 2024-04-04
Update to 1.4.0-alpha.1

xylar:
### Updates:
- 

### Other Changes:
- Add metis 5.1.0 built with `int64` and `real64` (needed for large partitions)

Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

## Testing

MPAS-Ocean with `pr`:
- [ ] Anvil
  - [ ] intel and impi
  - [ ] gnu and openmpi
- [ ] Chicoma
  - [ ] gnu and mpich
- [ ] Chrysalis
  - [ ] intel and openmpi
  - [ ] gnu and openmpi
- [ ] Compy
  - [ ] intel and impi
- [ ] Perlmutter 
  - [ ] gnu and mpich

MALI with `full_integration`:
- [ ] Chicoma
  - [ ] gnu and mpich
- [ ] Chrysalis 
  - [ ] gnu and openmpi
- [ ] Perlmutter
  - [ ] gnu and mpich

## Deployed

MPAS-Ocean with `pr`:
- [ ] Anvil
  - [ ] intel and impi
  - [ ] intel and openmpi
  - [ ] gnu and openmpi
- [ ] Chicoma
  - [ ] gnu and mpich
- [ ] Chrysalis
  - [ ] intel and openmpi
  - [ ] gnu and openmpi
- [ ] Compy
  - [ ] intel and impi
- [ ] Perlmutter 
  - [ ] gnu and mpich

MALI with `full_integration`:
- [ ] Chicoma
  - [ ] gnu and mpich
- [ ] Chrysalis 
  - [ ] gnu and openmpi
- [ ] Perlmutter
  - [ ] gnu and mpich

PR #816, opened 2024-04-14
Clean up `iceshelf` framework

xylar:
This PR has some clean up needed to make SSH adjustment in the `iceshelf` framework work for adjusting SSH instead of land-ice pressure.

This was pulled out of #813 because that work no longer uses the `iceshelf` framework.

Checklist

### DIFF
--- a/compass/ocean/iceshelf.py
+++ b/compass/ocean/iceshelf.py
@@ -79,11 +79,17 @@ def adjust_ssh(variable, iteration_count, step, update_pio=True,
         step.update_namelist_pio('namelist.ocean')
     partition(ntasks, config, logger)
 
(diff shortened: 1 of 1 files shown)


cbegeman (approved):
I think these changes look good. You can merge when you're ready. Happy to do testing if you like.

xylar:
I tried testing this on Chrysalis on Friday and was getting unexpected non-BFB results in tests without ice-shelf cavities.  I think it may be just the Chrysalis weirdness that we have been experiencing lately.  I'm going to wait until #810 has gone in because I'm hoping that will fix the issues.

xylar:
## Testing

I ran the `pr` suite with these changes, using the current `master` as a baseline.  All tests passed.

PR #803, opened 2024-03-28
Add conservative remapping of observational datasets onto Greenland Ice Sheet mesh

trhille:
This merge adds conservative remapping of observational data sets (thickness, bed topography, velocity, and uncertainties) onto the Greenland Ice Sheet mesh, and handles some of the peculiarities of the other gridded datasets (unit conversions, sign conventions, etc).

Checklist
* [x] User's Guide has been updated
* [x] Developer's Guide has been updated
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #750, opened 2023-12-22
Updated data interpolation for Antarctica mesh generation case

matthewhoffman:
This PR adds automates interpolation of observational data from gridded datasets to Antarctic mesh within COMPASS. This takes care of the peculiarities of the current gridded compilation dataset (antarctica_8km_2020_10_20.nc), as well as using conservative remapping directly from the high-resolution BedMachineAntarctica and MeASUReS velocity datasets. There is a fairly heavy degree of pre-processing done to get the BedMachine and MeASUReS datasets ready to be used here. The pre-processing includes renaming variables, setting reasonable _FillValue and missing_value attributes, extrapolating fields to avoid interpolation ramps at ice margins, updating mask values, and raising the bed topography at Lake Vostok to ensure a flat ice surface. Those data files and processing scripts currently live here on Chicoma: /usr/projects/climate/trhille/data. Eventually that pre-processing could be integrated into a new step in COMPASS, or the processed data files could be added to the server on Anvil and downloaded as needed.  However, until then, this test case provides a reproducible workflow for setting up Antarctic meshes at varying resolutions

Checklist
* [x] User's Guide has been updated
* [x] Developer's Guide has been updated
* [x] API documentation in the Developer's Guide (`api.rst`) has any new or modified class, method and/or functions listed
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #817, opened 2024-04-16
Landice interpolation unification 

andrewdnolan:
With PRs #803 and #750 the `mesh_gen` testcases in the `greenland` and `antarctica` test groups now make better use of mesh generation utilities within the landice framework (e.g. `compass.landice.mesh`).   But there is still some redundant code as it relates to interpolation (e.g. `interp_ais_bedmachine` and `interp_ais_bedmachine` could be generalized to `interp_ais`). Continuing along that line, there is no real need to have separate interpolation functions for Antarctica (`interp_ais`) and Greenland (`interp_gis`). 

The purpose of the PR is generalize the existing interpolation functions to be applicable the standard datasets (i.e. measures and bedmachine) and standard icesheets (Antarctica and Greenland). With the generalized interpolation routine, the respective `mesh_gen` testcases are cleaned up and unified as much as possible. 

Checklist

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -444,7 +444,7 @@ def get_dist_to_edge_and_gl(self, thk, topg, x, y,
         logger.info('WARNING: window_size was set to a value smaller'
                     ' than high_dist and/or high_dist_bed. Resetting'
                     f' window_size to {max(high_dist, high_dist_bed)}, '
-                    ' which is  max(high_dist, high_dist_bed)')
+                    ' which is max(high_dist, high_dist_bed)')
         window_size = max(high_dist, high_dist_bed)
 
     dx = x[1] - x[0]  # assumed constant and equal in x and y
@@ -971,8 +971,8 @@ def preprocess_ais_data(self, source_gridded_dataset,
 
         tic = time.perf_counter()
         logger.info(f"Beginning interpolation for {field}")
-        # NOTE: Do not need to evaluate the extrapolator at all gridcells.
-        #       Only needed for ice-free gricells, since it's NN extrapolation
+        # NOTE: Do not need to evaluate the extrapolator at all grid cells.
+        #       Only needed for ice-free grid cells, since it's NN extrapolation
         data.variables[field][0, :] = interp(xGrid, yGrid)
         toc = time.perf_counter()
         logger.info(f"Interpolation completed in {toc - tic} seconds")
@@ -1027,7 +1027,7 @@ def interp_gridded2mali(self, source_file, mali_scrip, nProcs, dest_file, proj,
         MALI input file to which data should be remapped
 
     proj: str
-        projection of the source dataset ...
+        projection of the source dataset
 
     variables: str or list of strings
 
@@ -1069,8 +1069,6 @@ def interp_gridded2mali(self, source_file, mali_scrip, nProcs, dest_file, proj,
     check_call(args, logger=logger)
 
     # Generate remapping weights
-    # Testing shows 5 badger/grizzly nodes works well.
-    # 2 nodes is too few. I have not tested anything in between.
     logger.info('generating gridded dataset -> MPAS weights')
     args = ['srun', '-n', nProcs, 'ESMF_RegridWeightGen',
             '--source', source_scrip,
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -1091,66 +1091,6 @@ def interp_gridded2mali(self, source_file, mali_scrip, nProcs, dest_file, proj,
     # conflicts when multiple interpolations are done?
 
 
-def interp_ais_measures(self, data_path, mali_scrip, nProcs, dest_file):
-    """
-    Interpolates MEASURES ice velocity dataset
-    to a MALI mesh
-
-    Parameters
-    ----------
-    data_path : str
-        path to AIS datasets, including BedMachine
-
-    mali_scrip : str
-        name of scrip file corresponding to destination MALI mesh
-
-    nProcs : int
-        number of processors to use for generating remapping weights
-
-    dest_file: str
-        MALI input file to which data should be remapped
-    """
-
-    logger = self.logger
-
-    logger.info('creating scrip file for velocity dataset')
-    # Note: writing scrip file to workdir
-    args = ['create_SCRIP_file_from_planar_rectangular_grid.py',
-            '-i',
-            os.path.join(data_path,
-                         'antarctica_ice_velocity_450m_v2_edits_extrap.nc'),
-            '-s',
-            'antarctica_ice_velocity_450m_v2.scrip.nc',
-            '-p', 'ais-bedmap2',
-            '-r', '2']
-    check_call(args, logger=logger)
-
-    # Generate remapping weights
-    logger.info('generating gridded dataset -> MPAS weights')
-    args = ['srun', '-n', nProcs, 'ESMF_RegridWeightGen',
-            '--source',
-            'antarctica_ice_velocity_450m_v2.scrip.nc',
-            '--destination', mali_scrip,
-            '--weight', 'measures_to_MPAS_weights.nc',
-            '--method', 'conserve',
-            "--netcdf4",
-            "--dst_regional", "--src_regional", '--ignore_unmapped']
-    check_call(args, logger=logger)
-
-    logger.info('calling interpolate_to_mpasli_grid.py')
-    args = ['interpolate_to_mpasli_grid.py',
-            '-s',
-            os.path.join(data_path,
-                         'antarctica_ice_velocity_450m_v2_edits_extrap.nc'),
-            '-d', dest_file,
-            '-m', 'e',
-            '-w', 'measures_to_MPAS_weights.nc',
-            '-v', 'observedSurfaceVelocityX',
-                  'observedSurfaceVelocityY',
-                  'observedSurfaceVelocityUncertainty']
-    check_call(args, logger=logger)
-
-
 def clean_up_after_interp(fname):
     """
     Perform some final clean up steps after interpolation
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -1050,11 +1050,12 @@ def interp_gridded2mali(self, source_file, mali_scrip, nProcs, dest_file, proj,
     source_scrip = __guess_scrip_name(os.path.basename(source_file))
     weights_filename = "gridded_to_MPAS_weights.nc"
 
-    if variables != "all":
-        # make sure this is a list
-
-        # if list, then join the list making it a space seprated list for cli
-        variables = " ".join(variables)
+    # make sure variables is a list, encompasses the variables="all" case
+    if isinstance(variables, str):
+        variables = [variables]
+    if not isinstance(variables, list):
+        raise TypeError("Arugment 'variables' is of incorrect type, must"
+                        " either the string 'all' or a list string")
 
     logger.info('creating scrip file for source dataset')
     # Note: writing scrip file to workdir
@@ -1085,7 +1086,7 @@ def interp_gridded2mali(self, source_file, mali_scrip, nProcs, dest_file, proj,
             '-d', dest_file,
             '-m', 'e',
             '-w', weights_filename,
-            '-v', variables]
+            '-v'] + variables
 
     check_call(args, logger=logger)
 
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -442,7 +442,7 @@ def get_dist_to_edge_and_gl(self, thk, topg, x, y,
     elif window_size < min(high_dist, high_dist_bed):
         logger.info('WARNING: window_size was set to a value smaller'
                     ' than high_dist and/or high_dist_bed. Resetting'
-                    f' window_size to {max(high_dist, high_dist_bed)},'
+                    f' window_size to {max(high_dist, high_dist_bed)}, '
                     ' which is  max(high_dist, high_dist_bed)')
         window_size = max(high_dist, high_dist_bed)
 
@@ -969,6 +969,8 @@ def preprocess_ais_data(self, source_gridded_dataset,
 
         tic = time.perf_counter()
         logger.info(f"Beginning interpolation for {field}")
+        # NOTE: Do not need to evaluate the extrapolator at all gridcells.
+        #       Only needed for ice-free gricells, since it's NN extrapolation
         data.variables[field][0, :] = interp(xGrid, yGrid)
         toc = time.perf_counter()
         logger.info(f"Interpolation completed in {toc - tic} seconds")
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 import time
 from shutil import copyfile
@@ -1002,15 +1003,15 @@ def preprocess_ais_data(self, source_gridded_dataset,
     return preprocessed_gridded_dataset
 
 
-def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
+def interp_gridded2mali(self, source_file, mali_scrip, nProcs, dest_file, proj,
+                        variables="all"):
     """
-    Interpolates BedMachine thickness and bedTopography dataset
-    to a MALI mesh
+    Interpolate gridded dataset (e.g. MEASURES, BedMachine) onto a MALI mesh
 
     Parameters
     ----------
-    data_path : str
-        path to AIS datasets, including BedMachine
+    source_file : str
+        filepath to the source gridded datatset to be interpolated
 
     mali_scrip : str
         name of scrip file corresponding to destination MALI mesh
@@ -1020,19 +1021,45 @@ def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
 
     dest_file: str
         MALI input file to which data should be remapped
+
+    proj: str
+        projection of the source dataset ...
+
+    variables: str or list of strings
+
     """
+
+    def __guess_scrip_name(filename):
+
+        # try searching for string followed by a version number
+        match = re.search(r'(^.*[_-]v\d*[_-])+', filename)
+
+        if match:
+            # slice string to end of match minus one to leave of final _ or -
+            base_fn = filename[:match.end() - 1]
+        else:
+            # no matches were found, just use the filename (minus extension)
+            base_fn = os.path.splitext(filename)[0]
+
+        return f"{base_fn}.scrip.nc"
 
     logger = self.logger
 
-    logger.info('creating scrip file for BedMachine dataset')
+    source_scrip = __guess_scrip_name(os.path.basename(source_file))
+    weights_filename = "gridded_to_MPAS_weights.nc"
+
+    if variables != "all":
+        # make sure this is a list
+
+        # if list, then join the list making it a space seprated list for cli
+        variables = " ".join(variables)
+
+    logger.info('creating scrip file for source dataset')
     # Note: writing scrip file to workdir
     args = ['create_SCRIP_file_from_planar_rectangular_grid.py',
-            '-i',
-            os.path.join(data_path,
-                         'BedMachineAntarctica_2020-07-15_v02_edits_floodFill_extrap_fillVostok.nc'),  # noqa
-            '-s',
-            'BedMachineAntarctica_2020-07-15_v02.scrip.nc',
-            '-p', 'ais-bedmap2',
+            '-i', source_file,
+            '-s', source_scrip,
+            '-p', proj,
             '-r', '2']
     check_call(args, logger=logger)
 
@@ -1041,10 +1068,9 @@ def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
     # 2 nodes is too few. I have not tested anything in between.
     logger.info('generating gridded dataset -> MPAS weights')
     args = ['srun', '-n', nProcs, 'ESMF_RegridWeightGen',
-            '--source',
-            'BedMachineAntarctica_2020-07-15_v02.scrip.nc',
+            '--source', source_scrip,
             '--destination', mali_scrip,
-            '--weight', 'BedMachine_to_MPAS_weights.nc',
+            '--weight', weights_filename,
             '--method', 'conserve',
             "--netcdf4",
             "--dst_regional", "--src_regional", '--ignore_unmapped']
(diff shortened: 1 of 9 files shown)


compass/landice/tests/greenland/mesh.py
```diff
@@ -56,136 +59,67 @@ def run(self):
...
-                '-w', 'measures_to_MPAS_weights.nc',
-                '-v', 'observedSurfaceVelocityX',
-                'observedSurfaceVelocityY',
-                'observedSurfaceVelocityUncertainty']
-        check_call(args, logger=logger)
-
-        logger.info('Marking domain boundaries dirichlet')
-        args = ['mark_domain_boundaries_dirichlet.py',
-                '-f', 'GIS.nc']
-        check_call(args, logger=logger)
```
andrewdnolan: As far as I can tell `mark_domain_boundaries_dirichlet.py` is called in the `compass.landice.mesh.build_mali_mesh` function, so I've deleted the additional call here. 

Is that OK to do? Or should this be left here?

compass/landice/tests/greenland/mesh.py
```diff
@@ -56,136 +59,67 @@ def run(self):
...
-        data.variables['thickness'][:] *= (data.variables['iceMask'][:] > 1.5)
-
-        mask = np.logical_or(
-            np.isnan(
-                data.variables['observedSurfaceVelocityUncertainty'][:]),
-            data.variables['thickness'][:] < 1.0)
-        mask = np.logical_or(
-            mask,
-            data.variables['observedSurfaceVelocityUncertainty'][:] == 0.0)
-        data.variables['observedSurfaceVelocityUncertainty'][0, mask[0, :]] = 1.0  # noqa
```
andrewdnolan: This was replaced by the cal to `compass.landice.mesh.clean_up_after_interp`. Given the code here and that within the function are basically identical, I'm assuming that's OK.

compass/landice/tests/greenland/mesh.py
```diff
@@ -80,3 +129,22 @@ def run(self):
...
+        dHdtErr = data.variables["observedThicknessTendencyUncertainty"][:]
+        # Arbitrary 5% uncertainty; improve this later
+        dHdtErr = np.abs(dHdt) * 0.05
+        # large uncertainty where data is missing
+        dHdtErr[np.abs(dHdt) > 1.0] = 1.0
+        dHdt[np.abs(dHdt) > 1.0] = 0.0  # Remove ridiculous values
+        data.variables["observedThicknessTendency"][:] = dHdt
+        data.variables["observedThicknessTendencyUncertainty"][:] = dHdtErr
+
+        data.close()
```
andrewdnolan, line 150: I'm having a hard time finding the analogous manipulation to Antarctica mesh within the `antarctica` testgroup and/or framework functions. 

If this is Greenland specific then I'll wrap with is a function (within the step) and call it a day, but I have a feeling that there's something I'm missing and a similar operation gets done for Antarctica.

compass/landice/mesh.py
```diff
@@ -1008,15 +1009,15 @@ def preprocess_ais_data(self, source_gridded_dataset,
     return preprocessed_gridded_dataset
 
 
-def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
+def interp_gridded2mali(self, source_file, mali_scrip, nProcs, dest_file, proj,
```
andrewdnolan, line 1012: Open to better names for the function, but thought this was at least a start.

compass/landice/mesh.py
```diff
@@ -1026,106 +1027,75 @@ def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
...
-            '--source',
-            'antarctica_ice_velocity_450m_v2.scrip.nc',
+            '--source', source_scrip,
             '--destination', mali_scrip,
-            '--weight', 'measures_to_MPAS_weights.nc',
+            '--weight', weights_filename,
             '--method', 'conserve',
             "--netcdf4",
             "--dst_regional", "--src_regional", '--ignore_unmapped']
     check_call(args, logger=logger)
```
andrewdnolan, line 1083: Trevor's call to `ESMF_RegridWeightGen` has the additional flag: `"-i", "-64bit_offset",`, which is missing here. Is that flag needed for Greenland, but not Antarctica? If so I can add an option to this function to toggle it. 

Or should that flag be added to the default arguments here?

compass/landice/tests/greenland/mesh.py
trhille: Good catch. Yes, I believe that's redundant. Go ahead and remove.

compass/landice/mesh.py
trhille: I'm not sure if it's needed. We can test it both way and remove it if it's not necessary.

compass/landice/tests/greenland/mesh.py
trhille: I think the treatment of `basalHeatFlux` can just be removed, because we've updated the source datasets and added checks on the sign in the interpolation script.
The treatment of `dHdt` and `dHdtErr` are Greenland specific because the gridded dataset for Greenland has missing uncertainties, etc, so don't worry about doing this for Antarctica.

compass/landice/mesh.py
andrewdnolan: Okay so the `-i` flag is just an alias for the `-64bit_offset` flag, which is already included here. So I'm definitely going to drop that. 

From the `ESMF_RegridWeightGen` documentation: 
>    --64bit_offset    - an optional argument specifying that the weight file
                        will be created in the NetCDF 64-bit offset format
                        to allow variables larger than 2GB.  Note the 64-bit
                        offset format is not supported in the NetCDF version
                        earlier than 3.6.0.  An error message will be generated
                        if this flag is specified while the application is
                        linked with a NetCDF library earlier than 3.6.0.

But we have the `--netcdf4` flag included, so I'm not totally sure if the 2GB variables are still a constraint in netcdf4. I'll do some looking to figure that one out.

andrewdnolan:
# Testing 

- Machine: perlmutter
- Compass version: 1.3.0-alpha.2

__Command Run :__
```
compass setup -n 0 45 -w wrkdir
```

## Results
```
landice/antarctica/mesh_gen
  * step: mesh
  test execution:      SUCCESS
  test runtime:        37:42
landice/greenland/mesh_gen
  * step: mesh
  test execution:      SUCCESS
  test runtime:        15:36
Test Runtimes:
37:42 PASS landice_antarctica_mesh_gen
15:36 PASS landice_greenland_mesh_gen
Total runtime 53:30
```

compass/landice/mesh.py
andrewdnolan: From the testing below,  the weights file for Antartica mesh is pretty large: 
```bash
$ du -sh ./gridded_to_MPAS_weights.nc
# 16G     ./gridded_to_MPAS_weights.nc
```

So doing a pretty crude loop over the variables in that file with:
```python 
import xarray as xr 

with xr.open_dataset("gridded_to_MPAS_weights.nc") as ds:
    for var in ds.variables:
        print(f"{var:<20}: {ds[var].nbytes/1e+9:1.3f} GB")
# ...
# yc_a                : 1.239 GB
# yc_b                : 0.003 GB
# xc_a                : 1.239 GB
# xc_b                : 0.003 GB
# yv_a                : 4.956 GB
# xv_a                : 4.956 GB
# ...
```
Looks like yv_a and xv_a are greater than 2GB, and that was generated using `--netcdf4` without `--64bit_offset`, so I think we are OK.

Also, per @matthewhoffman suggestion; I've taken a look through the [`pyremap` source code](https://github.com/MPAS-Dev/pyremap/blob/1836cdaa9220d81ca7880580402297f1b50dcfbf/pyremap/remapper.py#L216-L221) and looks like only the `` flag is used there. 

So, with all this I'm going to leave the `--64bit_offset` flag off since it seems unnecessary.

andrewdnolan: Thanks for catching that!

trhille:
CI seems to be broken. I'm going to close and reopen to see if that jump-starts it.

trhille:
@andrewdnolan, the `landice/antarctica/mesh_gen` test passes execution for me. It fails baseline comparison with the current head of `compass/main`. That's not necessarily a problem, but we should look into it.
`landice/greenland/mesh_gen` fails execution however. Here's the traceback:
```
Setting cell_width in outer regions to max_spac for 654997 cells
calling preprocess_gridded_data
      Failed
Exception raised while running the steps of the test case
  File "/global/cfs/cdirs/fanssie/users/trhille/compass/compass/run/serial.py", line 322, in _log_and_run_test
    _run_test(test_case, available_resources)
  File "/global/cfs/cdirs/fanssie/users/trhille/compass/compass/run/serial.py", line 419, in _run_test
    _run_step(test_case, step, test_case.new_step_log_file,
  File "/global/cfs/cdirs/fanssie/users/trhille/compass/compass/run/serial.py", line 470, in _run_step
    step.run()
  File "/global/cfs/cdirs/fanssie/users/trhille/compass/compass/landice/tests/greenland/mesh.py", line 89, in run
    preprocessed_gridded_dataset = preprocess_gridded_data(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/global/cfs/cdirs/fanssie/users/trhille/compass/compass/landice/mesh.py", line 912, in preprocess_gridded_data
    gg.variables['thk'][0, :, :] *= floodFillMask
  File "/pscratch/sd/t/trhille/miniforge3/envs/dev_compass_1.4.0-alpha.7/lib/python3.11/site-packages/numpy/ma/core.py", line 4415, in __imul__
    self._data.__imul__(other_data)
ValueError: operands could not be broadcast together with shapes (2816,1664) (1408,832) (2816,1664) 
```

Any idea what might be causing this?

compass/landice/tests/greenland/mesh.py
```diff
@@ -56,142 +58,79 @@ def run(self):
...
                 self, section_name=section_name,
-                gridded_dataset='greenland_2km_2024_01_29.epsg3413.nc',
+                gridded_dataset=source_gridded_dataset_2km,
                 flood_fill_start=[100, 700])
 
+        # Preprocess the gridded GrIS source datasets to work
+        # with the rest of the workflow
+        logger.info('calling preprocess_gridded_data')
+        preprocessed_gridded_dataset = preprocess_gridded_data(
+            self, source_gridded_dataset_1km, floodMask)
```
andrewdnolan, line 90: The array that causes the broadcasting error is half the size `(1408,832)` of the array that's attempting to be broadcast against `(2816,1664)`. My hunch is that this has to do with mixing and matching the 1km and 2km gridded datasets in the highlighted code section. 

That being said in the version of the code on main, there seems to be the same mixing and matching of 1km and 2km datasets. My guess is that this the root of the problem, but why it's a problem with the unified interpolation functions and not the previous version of the code isn't super clear to me yet.

andrewdnolan:
# Testing (Part One) 
----

__NOTE__: Testing needs to be done in two stages. With the addition of globally sorted cell ID's ([4bdc715](https://github.com/MPAS-Dev/compass/pull/817/commits/4bdc715874ccf13d1b06f273bfd176319e63cbc8)) there's no easy way difference against the mesh generated on `main`, because the `sort_mesh` function "resets" the `indexToCellID` array. (i.e. For both sorted and unsorted meshes the `indexToCellID` array is monotonically increasing from 1 to `nCells`). 

__So__, the baseline against `main` needs to be done from [626f11a](https://github.com/MPAS-Dev/compass/pull/817/commits/626f11a6a97ea5a8f3d681c58d39156d193b71de) then visual testing can be done from [4bdc715](https://github.com/MPAS-Dev/compass/pull/817/commits/4bdc715874ccf13d1b06f273bfd176319e63cbc8) to confirm the sorting has been implemented. 

----
Tested on `perlmutter` in `/pscratch/sd/a/anolan/landice_interpolation`: 

```
# using compass env script form main branch
compass setup -n 0 45 -w main@869dafea8 -f custom.cfg
# using compass env script form landice/interpolation_unification branch
compass setup -n 0 45 -w interpolation_unification@626f11a6a -f custom.cfg -b $PSCRATCH/landice_interpolation/main@869dafea8/

pushd main@869dafea8 && sbatch job_script.custom.sh && popd
pushd interpolation_unification@626f11a6a && sbatch job_script.custom.sh && popd
```
Which results in (for the PR branch):
```
landice/antarctica/mesh_gen
  * step: mesh
  test execution:      SUCCESS
  baseline comparison: FAIL
  see: case_outputs/landice_antarctica_mesh_gen.log
  test runtime:        35:26
landice/greenland/mesh_gen
  * step: mesh
  test execution:      SUCCESS
  baseline comparison: FAIL
  see: case_outputs/landice_greenland_mesh_gen.log
  test runtime:        17:08
Test Runtimes:
35:26 FAIL landice_antarctica_mesh_gen
17:08 FAIL landice_greenland_mesh_gen
Total runtime 53:13
FAIL: 2 tests failed, see above.
```

So both test pass but unfortunately the baseline comparison fails. To investigate things a little further, I've written my own mesh differencing script following compass's baseline approach: https://gist.github.com/andrewdnolan/162339430d602229379c83c5f72200f9 

Running that script on the mesh produced from `626f11a` results in: 
image

@matthewhoffman has OKed these difference offline, but might be worth double checking on this.

andrewdnolan:
# Testing (Part 2)

With the addition of the sorting commit, here is the `indexToCellID` array:
image

compass/landice/mesh.py
```diff
@@ -1020,106 +1026,76 @@ def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
...
-            'antarctica_ice_velocity_450m_v2.scrip.nc',
-            '-p', 'ais-bedmap2',
+            '-i', source_file,
+            '-s', source_scrip,
+            '-p', proj,
             '-r', '2']
     check_call(args, logger=logger)
 
     # Generate remapping weights
+    # Testing shows 5 badger/grizzly nodes works well.
```
trhille, line 1073:
```suggestion
```

compass/landice/mesh.py
```diff
@@ -1020,106 +1026,76 @@ def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
...
+    # 2 nodes is too few. I have not tested anything in between.
```
trhille, line 1074:
```suggestion
```

compass/landice/mesh.py
```diff
@@ -1020,106 +1026,76 @@ def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
...
-    logger.info('creating scrip file for velocity dataset')
+    source_scrip = __guess_scrip_name(os.path.basename(source_file))
+    weights_filename = "gridded_to_MPAS_weights.nc"
+
+    # make sure variables is a list, encompasses the variables="all" case
+    if isinstance(variables, str):
+        variables = [variables]
+    if not isinstance(variables, list):
+        raise TypeError("Arugment 'variables' is of incorrect type, must"
+                        " either the string 'all' or a list string")
```
trhille, line 1061: Should this say "list of strings"?

compass/landice/mesh.py
```diff
@@ -1020,106 +1026,76 @@ def interp_ais_bedmachine(self, data_path, mali_scrip, nProcs, dest_file):
...
-            '-w', 'measures_to_MPAS_weights.nc',
-            '-v', 'observedSurfaceVelocityX',
-                  'observedSurfaceVelocityY',
-                  'observedSurfaceVelocityUncertainty']
+            '-w', weights_filename,
+            '-v'] + variables
+
     check_call(args, logger=logger)
 
+    # should I delted the weights file, since that could cause namespace
```
trhille, line 1096: Needs cleanup

compass/landice/tests/antarctica/mesh.py
```diff
@@ -61,37 +60,38 @@ def run(self):
...
+        data_path = section_ais.get('data_path')
+        measures_filename = section_ais.get("measures_filename")
+        bedmachine_filename = section_ais.get("bedmachine_filename")
+
+        measures_dataset = os.path.join(data_path, measures_filename)
+        bedmachine_dataset = os.path.join(data_path, bedmachine_filename)
 
         section_name = 'mesh'
 
+        # do we want to add this to the config file?
```
trhille, line 77:
```suggestion
        # TODO: do we want to add this to the config file?
```

compass/landice/tests/antarctica/mesh.py
```diff
@@ -131,10 +138,16 @@ def run(self):
...
         # Now perform bespoke interpolation of geometry and velocity data
         # from their respective sources
-        interp_ais_bedmachine(self, data_path, dst_scrip_file, nProcs,
-                              self.mesh_filename)
-        interp_ais_measures(self, data_path, dst_scrip_file, nProcs,
-                            self.mesh_filename)
+        interp_gridded2mali(self, bedmachine_dataset, dst_scrip_file, nProcs,
+                            self.mesh_filename, src_proj, variables="all")
+
+        # only interpolate a subset of MEASURES varibles onto the MALI mesh
```
trhille, line 144:
```suggestion
        # only interpolate a subset of MEaSUREs variables onto the MALI mesh
```

compass/landice/tests/greenland/mesh.py
```diff
@@ -56,142 +61,73 @@ def run(self):
...
-
+        dst_scrip_file = f"{self.mesh_filename.split('.')[:-1][0]}_scrip.nc"
+        scrip_from_mpas(self.mesh_filename, dst_scrip_file)
+
+        # Now perform bespoke interpolation of geometry and velocity data
+        # from their respective sources
+        interp_gridded2mali(self, bedmachine_dataset, dst_scrip_file, nProcs,
+                            self.mesh_filename, src_proj, variables="all")
+
+        # only interpolate a subset of MEASURES varibles onto the MALI mesh
```
trhille, line 106:
```suggestion
        # only interpolate a subset of MEaSUREs variables onto the MALI mesh
```

compass/landice/tests/greenland/mesh.py
```diff
@@ -203,18 +139,22 @@ def run(self):
...
-        dHdtErr[np.abs(dHdt) > 1.0] = 1.0
-        dHdt[np.abs(dHdt) > 1.0] = 0.0  # Remove ridiculous values
-        data.variables["observedThicknessTendency"][:] = dHdt
-        data.variables["observedThicknessTendencyUncertainty"][:] = dHdtErr
-
-        data.close()
+        # Use threshold of |dHdt| > 1.0 to determine invalid data
+        mask = np.abs(dHdt) > 1.0
+        # Assign 100% uncertainty where data is missing
+        dHdtErr = dHdtErr.where(~mask, 1.0)
```
trhille, line 153: Is the tilde correct here? Won't this overwrite all the good data and leave the missing/bad data in place?

compass/landice/tests/greenland/mesh.py
```diff
@@ -203,18 +139,22 @@ def run(self):
...
+        # Remove ridiculous values
+        dHdt = dHdt.where(~mask, 0.0)
```
trhille, line 155: Same question regarding tilde.

compass/landice/tests/greenland/mesh_gen/mesh_gen.cfg
```diff
@@ -50,3 +47,23 @@ use_speed = True
...
+[greenland]
+# path to directory containing BedMachine and Measures datasets
+# (default value is for Perlmutter)
+data_path = /global/cfs/cdirs/fanssie/standard_datasets/GIS_datasets/
+
+# filename of the BedMachine thickness and bedTopography dataset
+# (default value is for Perlmutter)
+bedmachine_filename = BedMachineGreenland-v5_edits_floodFill_extrap.nc
+
+# filename of the MEASURES ice velocity dataset
```
trhille, line 60:
```suggestion
# filename of the MEaSUREs ice velocity dataset
```

docs/developers_guide/landice/framework.rst
```diff
@@ -56,13 +56,10 @@ clean up steps after interpolation for the AIS mesh case.
...
 
-:py:func:`compass.landice.mesh.interp_ais_bedmachine()` interpolates BedMachine
-thickness and bedTopography dataset to a MALI mesh, accounting for masking of
-the ice extent to avoid interpolation ramps.
-
-:py:func:`compass.landice.mesh.interp_ais_interp_ais_measures()` interpolates
-MEASURES ice velocity dataset to a MALI mesh, accounting for masking at the ice
-edge and extrapolation.
+:py:func:`compass.landice.mesh.interp_gridded2mali()` interpolates gridded data
+(e.g. BedMachine thickness or MEASURES ice velocity) to a MALI mesh, accounting
```
trhille, line 60:
```suggestion
(e.g. BedMachine thickness or MEaSUREs ice velocity) to a MALI mesh, accounting
```

docs/developers_guide/landice/framework.rst
```diff
@@ -56,13 +56,10 @@ clean up steps after interpolation for the AIS mesh case.
...
+for masking of the ice extent to avoid interpolation ramps. This functions works
+for either the Antarctic or Greenland Icesheets.
```
trhille, line 62:
```suggestion
for both Antarctica and Greenland. 
```

trhille (changes requested):
@andrewdnolan, this is looking really good! I have a lot of comments but they are all tiny. I'm running tests now.

andrewdnolan:
Screenshot 2024-09-30 at 2 38 55 PM

Here are the errors with the L-inf norm added. I'll updated the  gist linked above in case you want to run things yourself.

trhille:
Excellent, thanks. I'm not worried about these diffs at all.

compass/landice/tests/greenland/mesh.py
```diff
@@ -203,18 +139,22 @@ def run(self):
...
-        # large uncertainty where data is missing
-        dHdtErr[np.abs(dHdt) > 1.0] = 1.0
-        dHdt[np.abs(dHdt) > 1.0] = 0.0  # Remove ridiculous values
-        data.variables["observedThicknessTendency"][:] = dHdt
-        data.variables["observedThicknessTendencyUncertainty"][:] = dHdtErr
-
-        data.close()
+        # Use threshold of |dHdt| > 1.0 to determine invalid data
+        mask = np.abs(dHdt) > 1.0
+        # Assign 100% uncertainty where data is missing
```
trhille, line 152:
```suggestion
        # Assign very large uncertainty where data is missing
```

This is absolute uncertainty, since we can't assign a percent uncertainty to missing data.

compass/landice/mesh.py
```diff
@@ -681,6 +683,8 @@ def build_mali_mesh(self, cell_width, x1, y1, geom_points,
     dsMesh = xarray.open_dataset('base_mesh.nc')
     logger.info('culling mesh')
     dsMesh = cull(dsMesh, logger=logger)
+    logger.info('soring mesh')
```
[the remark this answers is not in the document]
xylar: As I noted on Slack, this needs to be done after the `convert()` call below, which will undo the sorting.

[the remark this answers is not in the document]
andrewdnolan: Thanks for the heads up @xylar. As documented below, your suggestion fixes things!

compass/landice/tests/greenland/mesh.py
andrewdnolan: Yes, it is (although I agree it's counterintuitive). If you check the `xr.DataArray.where` docstring: 

> other : scalar, DataArray, Dataset, or callable, optional
>     Value to use for locations in this object where ``cond`` is False.
>     By default, these locations are filled with NA. If a callable, it must
>     expect this object as its only parameter. 

So we need to make the mask `False` inorder for the replacing to work, which is why the tilde is there. I got burned in a first round of testing when I left if off. 

I could switch to using `xr.where`, which has the opposite behavior (i.e. `True` values are replaced), if that's more intuitive to read.

trhille: `xr.where` is definitely more intuitive, but given that this works and we now have documentation of this confusion, I'm fine with leaving it as-is.

andrewdnolan:
# Testing (Part 3) 
Here some plots to confirm the sorting is indeed working correctly, now that it's been placed after all the `convert` function calls. 

Slice zeros of edges on cell: 
edgesOnCell_slice0_sorted

Index to cell ID: 
indexToEdgeID_sorted

trhille (approved):
Awesome work, @andrewdnolan! I've confirmed that both test cases pass and that the edge sorting issue is resolved. I visually inspected all the important fields, but am relying on the diffs you provided above in order to bless changes.

PR #801, opened 2024-03-26
New Mesh: RRSwISC6to18E3r5

xylar:
Long name: RRSwISC6to18L80E3SMv3r5

This RRS (Rossby-radius scaled) mesh has:
* 6 km resolution near the poles
* 18 km resolution at the equator

Mesh, initial condition, dynamic adjustment and files for E3SM will be on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.3/chrysalis/e3smv3-meshes/rrswisc6to18e3r5
```

This mesh is similar to RRSwISC6to18E3r4 #754 but with #800 to fill in holes in the land-ice mask and with a minimum water-column thickness of 20 m.

Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #819, opened 2024-04-20
New Mesh: RRSwISC6to18E3r7

xylar:
Long name: RRSwISC6to18L80E3SMv3r7

```
/lcrc/group/e3sm/ac.xylar/compass_1.3/chrysalis/e3smv3-meshes/rrswisc6to18e3r7
```

This mesh is similar to RRSwISC6to18E3r5 #801 but with a minimum fraction for partial bottom cells of 0.2 instead of 0.1.

Checklist


xylar:
It seems unlikely we will use this mesh so closing for now.

PR #820, opened 2024-04-29
Set simulation length in the SLM namelist to 300 years 

hollyhan:
In current setup, ISMIP6-2300 historical simulation is performed for 15 years between 2000-2015 CE, and the projection simulations start from year 2015 until 2301. Previously, the sea-level model namelist was set such that the total length of simulation covered by the SLM is 285 years to cover years between 2015-2300, but since we've decided to run the coupled MALI-SLM model for the historical period as well, the simulation length option in the SLM naemlist needs to be corrected to cover the whole 300 years (2000-2300). This PR includes the change.


matthewhoffman (approved):
@hollyhan , thanks for correcting this!

PR #822, opened 2024-05-09
Add histograms of areaCell and dcEdge to init plots

xylar:
Last 2 plots in:
initial_state

Checklist

### DIFF
--- a/compass/ocean/plot.py
+++ b/compass/ocean/plot.py
@@ -1,8 +1,9 @@
+import datetime
+
+import matplotlib.pyplot as plt
+import numpy as np
 import xarray
 import xarray.plot
-import numpy as np
-import datetime
-import matplotlib.pyplot as plt
 from matplotlib.font_manager import FontProperties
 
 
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the IcoswISC30 mesh and init tests on Chrysalis to produce the image above.

xylar:
@golaz and @proteanplanet, are you okay with these or are there more tweaks you'd like to see?

xylar:
The frequency is on a log scale for other plots, Perhaps that would make sense here, too.  I'll try that.

xylar:
initial_state

Yeah, probably the version on a log scale makes a bit more sense for consistency and because there's a bit more info there.

xylar:
I fixed the min/max values to be the original ones (m^2 and m, rather than km^2 and km) to hopefully avoid potential confusion.

proteanplanet (approved):
Approved by visual inspection.  Thanks for adding this as a QC metric.

xylar:
Thanks @proteanplanet!

PR #809, opened 2024-04-02
New Mesh: RRS6to18E3r6

xylar:
Long name: RRS6to18L80E3SMv3r6

This RRS (Rossby-radius scaled) mesh has:
* 6 km resolution near the poles
* 18 km resolution at the equator

Mesh, initial condition, dynamic adjustment and files for E3SM will be on Chrysalis at:
```
/lcrc/group/e3sm/ac.xylar/compass_1.3/chrysalis/e3smv3-meshes/rrs6to18e3r6
```

This mesh is similar to RRSwISC6to18E3r5 #801 but without ice-shelf cavities.

Checklist
* [ ] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #828, opened 2024-06-10
Don't make DIB files if not ice-shelf cavities

xylar:
Checklist


xylar:
## Testing

This was successfully tested as part of creating the mesh in #809

PR #829, opened 2024-06-10
New Mesh: SOwISC12to30E3r4

xylar:
Long name: SOwISC12to30kmL80E3SMv3r4

* 30 km elsewhere
 
This mesh differs from r3 in https://github.com/MPAS-Dev/compass/pull/807 in that:

* MALI topography as well as masks for land ice, floating ice and ocean get combined with BedMachine Antarctica/GEBCO.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
+++ b/compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
@@ -43,7 +43,7 @@ class RemapMaliTopography(RemapTopography):
         mesh_name : str
             The name of the MPAS mesh to include in the mapping file
 
-                    mali_ais_topo : str, optional
+        mali_ais_topo : str, optional
             Short name for the MALI dataset to use for Antarctic Ice Sheet
             topography
         """
@@ -230,27 +230,27 @@ class RemapMaliTopography(RemapTopography):
                     'landIceThkObserved']:
             ds_out[var] = ds_mali[var]
 
-        ds_out['landIceFracObserved'] = ds_mali['landIceFracConserve']
+        ds_out['landIceFracObserved'] = ds_mali['landIceFrac']
 
         ds_out['landIceFloatingFracObserved'] = (
             ds_mali['landIceFrac'] -
             ds_mali['landIceGroundedFrac'])
 
-        for var in ['maliFracBilinear', 'maliFracConserve']:
+        for var in ['maliFrac']:
             mali_field = ds_mali[var]
             mali_field = xr.where(mali_field.notnull(), mali_field, 0.)
             ds_out[var] = mali_field
 
         # for now, blend topography at calving fronts, but we may want a
         # smoother blend in the future
-        alpha = ds_mali.landIceFracBilinear
+        alpha = ds_mali.landIceFrac
         ds_out['bed_elevation'] = (
             alpha * ds_mali.bed_elevation +
             (1.0 - alpha) * ds_bedmachine.bed_elevation)
 
-        alpha = ds_out.maliFracConserve
+        alpha = ds_out.maliFrac
         ds_out['oceanFracObserved'] = (
-            alpha * ds_mali.oceanFracConserve +
+            alpha * ds_mali.oceanFrac +
             (1.0 - alpha) * ds_bedmachine.oceanFracObserved)
 
         ds_out['ssh'] = ds_out.landIceDraftObserved
--- a/compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
+++ b/compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
@@ -1,0 +1,252 @@
+import os
+
+import numpy as np
+import xarray as xr
+from mpas_tools.cime.constants import constants
+from mpas_tools.io import write_netcdf
+from mpas_tools.logging import check_call
+from pyremap import MpasCellMeshDescriptor
+
+from compass.ocean.mesh.remap_topography import RemapTopography
+from compass.parallel import run_command
+
+
+class RemapMaliTopography(RemapTopography):
+    """
+    A step for remapping bathymetry and ice-shelf topography from a MALI
+    datase to a global MPAS-Ocean mesh and combining it with a global
+    topography dataset
+
+    Attributes
+    ----------
+    mali_ais_topo : str
+        Short name for the MALI dataset to use for Antarctic Ice Sheet
+        topography
+    """
+
+    def __init__(self, test_case, base_mesh_step, mesh_name, mali_ais_topo):
+        """
+        Create a new step
+
+        Parameters
+        ----------
+        test_case : compass.ocean.tests.global_ocean.mesh.Mesh
+            The test case this step belongs to
+
+        base_mesh_step : compass.mesh.spherical.SphericalBaseStep
+            The base mesh step containing input files to this step
+
+        remap_topography : compass.ocean.mesh.remap_topography.RemapTopography
+            A step for remapping topography. If provided, the remapped
+            topography is used to determine the land mask
+
+        mesh_name : str
+            The name of the MPAS mesh to include in the mapping file
+
+                    mali_ais_topo : str, optional
+            Short name for the MALI dataset to use for Antarctic Ice Sheet
+            topography
+        """
+        super().__init__(test_case=test_case, base_mesh_step=base_mesh_step,
+                         mesh_name=mesh_name)
+        self.mali_ais_topo = mali_ais_topo
+
+        self.add_output_file(filename='mali_topography_remapped.nc')
+
+    def setup(self):
+        """
+        Set up the step in the work directory, including downloading any
+        dependencies.
+        """
+        super().setup()
+
+        mali_filename = self.config.get('remap_mali_topography',
+                                        'mali_filename')
+        self.add_input_file(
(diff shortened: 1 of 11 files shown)


xylar:
This is based off of #807 (which is based off of #813) so it will be a bit before this is ready to go in.  It also needs documentation.

xylar:
## Comparison with SOwISC12to30E3r3

Here are the `landIceMask` fields for the SORRM mesh with MALI topography:
mali_topo_land_ice_mask

And the same but with BedMachine from #807:
sowiscr3_land_ice_mask

xylar:
## Updated Comparison with SOwISC12to30E3r3

Here are the `landIceMask` fields for the SORRM mesh with MALI (`AIS_2to10km`) topography:
sorrm_ais_2to10km_land_ice_mask


And the same but with BedMachine from #807:
sorrm_r3_land_ice_mask

xylar:
I was able to run the workflow through the performance test.  In a discussion with @cbegeman, @matthewhoffman and @trhille, we decided we'd like to work with a 4km initial condition that's more similar to the 2km one I used in the plots above, and that that initial condition would also be swapped into https://github.com/E3SM-Project/E3SM/pull/6440, the PR bringing in the new MALI AIS meshes and initial conditions.

xylar:
## Another Updated Comparison with SOwISC12to30E3r3

Here are the `landIceMask` fields for the SORRM r4 mesh with MALI (`AIS_4to20km`) topography:
sorrmr4_land_ice_mask

And the same but SORRM r3 with BedMachine topography from #807:
sorrmr3_land_ice_mask

xylar:
The same as above but zooming in on the Amundsen Sea.

SORRM r4:
sorrmr4_amundsen

SORRM r3:
sorrmr3_amundsen

cbegeman:
@xylar have you been looking at all at `landIceFloatingFraction`?

xylar:
Not yet. I should.

xylar:
## landIceFloatingFraction

### SORRM r4
sorrmr4_land_ice_floating_fraction

### SORRM r3
sorrmr3_land_ice_floating_fraction

@cbegeman, no surprises to me.  What do you think?

cbegeman:
@xylar Thanks for sharing those. 

The new map is nice in that the "shadow" islands of floating ice are removed near the peninsula. 

It looks like not all the cells near the coastline are capped at 0.5. Isn't that supposed to be the case? Or is that only true when there's grounded ice upstream as opposed to exposed rock?

What do you think of the pocket of partial ice cover at Brunt near an island? I'm thinking that since it's part of the ice mask there won't be sea ice formation. And there also won't be any interaction with the atmosphere, right (like fluxes scaled by the ice fraction)? We could potentially get a bit of frazil there since frazil doesn't change the surface pressure. But I don't know if this is really worth worrying about. Should we look at the land ice pressure field and make sure that that's reasonably smooth. I see one other low fraction "island" between Ross and Getz.

xylar:
> It looks like not all the cells near the coastline are capped at 0.5. Isn't that supposed to be the case? Or is that only true when there's grounded ice upstream as opposed to exposed rock?

No that's no longer the case.  That was a change with r3.  I think we want to allow melting in partial open ocean, partial ice-shelf cells for conservation.  Otherwise, we need to reroute the missing melt somewhere else.  When melt is happening in the coupler, this will presumably be happening anyway so this is to make melt computed in MPAS-Ocean more similar to what will happen in the coupler.

cbegeman:
@xylar That's ringing some bells. Thanks for the reminder

xylar:
> What do you think of the pocket of partial ice cover at Brunt near an island? I'm thinking that since it's part of the ice mask there won't be sea ice formation. And there also won't be any interaction with the atmosphere, right (like fluxes scaled by the ice fraction)? We could potentially get a bit of frazil there since frazil doesn't change the surface pressure. But I don't know if this is really worth worrying about. Should we look at the land ice pressure field and make sure that that's reasonably smooth. I see one other low fraction "island" between Ross and Getz.

These are presumably places where isolated pockets of open ocean have been filled in because they're not resolved enough.  Frazil could certainly form here but I would expect these mostly to be pretty passive areas.  You could be right though that it's worth looking at ice draft or land-ice pressure here to make sure these areas don't look *too* weird.  They could be isolated pockets of essentially zero-thickness ice shelf where water just gets trapped.

xylar:
The files are in:
```
/lcrc/group/e3sm/ac.xylar/compass_1.5/chrysalis/e3smv3-meshes/sowisc12to30e3r4/ocean/global_ocean/SOwISC12to30/MALI_topo_AIS_4to20km/
```
and also staged for testing in:
```
/lcrc/group/e3sm/data/inputdata/ocn/mpas-o/SOwISC12to30E3r4
```

xylar:
I will make a proper review page soon.

xylar:
@matthewhoffman found issues with `landIceThickness` in the dataset we are using from MALI.  I'll need to rerun once that's fixed.

xylar:
The land-ice mask for the Amundsen sea is looking reasonable with the latest MALI topography dataset (`AIS_4to20km_20241118_relaxation_0TGmelt_10yr.20241224.nc`):
land_ice_mask

cbegeman:
@xylar Great! I'm wondering about the distribution of `0 > landIceGroundedFraction < 1`. It looks from the figure above that these cells are only along the transition between grounded and floating ice or open ocean, and not along the transition from grounded ice to land above sea level. Is this intentional?

xylar:
@cbegeman, I think we do a flood fill to remove all land-ice-to-land so, yeah, that transition is intentionally not there and I don't think we want it.

xylar:
I think that might have been one of the important fixes I had to make in SORRMr3 -- if I accidentally didn't eliminate the land-ice-to-land transition, I ended up with weird ice drafts above sea level in some cases because of interpolation between areas that are above and below sea level.

cbegeman:
On giving this a little more thought I agree that the hard transition is what we want. Thanks for explaining the reasoning a bit more.

xylar:
## Testing

Tested on an 11-year G-case as described in https://github.com/E3SM-Ocean-Discussion/E3SM/pull/120

xylar:
@matthewhoffman and @cbegeman, if you could review this along with https://github.com/E3SM-Project/E3SM/pull/7195, that would be wonderful!

compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
```diff
@@ -0,0 +1,263 @@
...
+        logger.info('Modifying MALI topography fields and names')
+
+        ds_mali = xr.open_dataset('mali_topography_orig.nc')
+        if 'Time' in ds_mali.dims:
+            ds_mali = ds_mali.isel(Time=0)
+        bed = ds_mali.bedTopography
+        thickness = ds_mali.thickness
+
+        ice_density = config.getfloat('remap_topography', 'ice_density')
+        ocean_density = config.getfloat('remap_topography', 'ocean_density')
```
cbegeman, line 129: Check that config section has option `ocean_density` in `setup` since this is not included in all mesh cfg files?

maltrud:
the differences in Arctic SSS appear to be due to changing from PHC2.0 to WOA23 for the SSS climatology.  here's a plot of annual average SSS (WOA - PHC):
Screenshot 2025-03-31 at 12 21 21 PM

cbegeman (approved):
Approving on the basis of testing and visual inspection of these code changes. Thanks @xylar!

xylar:
@maltrud, I agree, that's what I attribute those changes to, as well.  The freshening seems like a move in the wrong direction, right?  So something we may want to revisit (separate from this PR).

compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
xylar: I added a check in `setup()` as requested and I set-up (but did not run) the SORRMr4 tests to make sure the new check passes.

compass/ocean/tests/global_ocean/mesh/remap_mali_topography/ais_4to20km.cfg
```diff
@@ -0,0 +1,15 @@
+# config options related to remapping topography to an MPAS-Ocean mesh
+[remap_topography]
+
+# the density of ocean water (kg/m^3), equivalent to config_ocean_density in MALI
+ocean_density = 1028.0
+
+# the elevation of sea level, equivalent to config_sea_level in MALI
+sea_level = 0.0
+
```
matthewhoffman, line 9: Did you mean to include `ice_density` here (and in the 2km cfg)?

xylar: I agree that would be more intuitive.  The config option is currently set here:
https://github.com/MPAS-Dev/compass/blob/559418ab7ebd4da5816a66de590558d9809c4d14/compass/ocean/mesh/remap_topography.cfg#L29
And that applies whether MALI topography is used or not.  But it makes sense to have both constants in the same place.

xylar: Done!

matthewhoffman (approved):
@xylar , no, I don't have any other suggestions.

xylar:
Great, thanks so much @matthewhoffman and @cbegeman!!!

PR #830, opened 2024-06-17
Mask only topography above sea level when remapping BedMachine topo 

xylar:
This merge updates the utility that creates the merged BedMachine and GEBCO topography dataset to include bathymetry and ice topography under grounded ice as long as that ice below sea level.  This is done to make the topography used in simulations without MALI coupling more similar to simulations where the topography will come from MALI in the near future.  Since grounding lines will move, we do not wish to mask the topography to only the floating domain.

However, including topography above sea level is problematic.  It can result in both bathymetry and ice-draft values that are above sea level even in cells where a significant fraction of the cell is below sea level (e.g. when bathymetry is shallow in the portion of the cell that is below sea level but steep in the portion that is above sea level).

Similarly, we will want to have MALI pass only the portion of the topography that is below sea level, along with a fraction of each cell that is below sea level.

Checklist

### DIFF
--- a/compass/ocean/tests/utility/combine_topo/__init__.py
+++ b/compass/ocean/tests/utility/combine_topo/__init__.py
@@ -170,13 +170,10 @@ class Combine(Step):
         grounded_mask = np.logical_or(np.logical_or(mask == 1, mask == 2),
                                       mask == 4).astype(float)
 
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

This has been successfully tested along with #807 in:
```
/lcrc/group/e3sm/ac.xylar/compass_1.4/chrysalis/test_20240611/sowiscr3_expand_rx1_mask/
```

PR #837, opened 2024-07-01
Fix files for E3SM following reorg of init

xylar:
Following https://github.com/MPAS-Dev/compass/pull/813, the initial state is the same whether ice shelf cavities are present or not.  It is the result of the `init/initial_state` step.

The `remap_ice_shelf_melt` step in `files_for_e3sm` needs to find the output in the corresponding `remap_ice_shelf_melt` in the `init` test case during `setup()` because that step no longer exists at construction.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/__init__.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/__init__.py
@@ -162,12 +162,8 @@ class FilesForE3SM(TestCase):
                        base_mesh_filename)
 
         if init is not None:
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I ran the IcoswISC240 mesh through `files_for_e3sm` on Chrysalis.  With the 2 fixes in this branch, it completed successfully, whereas without them it did not.
```
/lcrc/group/e3sm/ac.xylar/compass_1.4/chrysalis/test_20240701/icoswisc240_fix_files_for_e3sm
```

PR #838, opened 2024-07-03
Update to compass 1.4.0-alpha.4, numpy >=2.0, mpas_tools 0.34.1, matplotlib >=3.9.0

xylar:
Checklist
* [x] Document (in a comment titled `Testing` in this PR) any testing that was used to verify the changes

PR #839, opened 2024-07-04
Update the E3SM-Project submodule

xylar:
This merge updates the E3SM-Project submodule from [31f771c](https://github.com/E3SM-Project/E3SM/tree/31f771c) to [c7d7998](https://github.com/E3SM-Project/E3SM/tree/c7d7998).

This update includes the following MPAS-Ocean and MPAS-Frameworks PRs (check mark indicates bit-for-bit with previous PR in the list):
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/6472
- [x]  (ocn) https://github.com/E3SM-Project/E3SM/pull/6471
- [ ]  (ocn) https://github.com/E3SM-Project/E3SM/pull/6481 -- as expected, there are non-BFB differences in all global ocean forward runs due to the chantes in Omega and pi.  The changes are of order 1e-6 to 1e-8, so small but definitely not at machine percision


xylar:
It seems like I jumped the gun on `matplotlib` 3.9.0 in #838.  It appears to have been published erroneously and we're still waiting on an official release on conda-forge (which will actually be 3.9.1: https://github.com/conda-forge/matplotlib-feedstock/pull/390).

To make progress with this testing in the meantime, I temporarily allowed 3.8.4 in my local checkout and that also seemed to work fine.

PR #843, opened 2024-07-11
Only add metadata to forward runs when useful

xylar:
A lot of time is being taken to add metadata to output files during regression testing, even though that metadata will not be used by anyone and is not very useful.

We only want to be adding metadata to mesh files, initial conditions, outputs from dynamic adjustment and files staged for use as initial conditions in E3SM.

Checklist


xylar:
## Testing

I ran the `pr` suite with this change.  It is twice as fast as it was before this change (17 min instead of 34).

PR #844, opened 2024-07-18
Fix land- and land-ice-locked cells

xylar:
We need to address land-locked and land-ice-locked cells within the process for flood-filling to ensure that the land-ice mask is connected.  Otherwise, land-ice-locked cells added to the land-ice mask can create new regions of open ocean within the land ice.

Checklist

### DIFF
--- a/compass/ocean/mesh/cull.py
+++ b/compass/ocean/mesh/cull.py
@@ -1,6 +1,7 @@
 import os
 
 import mpas_tools.io
+import numpy as np
 import xarray as xr
 from geometric_features import (
     FeatureCollection,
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I created a new SORRM r3 (#807) land-ice mask during the `cull_mesh` step with this procedure.  Whereas before this fix there is a region of isolated open ocean surrounded by land or land ice near Stange Ice Shelf:
stange_orig
this is filled in with land-ice with this fix:
stange_fix

xylar:
My actual preferred fix is to add a critical passage to keep this location (called the Carroll Inlet) open, see https://github.com/MPAS-Dev/geometric_features/pull/203.  But the algorithm here is behaving consistently with what we want the *general* solution to these kinds of issues to be.

PR #847, opened 2024-07-24
Set more topographic variables in critical passages

xylar:
We need to set all land-ice related variables to zero in critical passages:

* `landIceFracObserved` (currently the case)
* `landIcePressureObserved`
* `landIceDraftObserved`
* `landIceGroundedFracObserved`
* `landIceFloatingFracObserved`
* `landIceThkObserved`

We also need to set `oceanFracObserved` to one.

Checklist


xylar:
## Testing

I tested this on Chrysalis along with #807 and examined the resulting `landIce*` variables and `oceanFracObserved` in ParaView.  (In the process, I discovered that `landIceThkObserved` had been missed, and has been subsequently added here.)

PR #848, opened 2024-07-30
Floodfill `landIceFraction`, not `landIceMask`

xylar:
During the `cull_mesh` step of creating a global ocean mesh, if ice-shelf cavities are present, we need to flood fill to make sure all land ice is connected (no islands of isolated land ice). This merge changes the flood fill to be on all locations where `landIceFraction > 0`, not all locations where `landIceFraction > 0.5`. It then masks all `landIce*` fields to only be nonzero within the flood-filled region.  This prevents islands of nonzero `landIceFraction`, `landIcePressure`, `landIceDraft`, etc.

Checklist

### DIFF
--- a/compass/ocean/mesh/cull.py
+++ b/compass/ocean/mesh/cull.py
@@ -488,7 +488,7 @@ def _cull_topo(with_cavities, process_count, logger, latitude_threshold,
     write_netcdf(ds_map_culled_to_base, 'map_culled_to_base.nc')
 
     if with_cavities:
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I tested this in conjunction with #807.  Without this fix, there were still regions where `landIceFraction` was nonzero around islands, which I think we probably don't want.  

Having nonzero `landIceFraciton` around islands was also caused confusion since @proteanplanet's sanity checking was using `landIceFraction` (rather than, e.g., `landIceMask` or `landIceFloatingFraction`) to determine the ice-shelf extent.

PR #850, opened 2024-08-15
Set ssh to land-ice draft only in meshes with cavities

xylar:
Without this fix, meshes without cavities were ending up with a ring of depressed sea surface height around Antarctica.

Checklist

### DIFF
--- a/compass/ocean/mesh/cull.py
+++ b/compass/ocean/mesh/cull.py
@@ -492,6 +492,8 @@ def _cull_topo(with_cavities, process_count, logger, latitude_threshold,
                                           ds_map_culled_to_base, ds_preserve,
                                           logger, latitude_threshold,
                                           sweep_count)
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I produced an Icos60 mesh with this branch.  As expected, the ssh is zero for the whole domain and the thickness of the top layer is 10 m everywhere.

xylar:
## More testing

I also ran the SORRM mesh through `performance_test` and it looked as expected -- the ssh is depressed around Antarctic just like `landIceDraft`.

xylar:
@mark-petersen, please review at your earliest convencience.

mark-petersen (approved):
@xylar thanks for fixing this! Approved by visual inspection and @xylar's testing.

xylar:
Thanks, @mark-petersen!

PR #852, opened 2024-08-20
Fix tolerance used for floating point comparison (Fixes #851) 

### LINKED ISSUE #851: Landice mismip+ mesh has incorrect boundary conditions at high resolution
For cell spacing ≤1km, the northern boundary of the mismip+ mesh has a `dirichletVelocityMask` multiple cells deep. Here is the issue on the 1km mesh:
image

and here is 500m:
image

The effect on the velocity field is clearly visible for high resolution runs.
Presumably that is happening in this block of code, but I am not sure why:
https://github.com/MPAS-Dev/compass/blob/main/compass/landice/tests/mismipplus/setup_mesh.py#L452-L463

andrewdnolan:
Per @trhille's experimenting, there are problems setting the Dirichlet BC's for resolutions $\leq$ 1000(m). This PR fixes #851 by using an absolute tolerance in all floating point comparisons of 5% of the appropriate cell spacing quantity.

### DIFF
--- a/compass/landice/tests/mismipplus/setup_mesh.py
+++ b/compass/landice/tests/mismipplus/setup_mesh.py
@@ -275,7 +275,7 @@ def mark_cull_cells_for_MISMIP(ds_mesh):
     # Get the y position of the top row
     y_max = ds_mesh.yCell.max()
 
-    # set the absolute tolerance to `dy` +/- a 5% buffer
+    # set the absolute tolerance to +/- 5% of `dy`
     atol = dy * 0.05
     # find the first interior row along the top of the domain
     mask = np.isclose(ds_mesh.yCell, y_max - dy, atol=atol, rtol=0)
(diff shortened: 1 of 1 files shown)


andrewdnolan:
# Testing: 

## Current implementation 

Here's the quantities of interest from the `HEAD` of `main`: 

image
As @trhille pointed out there are too many cells along the North boundary marked as Dirchlet, same problem applies to `areaCell`. `dvEdge` is a bit more complicated I'll explain that latter. 

## Using a 5% of cell spacing absolute tolerance

Here's the quantities of interest produced with this branch: 
image

There are two improvements: 
1. The culling is now correct in that the top most cell center corresponds to exactly to 80e3 (c.f. before). 
2. The masking of `dirichletVelocityMask` and `areaCelll` now only affect the top most cell.

andrewdnolan:
# Testing (Cont.)

So about the `dvEdge` field. Unfortunately the way I construct edge patches in [`mosaic` ](https://github.com/E3SM-Project/mosaic) is too simple to deal with these slightly complicated boundary edges. So, for now I've got to go to paraview (for now).  

Here's an example of the same `dvEdge` field for a `1000m` mesh, where the edges above the boundary (80e3) have been set to zero and the edges __at__ the boundary have been set to half their original distance. Which is needed to match the half area of the cells along the boundary (see above). 

image

trhille (approved):
Thanks @andrewdnolan! I tested for 1000m and 500m mesh creation and it now works as expected.

PR #854, opened 2024-08-28
Fix masking of ice draft and ice thickness in combine_topo

xylar:
Checklist


xylar:
# Testing

I ran this along with #855 and #807, and now see sensible masking of ice thickness (which is also bounded to be below the flotation thickness because of #855):

limited_thickness

PR #855, opened 2024-08-28
Fix ice thickness and pressure where above flotation

xylar:
We don't want ice thickness and pressure to exceed their flotation
values.  Previously, we were only making sure that ice draft did
not exceed its flotation value (i.e. the draft was above the bed).

Checklist


xylar:
## Testing

Tested with #854 and #807, see https://github.com/MPAS-Dev/compass/pull/854#issuecomment-2317646912

PR #856, opened 2024-09-04
Update supported python versions

altheaden:
Python version 3.8 has now been dropped from conda-forge, and 3.12 has been added. This PR just updates the related checks and files to be consistent with those supported versions.

Checklist
* [x] Developer's Guide has been updatedbuilding_docs.html) and changes look as expected


altheaden:
@xylar as you will see in the changed files, I deleted and added some files in the `ci` folder. I just guessed based on the existing files that the only change within these files would be the version number, but you should definitely look that over in case it should be different. Also, please take a look at my change in the developer's guide, it's very small but I want to make sure that it's accurate.

altheaden:
## Testing
I haven't done any testing yet, as I'm not sure if any of these changes would affect building/running compass. I am happy to run the PR suite if that's desired, but I'll hold off for now.

docs/developers_guide/quick_start.rst
```diff
@@ -217,7 +217,7 @@ Optional flags
     Check to make sure expected commands are present
 
 ``--python``
-    Select a particular python version (the default is currently 3.8)
+    Select a particular python version (the default is currently 3.9)
```
xylar, line 220:
```suggestion
    Select a particular python version (the default is currently 3.12)
```

xylar (commented):
Wonderful!  One tiny change.

docs/developers_guide/quick_start.rst
xylar: This was *very* out of date.

xylar:
@altheaden, these could in theory affect tests but I'm not worried.  As long as CI passes, I think we can consider that sufficient testing.

altheaden:
@xylar Done!

xylar:
Thanks @altheaden!

PR #858, opened 2024-09-11
Tweak behaviors of CI workflows

altheaden:
In this PR I made a few tweaks to the behavior of the CI workflows. These changes are:
1. `cancel_others` is now set to false by default, which should make it easier to decipher if a bug is isolated to a specific version of python via CI
2. The `defaults` channel has been removed as we can now exclusively use `conda-forge`
3. The timeout for the `pre-commit` job has been upped to 5 minutes

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -10,14 +10,14 @@ on:
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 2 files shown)


xylar (approved):
@altheaden, again, this looks great!

PR #860, opened 2024-09-23
Add additional flags to `pip install` command

altheaden:
In this PR I added a few flags to the `python -m pip install [-e] .` command. The flags are:
* `-vv` indicates very verbose output. This flag is used during the CI runs to help with debugging.
* `--no-deps` indicates that pip should not install runtime dependencies from PyPi.
* `--no-build-isolation` indicates that pip should not install build time dependencies from PyPi.

I also had to make several formatting changes to existing code in order to satisfy the linter.

Checklist

### DIFF
--- a/conda/bootstrap.py
+++ b/conda/bootstrap.py
@@ -312,8 +312,9 @@ def build_conda_env(env_type, recreate, mpi, conda_mpi, version,
             # conda packages don't like dashes
             version_conda = version.replace('-', '')
             packages = f'{packages} "compass={version_conda}={mpi_prefix}_*"'
-            commands = f'{activate_base} && ' \
-                       f'conda create -y -n {env_name} {channels} {packages}'
+            commands = \
+                f'{activate_base} && ' \
+                f'conda create -y -n {env_name} {channels} {packages}'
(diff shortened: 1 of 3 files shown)


xylar:
This looks good!  I feel like the changes are significant enough that I'd better do a quick test, though.

altheaden:
Sounds good, if you would like me to test it instead (or in addition) I would be happy to do so as well.

xylar:
## Testing

I ran:
```
./conda/configure_compass_env.py --conda ~/chrysalis/miniforge3/ --verbose --recreate --env_name compass_test
```
As expected, I saw the new `--no-deps` and `--no-build-isolation` flags where we want them.

Then, just to make sure `compass` was installed properly, I ran:
```
source load_compass_test_chrysalis_intel_openmpi.sh
compass list
```
and saw the expected list of packages.

xylar (approved):
@altheaden, excellent work as always.  Thanks for taking care of the nasty linting!

xylar:
> if you would like me to test it instead (or in addition) I would be happy to do so as well.

I don't think that's necessary.  Just wanted to catch anything unexpected.

PR #861, opened 2024-09-25
Added `cubed_sphere` target grid test to `utility/combine_topo`

bmooremaley:
The `ocean/utility/combine_topo` test has been split into two separate tests: one for a `lat_lon` target grid and one for a `cubed_sphere` target grid. The `cubed_sphere` test allows the migration to `mbtempest` for remapping bathymetry to MPAS grids by removing the pole singularities. The workflow for the existing `lat_lon` target grid has also been updated to use remapping tools more consistently between the GEBCO and BedMachineAntarctica data sets, and between the `lat_lon` and `cubed_sphere` target mesh workflows. Specifically, the GEBCO dataset is separated into tiles prior to remapping, `ESMF_RegridWeightGen` is used for building weights files, and `ncremap` is used for the final remapping step.

Checklist

### DIFF
--- a/compass/ocean/tests/utility/combine_topo/__init__.py
+++ b/compass/ocean/tests/utility/combine_topo/__init__.py
@@ -3,10 +3,10 @@ import subprocess
 from datetime import datetime
 
 import numpy as np
-import progressbar
 import pyproj
 import xarray as xr
-from pyremap import LatLonGridDescriptor, ProjectionGridDescriptor, Remapper
+from dask.diagnostics import ProgressBar
+from pyremap import ProjectionGridDescriptor
 
 from compass.step import Step
 from compass.testcase import TestCase
@@ -57,13 +57,14 @@ class Combine(Step):
         test_case : compass.ocean.tests.utility.combine_topo.CombineTopo
             The test case this step belongs to
         target_grid : `str`, either "lat_lon" or "cubed_sphere"
-        date_stamp : `datetime.datetime`, date when test was run
         """
         super().__init__(
             test_case, name='combine', ntasks=None, min_tasks=None,
         )
         self.target_grid = target_grid
-        self.date_stamp = None
+        self.resolution = None
+        self.resolution_name = None
+
 
     def setup(self):
         """
@@ -78,16 +79,20 @@ class Combine(Step):
         # Get input filenames and resolution
         antarctic_filename = section.get('antarctic_filename')
         global_filename = section.get('global_filename')
-        resolution = section.getint('resolution')
 
-        # Datestamp
-        self.datestamp = datetime.now().strftime('%Y%m%d')
+        # Parse resolution
+        if self.target_grid == 'cubed_sphere':
+            self.resolution = section.getint('resolution')
+            self.resolution_name = f'ne{self.resolution}'
+        elif self.target_grid == 'lat_lon':
+            self.resolution = section.getfloat('resolution')
+            self.resolution_name = f'{self.resolution:.4f}_degree'
 
-        # Build output filename
-        combined_filename = _get_combined_filename(
-            self.target_grid, self.datestamp, resolution,
-            antarctic_filename, global_filename,
-        )
+        # Build combined filename
+        combined_filename = '_'.join([
+            antarctic_filename.strip('.nc'), global_filename.strip('.nc'),
+            self.resolution_name, datetime.now().strftime('%Y%m%d.nc'),
+        ])
 
         # Add bathymetry data input files
         self.add_input_file(
@@ -102,12 +107,10 @@ class Combine(Step):
         )
         self.add_output_file(filename=combined_filename)
 
-        # Make tiles directory
-        os.makedirs('tiles', exist_ok=True)
-
         # Get ntasks and min_tasks
         self.ntasks = section.getint('ntasks')
         self.min_tasks = section.getint('min_tasks')
+
 
     def constrain_resources(self, available_resources):
         """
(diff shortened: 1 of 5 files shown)


compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -158,169 +176,411 @@ def _modify_bedmachine(self):
...
+        # Build cubed sphere SCRIP file using tempestremap
+        if self.target_grid == 'cubed_sphere':
+
+            # Create EXODUS file
+            args = [
+                'GenerateCSMesh', '--alt', '--res', f'{self.resolution}',
+                '--file', f'{self.resolution_name}.g',
+            ]
+            logger.info(f'    Running: {" ".join(args)}')
+            subprocess.run(args, check=True)
```
xylar, line 321: In compass, there's a strong preference to call `mpas_tools.logging.check_call()` and pass self.logger.  If you do this, the output from these calls goes to the log file, not to the "terminal" or the log for the job script.  We try to keep that output really minimal so it's obvious if the task succeeded and how long it took.  The same goes for all the `subprocess.run()` calls.

However, I know you were seeing hanging so we will have to see if that recurs with calls to `check_call()`.

compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -158,169 +176,411 @@ def _modify_bedmachine(self):
...
+        """
+        logger = self.logger
 
-        out_descriptor = LatLonGridDescriptor.read(fileName=gebco_filename)
+        config = self.config
+        method = config.get('combine_topo', 'method')
+
+        # Generate weights file
+        args = [
+            'srun', '-n', f'{self.ntasks}',
```
xylar, line 358: We want to let compass take care of this, rather than hard-coding `srun` and its flags.  The preferred way to do this it to call `compass.parallal.run_command()` something like the way it's done for the model:
https://github.com/MPAS-Dev/compass/blob/2e06f2e0452cc7dd410bc47a5374f90ed34ea169/compass/model.py#L60

xylar (commented):
@bmooremaley, this is excellent work!  I'm very happy with the way you have reorganized things and the clean-up you've done along the way.

I have some suggested changes that we can discuss today.

xylar:
Presumably, you'll need to do a commit to make the linter happy.  The easiest way to do that is probably to add an extra newline somewhere (maybe at the end of the file) and then fix whatever it complains about.  Make sure you do that with the conda environment loaded so pre-commit can happen.

compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -55,101 +77,97 @@ def setup(self):
...
+            self.resolution_name = f'ne{self.resolution}'
+        elif self.target_grid == 'lat_lon':
+            self.resolution = section.getfloat('resolution_latlon')
+            self.resolution_name = f'{self.resolution:.4f}_degree'
+
+        # Build combined filename
+        combined_filename = '_'.join([
+            antarctic_filename.strip('.nc'), global_filename.strip('.nc'),
+            self.resolution_name, datetime.now().strftime('%Y%m%d.nc'),
+        ])
```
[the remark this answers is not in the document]
bmooremaley: Thanks!

compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -158,169 +180,410 @@ def _modify_bedmachine(self):
...
+                '--out', out_filename,
+            ]
+            check_call(args, logger)
+
+        # Build lat-lon SCRIP file using pyremap
+        elif self.target_grid == 'lat_lon':
+            descriptor = get_lat_lon_descriptor(
+                dLon=self.resolution, dLat=self.resolution,
+            )
+            descriptor.to_scrip(out_filename)
```
xylar, line 337: Need to make sure this is in NETCDF3 format.  This should be possible by setting the `.format` argument in the descriptor (see pyremap for details).

compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -14,37 +19,54 @@ class CombineTopo(TestCase):
...
 class Combine(Step):
     """
     A step for combining GEBCO 2023 with BedMachineAntarctica topography
     datasets
+
+    Attributes
+    ----------
+    target_grid : `str`, either "lat_lon" or "cubed_sphere"
+    resolution : `float` degrees, or `int` NExxx
+    resolution_name: `str`, either x.xxxx_degrees or NExxx
```
xylar, line 51: To parse properly in the docs, these need to be formatted as:
```suggestion
    target_grid : str
        either "lat_lon" or "cubed_sphere"
    resolution : float or int
        degrees (float) or face subdivisions (int): NExxx
    resolution_name: str
        either x.xxxx_degrees or NExxx
```
Similarly for other docstrings below.

compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -55,101 +77,101 @@ def setup(self):
...
-        cobined_filename = section.get('cobined_filename')
-        self.add_output_file(filename=cobined_filename)
 
+        # Parse resolution and assign resolution attributes
+        if self.target_grid == 'cubed_sphere':
+            self.resolution = section.getint('resolution_cubedsphere')
+            self.resolution_name = f'ne{self.resolution}'
+        elif self.target_grid == 'lat_lon':
+            self.resolution = section.getfloat('resolution_latlon')
+            self.resolution_name = f'{self.resolution:.4f}_degree'
```
xylar, line 91: Unfortunately, I discovered that this didn't have the effect I was expecting.  I was assuming I could set up the test case and then change `resolution_cubedsphere = 30` and see the effects, but instead I still got an ne3000 dataset.  The reason for this is that, although `resolution_cubedsphere` was changed, it was not read back in a runtime and the corresponding `resolution_name` and output filenames were not set.

This isn't too hard to fix.  My suggested solution would be https://github.com/xylar/compass/commit/c68ffce6c8b8b219e6551518664ab1f7b61ce813.  You should just be able to cherry-pick that commit onto your branch if you add my for as a remote.

I'm in the process of testing this out but it does seem to work as intended.  There is no need for a user config file in this case, you just modify the `combine_topo.cfg` file before running the test case.

xylar (changes requested):
@bmooremaley, this looks good and seems to work for the most part.  I did run into an unexpected behavior that I describe below and provide a suggested fix for.  I also noticed that the docstrings need some small tweaks in formatting.  We use the numpydoc format so this may be a little different than what you've done in the past.

compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -55,101 +77,101 @@ def setup(self):
...
+
+        # Build output filenames
+        datestamp = datetime.now().strftime('%Y%m%d')
+        scrip_filename = f'{self.resolution_name}_{datestamp}.scrip.nc'
+        combined_filename = '_'.join([
+            antarctic_filename.strip('.nc'),
+            global_filename.strip('.nc'),
```
xylar, line 98: For consistency with existing filenames:
```suggestion
            antarctic_filename.strip('.nc'),
            'and',
            global_filename.strip('.nc'),
```

xylar: Hmm, I see that this still isn't consistent with other filename because we also converted dashes to underscores, so probably not worth the trouble.

compass/ocean/tests/utility/combine_topo/__init__.py
```diff
@@ -158,169 +180,410 @@ def _modify_bedmachine(self):
...
+        # Parse config
+        config = self.config
+        section = config['combine_topo']
+        global_name = section.get('global_filename').strip('.nc')
+        method = section.get('method')
+        lat_tiles = section.getint('lat_tiles')
+        lon_tiles = section.getint('lon_tiles')
+
+        # Make tiles directory
+        os.mkdir('tiles')
```
xylar, line 436: I think we want:
```suggestion
        os.makedirs('tiles', exist_ok=True)
```
This works if `tiles` already exists.  The alternative is that we want to delete `tiles` if it exists and start fresh.

xylar: I don't think it matters which, since everything gets rerun either way.

compass/ocean/tests/utility/combine_topo/__init__.py
bmooremaley: See 7c98497d53171b8c17587e14090640ca62e564e6

compass/ocean/tests/utility/combine_topo/__init__.py
bmooremaley: Happy with your changes, I did a hard reset to your fork.

compass/ocean/tests/utility/combine_topo/__init__.py
bmooremaley: Happy with your changes, I did a hard reset to your fork.

compass/ocean/tests/utility/combine_topo/__init__.py
bmooremaley: Coming back to this comment. At the moment, I believe that all files are NETCDF4 until the `mbtempest` step in `remap_topography`. The conversion happens there, and my understanding is that `mbtempest` is the only part of the workflow that requires NETCDF3, correct? We can do the conversion here instead, but we would probably want to apply it both to the lat-lon scripfile produced by `pyremap` and the cubed-sphere scripfile produced by `tempestremap`, correct?

xylar: We could set `descriptor.format = NETCDF3_64BIT_DATA` similar to what I do here with `NETCDF3_64BIT`:
https://github.com/MPAS-Dev/compass/blob/5974cb476187e531555097621b40a1bf854cf414/compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py#L83
That would be my preferred way.

xylar: >  cubed-sphere scripfile produced by tempestremap, correct?

I think those files are pretty unlikely to be in NETCDF4 format but we would need to check.

bmooremaley: They are not, I checked when developing `remap_topography`. I was thinking to convert using `ncks -5` here for consistency and then remove the conversion in `remap_topography`.

xylar: Yes, my preference would be to get rid of the conversion in `remap_topography` and just make sure they're in the right format here.  We'll want to delete the old scrip files (before this fix) that we added to the database to avoid confusion.

xylar (approved):
As soon as #879 goes in, I will rebase this an merge it.

Thanks for the very hard work on this, @bmooremaley!!

xylar:
A bit more debugging on how to handle `NETCDF3_64BIT_DATA` but I think I've got it working now.  I have the ne120 data done and the others in progress.

PR #863, opened 2024-10-07
Modified `ocean/mesh/remap_topography` to allow smoothing

bmooremaley:
The `ocean/mesh/remap_topography` step has been modified to remap combined topography from a cubed sphere source grid to MPAS meshes, instead of a lat-lon source grid. The weights file generation has been switched to `mbtempest` so that smoothing can be defined in the target MPAS mesh SCRIP file. Smoothing parameters have been added to `ocean/mesh/remap_topography.cfg`. This new workflow also uses a pre-build source SCRIP file stored in the bathymetry database, in addition to the combined topography file.

Checklist


compass/ocean/mesh/remap_topography.py
```diff
@@ -93,65 +107,151 @@ def run(self):
...
+
+        descriptor = MpasCellMeshDescriptor(
+            fileName='base_mesh.nc',
+            meshName=self.mesh_name,
+        )
+        descriptor.to_scrip(
+            'target.scrip.nc',
+            expandDist=expandDist,
+            expandFactor=expandFactor,
+        )
```
xylar, line 137: Before this call, we want to make sure to set `descriptor.format = 'NETCDF3_64BITDATA'` (Check this, it's just my best guess on the names of the attribute and its value.)

xylar:
@bmooremaley, as we discussed, we'll need some way to support a different topography dataset for lower res regression testing.  I pulled out the config options into their own config file:
https://github.com/MPAS-Dev/compass/pull/864
If we want to continue to support that 1/20 degree dataset, we would need a way to still use ESMF for that resolution.  An alternative (if mbtempest if fast enough) is to use a lower resolution cubed-sphere dataset as we discussed and stick with the modified workflow you have here.  I would lean slightly toward the latter so that the regression testing is using a more similar workflow to the high resolution runs.

xylar:
I'm moving 5de71c4 to #879.  This branch should be rebased once that PR gets merged.

xylar (commented):
@bmooremaley, some preliminary comments and suggested changes while you work on the documentation.

compass/ocean/mesh/remap_topography.cfg
```diff
@@ -2,7 +2,14 @@
 [remap_topography]
 
 # the name of the topography file in the bathymetry database
-topo_filename = BedMachineAntarctica_v3_and_GEBCO_2023_0.0125_degree_20240828.nc
+#topo_filename = BedMachineAntarctica_v3_and_GEBCO_2023_ne3000_20241004.nc
+#src_scrip_filename = ne3000_20241004.scrip.nc
```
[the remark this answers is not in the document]
xylar: I tried both the ne30 and ne120 topography for Icos240 and IcoswISC240.  I saw crashes in forward runs in both cases.  This might just mean we need to lower the time step but it might indicate problems.  I'm just noting this for now so I don't forget to fix it, since we need these tests to work before this PR can be merged.

compass/ocean/mesh/remap_topography.py
```diff
@@ -94,64 +108,203 @@ def run(self):
...
+        config = self.config
+        method = config.get('remap_topography', 'method')
 
-        remapper.build_mapping_file(method=method, mpiTasks=self.ntasks,
-                                    tempdir='.', logger=logger,
-                                    esmf_parallel_exec=parallel_executable)
+        # Build command args
+        # Unused options:
+        # -P mpas, handles some MPAS-specific index ordering, CF, etc...
+        # -C climatology, basically bypasses fill values
```
xylar, line 252: Let's remove these comments.  They were probably helpful for your learning process but aren't going to be useful to the average developer.

compass/ocean/mesh/remap_topography.py
```diff
@@ -94,64 +108,203 @@ def run(self):
...
+        """
         logger = self.logger
-        parallel_executable = config.get('parallel', 'parallel_executable')
+        logger.info('Partition SCRIP file')
+
+        # Convert to NetCDF3 64-bit
+        args = [
+            'ncks', '-5',
+            in_filename,
+            in_filename.replace('.nc', '.64bit.nc'),
```
xylar, line 161: I'm not particularly fond of these `relpace()` calls.  Could we instead make all the filenames at the top of this method?  Could we extract the `stem` and then use f-strings to make the filenames, e.g.:
```
        stem = pathlib.Path(in_filename).stem
        64bit_filename = f'{stem}.64bit.nc'
        h5m_filename = f'{stem}.64bit.h5m'
        part_filename = f'{stem}.64bit.p{self.ntasks}.h5m'
```
It's unclear to me if the `.64bit` suffix is meaningful for the `.h5m` files and I'd suggest removing it.

xylar (commented):
@bmooremaley, a few more suggested edits.  I'm extending this code in #829 to combine this topography with a dataset from MALI.  That process is qualitatively similar to your modified `remap_topography` so it was a useful chance to look through your code in more detail.

docs/users_guide/ocean/test_groups/global_ocean.rst
```diff
@@ -476,7 +476,7 @@ The mesh has 12-km resolution around Antarctica and 30-km resolution elsewhere.
 The mesh includes the :ref:`global_ocean_ice_shelf_cavities` around Antarctica
 in the ocean domain.
 
-.. image:: images/sowisc12to30.png
+.. image:: images/sowisc12to60.png
```
xylar, line 479: Unrelated to this PR but an error in the docs build that should be fixed.  (We would ideally plot the update mesh.)

xylar:
I think the only remaining steps are:
* run all the global ocean tests from the `pr` suite but without cached meshes and initial conditions
* update the cached mesh and initial condition files
* make sure the `pr` suite succeeds

xylar:
## Testing

I tested this on Chrysalis with Intel and OpenMPI.  I ran the `Icos240`, `IcoswISC240`, `Icos` and `IcoswISC` meshes through the `mesh`, `init` and `performance_test` test cases without caching.  Then, I cached `mesh` and `init` steps.  Then, I successfully ran the `pr` suite (which uses the new cached files).  I looked at the relevant topography fields from the `initial_state` step for each mesh and everything looked great to me.

xylar (approved):
@bmooremaley, thanks for very hard work on this!  I took care of the last few finishing touches so I think it's ready to merge!

PR #864, opened 2024-10-09
Put low res topography options in a config file

xylar:
This is much less confusing than having them hard-coded deep inside of global ocean.

Checklist


xylar:
## Testing

I successfully ran the `pr` suite on Chrysalis with Intel and OpenMPI.

I verified that the config options in `mesh.cfg` for the `Icos` test case looked like they should (i.e. the low res topography data set, not the high res).

PR #870, opened 2024-10-29
Fix filename produced by extrap_woa test case

xylar:
The filename now indicates that it is for January (`_jan`) and includes a datestamp.

Checklist


xylar:
## Testing

I ran the test on Chrysalis.  As expected, it produced a file `extrap/woa23_decav_0.25_jan_extrap.20241029.nc` that we could add to the database but won't in this case.

xylar:
@vanroekel, do you want to do a quick review of this just by inspection?  That should be sufficient, since I tested the file creation and I don't think there's any chance of a mistake in the `global_ocean` change.

xylar:
Happy for suggestions if the `_jan` suffix seems insufficient.

vanroekel (approved):
Thanks for the quick fix @xylar this looks great and the `_jan` addition is enough for me.

xylar:
Thanks @vanroekel!

PR #872, opened 2024-10-30
Adds WOA 23 SSS restoring file script

vanroekel:
This adds WOA23 surface salinity restoring for E3SM G-cases.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
@@ -29,11 +29,11 @@ class RemapSeaSurfaceSalinityRestoring(FilesForE3SMStep):
                          ntasks=512, min_tasks=1)
 
         self.add_input_file(
-            filename='PHC2_salx.2004_08_03.filled_double_precision.nc',
-            target='PHC2_salx.2004_08_03.filled_double_precision.nc',
+            filename='woa_decav_0.25_sss_monthly_extrap.20241031.nc',
+            target='woa_decav_0.25_sss_monthly_extrap.20241031.nc',
             database='initial_condition_database')
 
-        self.add_output_file(filename='sss.PHC2_monthlyClimatology.nc')
+        self.add_output_file(filename='sss.WOA23_monthlyClimatology.nc')
 
     def run(self):
         """
@@ -44,9 +44,9 @@ class RemapSeaSurfaceSalinityRestoring(FilesForE3SMStep):
         config = self.config
         ntasks = self.ntasks
 
-        in_filename = 'PHC2_salx.2004_08_03.filled_double_precision.nc'
+        in_filename = 'WOA23_monthly_sfc_salin.20241030.nc'
 
-        prefix = 'sss.PHC2_monthlyClimatology'
+        prefix = 'sss.WOA23_monthlyClimatology'
         suffix = f'{self.mesh_short_name}.{self.creation_date}'
 
         remapped_filename = f'{prefix}.nc'
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
@@ -29,8 +29,8 @@ class RemapSeaSurfaceSalinityRestoring(FilesForE3SMStep):
                          ntasks=512, min_tasks=1)
 
         self.add_input_file(
-            filename='woa23_decav_0.25_sss_monthly_extrap.20241031.nc',
-            target='woa23_decav_0.25_sss_monthly_extrap.20241031.nc',
+            filename='woa23_decav_0.25_sss_monthly_extrap.20241101.nc',
+            target='woa23_decav_0.25_sss_monthly_extrap.20241101.nc',
             database='initial_condition_database')
 
         self.add_output_file(filename='sss.WOA23_monthlyClimatology.nc')
@@ -44,7 +44,7 @@ class RemapSeaSurfaceSalinityRestoring(FilesForE3SMStep):
         config = self.config
         ntasks = self.ntasks
 
-        in_filename = 'WOA23_monthly_sfc_salin.20241030.nc'
+        in_filename = 'woa23_decav_0.25_sss_monthly_extrap.20241101.nc'
 
         prefix = 'sss.WOA23_monthlyClimatology'
         suffix = f'{self.mesh_short_name}.{self.creation_date}'
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
@@ -29,8 +29,8 @@ class RemapSeaSurfaceSalinityRestoring(FilesForE3SMStep):
                          ntasks=512, min_tasks=1)
 
         self.add_input_file(
-            filename='woa_decav_0.25_sss_monthly_extrap.20241031.nc',
-            target='woa_decav_0.25_sss_monthly_extrap.20241031.nc',
+            filename='woa23_decav_0.25_sss_monthly_extrap.20241031.nc',
+            target='woa23_decav_0.25_sss_monthly_extrap.20241031.nc',
             database='initial_condition_database')
 
         self.add_output_file(filename='sss.WOA23_monthlyClimatology.nc')
--- a/compass/ocean/tests/utility/__init__.py
+++ b/compass/ocean/tests/utility/__init__.py
@@ -1,6 +1,7 @@
 from compass.ocean.tests.utility.combine_topo import CombineTopo
 from compass.ocean.tests.utility.cull_restarts import CullRestarts
 from compass.ocean.tests.utility.extrap_woa import ExtrapWoa
+from compass.ocean.tests.utility.create_salinity_restoring import CreateSalinRestoring
 from compass.testgroup import TestGroup
 
 
@@ -19,3 +20,4 @@ class Utility(TestGroup):
         self.add_test_case(CombineTopo(test_group=self))
         self.add_test_case(CullRestarts(test_group=self))
         self.add_test_case(ExtrapWoa(test_group=self))
+        self.add_test_case(CreateSalinRestoring(test_group=self))
--- a/compass/ocean/tests/utility/__init__.py
+++ b/compass/ocean/tests/utility/__init__.py
@@ -1,7 +1,9 @@
 from compass.ocean.tests.utility.combine_topo import CombineTopo
+from compass.ocean.tests.utility.create_salin_restoring import (
+    CreateSalinRestoring,
+)
 from compass.ocean.tests.utility.cull_restarts import CullRestarts
 from compass.ocean.tests.utility.extrap_woa import ExtrapWoa
-from compass.ocean.tests.utility.create_salinity_restoring import CreateSalinRestoring
 from compass.testgroup import TestGroup
 
 
--- a/compass/ocean/tests/utility/create_salin_restoring/__init__.py
+++ b/compass/ocean/tests/utility/create_salin_restoring/__init__.py
@@ -1,11 +1,15 @@
-from compass.ocean.tests.utility.create_salinity_restoring.salinity_restoring import Salinity
-from compass.ocean.tests.utility.create_salinity_restoring.extrap_salin import ExtrapSalin
+from compass.ocean.tests.utility.create_salin_restoring.extrap_salin import (
+    ExtrapSalin,
+)
+from compass.ocean.tests.utility.create_salin_restoring.salinity_restoring import (  # noqa: E501
+    Salinity,
+)
 from compass.testcase import TestCase
 
 
 class CreateSalinRestoring(TestCase):
     """
-    A test case for first creating monthly sea surface salinity from WOA23 
+    A test case for first creating monthly sea surface salinity from WOA23
     then extrapolating the WOA23 data into missing ocean regions such as
     ice-shelf cavities and coasts
     """
@@ -19,7 +23,7 @@ class CreateSalinRestoring(TestCase):
         test_group : compass.ocean.tests.utility.Utility
             The test group that this test case belongs to
         """
-        super().__init__(test_group=test_group, name='create_salinity_restoring')
+        super().__init__(test_group=test_group, name='create_salin_restoring')
 
         self.add_step(Salinity(test_case=self))
         self.add_step(ExtrapSalin(test_case=self))
--- a/compass/ocean/tests/utility/create_salin_restoring/extrap_salin.py
+++ b/compass/ocean/tests/utility/create_salin_restoring/extrap_salin.py
@@ -47,8 +47,8 @@ class ExtrapSalin(Step):
 
         datestring = now.strftime("%Y%m%d")
 
-        self.woa_filename = f'woa23_decav_0.25_sss_monthly_extrap.\
-                              {datestring}.nc'
+        self.woa_filename = f'woa23_decav_0.25_sss_monthly_extrap.' \
+                            f'{datestring}.nc'
         self.add_output_file(self.woa_filename)
 
     def run(self):
--- a/compass/ocean/tests/utility/create_salin_restoring/extrap_salin.py
+++ b/compass/ocean/tests/utility/create_salin_restoring/extrap_salin.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 import numpy as np
 import xarray as xr
 from scipy.signal import convolve2d
@@ -7,8 +9,8 @@ from compass.step import Step
 
 class ExtrapSalin(Step):
     """
-    Extrapolate WOA 2023 monthly sea surface salinity data into missing ocean 
-	regions, including ice cavities and coasts
+    Extrapolate WOA 2023 monthly sea surface salinity data into missing ocean
+    regions, including ice cavities and coasts
 
     Attributes
     ----------
@@ -22,7 +24,8 @@ class ExtrapSalin(Step):
 
         Parameters
         ----------
-        test_case : compass.ocean.tests.utility.extrap_woa.ExtrapWoa
+        test_case : compass.ocean.tests.utility.create_salin_restoring.
+        CreateSalinRestoring
             The test case this step belongs to
 
         """
@@ -39,7 +42,13 @@ class ExtrapSalin(Step):
         """
         Determine the output filename
         """
-        self.woa_filename = 'woa_surface_salinity_monthly_extrap.nc'
+
+        now = datetime.now()
+
+        datestring = now.strftime("%Y%m%d")
+
+        self.woa_filename = f'woa23_decav_0.25_sss_monthly_extrap.\
+                              {datestring}.nc'
         self.add_output_file(self.woa_filename)
 
     def run(self):
@@ -48,9 +57,10 @@ class ExtrapSalin(Step):
         cavities.
         """
         # extrapolate horizontally using the ocean mask
-        _extrap_level(self.woa_filename)
+        _extrap(self.woa_filename)
 
-def _extrap_level(out_filename):
+
+def _extrap(out_filename):
 
     in_filename = 'woa_surface_salinity_monthly.nc'
     ds = xr.open_dataset(in_filename)
@@ -69,12 +79,12 @@ def _extrap_level(out_filename):
     lon_no_halo = list(range(2, nlon + 2))
 
     for i in range(12):
-        valid = np.isfinite(field[i,:,:])
+        valid = np.isfinite(field[i, :, :])
         orig_mask = valid
         prev_fill_count = 0
         while True:
             valid_weight_sum = _extrap_with_halo(valid, kernel, valid,
-                                               lon_with_halo, lon_no_halo)
+                                                 lon_with_halo, lon_no_halo)
 
             new_valid = valid_weight_sum > threshold
 
@@ -87,9 +97,10 @@ def _extrap_level(out_filename):
                 # no change so we're done
                 break
 
-            field_extrap = _extrap_with_halo(field[i,:,:], kernel, valid,
-                                           lon_with_halo, lon_no_halo)
-            field[i,fill_mask] = field_extrap[fill_mask]/valid_weight_sum[fill_mask]
+            field_extrap = _extrap_with_halo(field[i, :, :], kernel, valid,
+                                             lon_with_halo, lon_no_halo)
+            field[i, fill_mask] = field_extrap[fill_mask] / \
+                valid_weight_sum[fill_mask]
 
             valid = new_valid
             prev_fill_count = fill_count
@@ -100,6 +111,7 @@ def _extrap_level(out_filename):
     ds['SALT'].attrs = attrs
 
     ds.to_netcdf(out_filename)
+
 
 def _extrap_with_halo(field, kernel, valid, lon_with_halo, lon_no_halo):
     field = field.copy()
--- a/compass/ocean/tests/utility/create_salinity_restoring/__init__.py
+++ b/compass/ocean/tests/utility/create_salinity_restoring/__init__.py
@@ -1,0 +1,25 @@
+from compass.ocean.tests.utility.create_salinity_restoring.salinity_restoring import Salinity
+from compass.ocean.tests.utility.create_salinity_restoring.extrap_salin import ExtrapSalin
+from compass.testcase import TestCase
+
+
+class CreateSalinRestoring(TestCase):
+    """
+    A test case for first creating monthly sea surface salinity from WOA23 
+    then extrapolating the WOA23 data into missing ocean regions such as
+    ice-shelf cavities and coasts
+    """
+
+    def __init__(self, test_group):
+        """
+        Create the test case
+
+        Parameters
+        ----------
+        test_group : compass.ocean.tests.utility.Utility
+            The test group that this test case belongs to
+        """
+        super().__init__(test_group=test_group, name='create_salinity_restoring')
+
+        self.add_step(Salinity(test_case=self))
+        self.add_step(ExtrapSalin(test_case=self))
--- a/compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
+++ b/compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
@@ -1,0 +1,113 @@
+import os
+from functools import partial
+
+import numpy as np
+import xarray as xr
+from scipy.signal import convolve2d
+
+from compass.step import Step
+
+
+class ExtrapSalin(Step):
+    """
+    Extrapolate WOA 2023 monthly sea surface salinity data into missing ocean 
+	regions, including ice cavities and coasts
+
+    Attributes
+    ----------
+    woa_filename : str
+        The name of the output file name after extrapolation
+
+    """
+    def __init__(self, test_case):
+        """
+        Create a new test case
+
+        Parameters
+        ----------
+        test_case : compass.ocean.tests.utility.extrap_woa.ExtrapWoa
+            The test case this step belongs to
+
+        """
+        super().__init__(test_case=test_case, name='extrap', cpus_per_task=64,
+                         min_cpus_per_task=1, openmp_threads=1)
+
+        self.add_input_file(
+            filename='woa_surface_salinity_monthly.nc',
+            target='../salinity_restoring/woa_surface_salinity_monthly.nc')
+
+        self.woa_filename = None
+
+    def setup(self):
+        """
+        Determine the output filename
+        """
+        self.woa_filename = 'woa_surface_salinity_monthly_extrap.nc'
+        self.add_output_file(self.woa_filename)
+
+    def run(self):
+        """
+        Extrapolate WOA 2023 model temperature and salinity into ice-shelf
+        cavities.
+        """
+        # extrapolate horizontally using the ocean mask
+        _extrap_level(self.woa_filename)
+
+def _extrap_level(out_filename):
+
+    in_filename = 'woa_surface_salinity_monthly.nc'
+    ds = xr.open_dataset(in_filename)
+
+    field = ds.SALT.values.copy()
+
+    # a small averaging kernel
+    x = np.arange(-1, 2)
+    x, y = np.meshgrid(x, x)
+    kernel = np.exp(-0.5 * (x**2 + y**2))
+
+    threshold = 0.01
+    nlon = field.shape[-1]
+
+    lon_with_halo = np.array([nlon - 2, nlon - 1] + list(range(nlon)) + [0, 1])
+    lon_no_halo = list(range(2, nlon + 2))
+
+    for i in range(12):
+        valid = np.isfinite(field[i,:,:])
+        orig_mask = valid
+        prev_fill_count = 0
+        while True:
+            valid_weight_sum = _extrap_with_halo(valid, kernel, valid,
+                                               lon_with_halo, lon_no_halo)
+
+            new_valid = valid_weight_sum > threshold
+
+            # don't want to overwrite original data but do want ot smooth
+            # extrapolated data
+            fill_mask = np.logical_and(new_valid, np.logical_not(orig_mask))
+
+            fill_count = np.count_nonzero(fill_mask)
+            if fill_count == prev_fill_count:
+                # no change so we're done
+                break
+
+            field_extrap = _extrap_with_halo(field[i,:,:], kernel, valid,
+                                           lon_with_halo, lon_no_halo)
+            field[i,fill_mask] = field_extrap[fill_mask]/valid_weight_sum[fill_mask]
+
+            valid = new_valid
+            prev_fill_count = fill_count
+
+    attrs = ds.SALT.attrs
+    dims = ds.SALT.dims
+    ds['SALT'] = (dims, field)
+    ds['SALT'].attrs = attrs
+
+    ds.to_netcdf(out_filename)
+
+def _extrap_with_halo(field, kernel, valid, lon_with_halo, lon_no_halo):
+    field = field.copy()
+    field[np.logical_not(valid)] = 0.
+    field_with_halo = field[:, lon_with_halo]
+    field_extrap = convolve2d(field_with_halo, kernel, mode='same')
+    field_extrap = field_extrap[:, lon_no_halo]
+    return field_extrap
--- a/compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
+++ b/compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
@@ -1,6 +1,3 @@
-import os
-from functools import partial
-
 import numpy as np
 import xarray as xr
 from scipy.signal import convolve2d
--- a/compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
+++ b/compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
@@ -1,4 +1,5 @@
 import numpy as np
+from mpas_tools.io import write_netcdf
 import xarray as xr
 
 from compass.step import Step
@@ -49,7 +50,8 @@ class Salinity(Step):
                       nov='woa23_decav91C0_s11_04.nc',
                       dec='woa23_decav91C0_s12_04.nc')
 
-        for month in ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep', 'octo', 'nov', 'dec']:
+        for month in ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep',
+			          'octo', 'nov', 'dec']:
             woa_filename = woa_files[month]
             woa_url = f'{base_url}/{woa_dir}/{woa_filename}'
 
@@ -72,7 +74,8 @@ class Salinity(Step):
             ds_out[f'{var}_bnds'] = ds_jan[f'{var}_bnds']
 
         slices = list()
-        for month in ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep', 'octo', 'nov', 'dec']: 
+        for month in ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep',
+                      'octo', 'nov', 'dec']:
             ds = xr.open_dataset(
                     f'woa_salin_{month}.nc',
                     decode_times=False).isel(depth=0).drop_vars('depth')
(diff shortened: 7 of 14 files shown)


vanroekel:
@xylar I've made progress on the salinity restoring we discussed.  I was hoping to get a quick look from you to see if i'm on a good track.  I've made a script to create the surface salinity climatology file as close as possible to the PHC2 version (variable names and such).  I haven't hooked up to files for e3sm yet.  I will do so tomorrow and test.

compass/ocean/tests/utility/extrap_woa/salinity_restoring.py
```diff
@@ -0,0 +1,97 @@
...
+                    decode_times=False).isel(depth=0).drop_vars('depth')
+            slices.append(ds.s_an)
+
+        ds_out['s_an'] = xr.concat(slices, dim='time')
+        ds_out['s_an'].attrs = ds_jan['s_an'].attrs
+
+        #Change names of variables and alter attributes
+        ds_out = ds_out.rename_dims({'time':'Time'})
+        ds_out = ds_out.rename_vars({'s_an':'SALT'})
+#        ds_out.SALT.attrs['coordinates'] = "Time lat lon"
```
vanroekel, line 87: @xylar the coordinate from the netcdf files is "Time lat lon depth". I want to change it here but get an error if I try (Value error about an encoding in the attributes and coordinates).

xylar: @vanroekel, the coordinates attribute is added automatically by xarray.  If there is a 1D variable with the same name as a dimension, it is automatically a coordinate.  Otherwise, you have to assign variables as coordinates, e.g.:
```python
ds_out.set_coords('Time', 'lat', 'lon')
```

vanroekel:
The restoring has been moved to its own test case and extrapolation and connection to files for e3sm is completed.  I've run through for the `IcoswISC30E3r5` mesh.  Here is SSS WOA for November
image

Here is the field on the mpas mesh
image

vanroekel:
the new lat, lon WOA23 SSS file has been place in the `initial_condition_database` on anvil -- woa_decav_0.25_sss_monthly_extrap.20241031.nc

vanroekel:
@xylar I think this is ready for your look.  I haven't done the documentation yet, if you'd rather wait for that let me know.  I'll get to that soon

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+import xarray as xr
+
+from compass.step import Step
+
+
+class Salinity(Step):
+    """
+    A step for combining January through December sea surface salinity
+    data into a single file for salinity restoring in G-cases.
+    The top level data of the monthly woa is utilized.
```
xylar, line 12:
```suggestion
    The top level data of the monthly WOA23 is utilized.
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+
+    def __init__(self, test_case):
+        """
+        Create a new step
+
+        Parameters
+        ----------
+        test_case : compass.ocean.tests.utility.extrap_woa.ExtraWoa
```
xylar, line 21:
```suggestion
        test_case : compass.ocean.tests.utility.create_salin_restoring.CreateSalinRestoring
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+                      'octo', 'nov', 'dec']:
+            ds = xr.open_dataset(
+                    f'woa_salin_{month}.nc',
+                    decode_times=False).isel(depth=0).drop_vars('depth')
+            slices.append(ds.s_an)
+
+        ds_out['s_an'] = xr.concat(slices, dim='time')
+        ds_out['s_an'].attrs = ds_jan['s_an'].attrs
+
+        #Change names of variables and alter attributes
```
xylar, line 87:
```suggestion
        # Change names of variables and alter attributes
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+        ds_out = ds_out.rename_dims({'time':'Time'})
+        ds_out = ds_out.rename_vars({'s_an':'SALT'})
+        #ds_out.SALT.attrs['coordinates'] = "Time lat lon"
```
xylar, line 90:
```suggestion
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+        ds_out = ds_out.drop_vars('time')
+
+        # Create a time index
+        time_var = np.arange(1,12.1,1)
```
xylar, line 94:
```suggestion
        time_var = np.arange(1, 12.1, 1)
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
xylar: It looks like a linting step got skipped.  Maybe you didn't have the conda environment loaded when you committed the code so pre-commit was not active.  We'll need to lint the code before it goes in.

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+        ds_out = ds_out.assign(Time=xr.DataArray(time_var, dims=['Time']))
+
+        # Create a xtime array
+        xtime_list = []
+        for i in range(1,13):
```
xylar, line 99:
```suggestion
        for i in range(1, 13):
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+            xtime_string = f"0000-{i:02d}-15_00:00:00"
+            xtime_list.append(xtime_string)
+        xtime_out = np.array(xtime_list,dtype='S64')
+
+        ds_out = ds_out.assign(xtime=xr.DataArray(xtime_out, dims=['Time']))
+
+        #Change attributes to be consistent with PHC restoring file
```
xylar, line 106:
```suggestion
        # Change attributes to be consistent with PHC restoring file
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+        ds_out.Time.attrs['long_name'] = "Month Index"
+        ds_out.Time.attrs['units'] = "month"
+        ds_out.Time.attrs['axis'] = "T"
+
+        #Save file and change time dimension to unlimited
```
xylar, line 111:
```suggestion
        # Save file and change time dimension to unlimited
```

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
```diff
@@ -0,0 +1,112 @@
...
+        write_netcdf(ds_out,'woa_surface_salinity_monthly.nc')
```
xylar, line 112:
```suggestion
        write_netcdf(ds_out, 'woa_surface_salinity_monthly.nc')
```

compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
```diff
@@ -0,0 +1,110 @@
...
+import xarray as xr
+from scipy.signal import convolve2d
+
+from compass.step import Step
+
+
+class ExtrapSalin(Step):
+    """
+    Extrapolate WOA 2023 monthly sea surface salinity data into missing ocean
+	regions, including ice cavities and coasts
```
xylar, line 11:
```suggestion
    regions, including ice cavities and coasts
```

compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
```diff
@@ -0,0 +1,110 @@
...
+        self.woa_filename = 'woa_surface_salinity_monthly_extrap.nc'
+        self.add_output_file(self.woa_filename)
+
+    def run(self):
+        """
+        Extrapolate WOA 2023 model temperature and salinity into ice-shelf
+        cavities.
+        """
+        # extrapolate horizontally using the ocean mask
+        _extrap_level(self.woa_filename)
```
xylar, line 51:
```suggestion
        _extrap(self.woa_filename)
```

compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
```diff
@@ -0,0 +1,110 @@
...
+
+def _extrap_level(out_filename):
```
xylar, line 53:
```suggestion
def _extrap(out_filename):
```

compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
xylar: I don't think "level" makes sense here, where there is only a single level.

compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
```diff
@@ -0,0 +1,110 @@
...
+            filename='woa_surface_salinity_monthly.nc',
+            target='../salinity_restoring/woa_surface_salinity_monthly.nc')
+
+        self.woa_filename = None
+
+    def setup(self):
+        """
+        Determine the output filename
+        """
+        self.woa_filename = 'woa_surface_salinity_monthly_extrap.nc'
```
xylar, line 42:
```suggestion
        now = datetime.now()
        datestring = now.strftime("%Y%m%d")
        self.woa_filename = ''woa23_decav_0.25_sss_monthly_extrap.{datestring}.nc'
```

compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
```diff
@@ -0,0 +1,110 @@
+import numpy as np
+import xarray as xr
+from scipy.signal import convolve2d
```
xylar, line 3:
```suggestion
from datetime import datetime
from scipy.signal import convolve2d
```
(Needed for datestring below.)

compass/ocean/tests/utility/create_salinity_restoring/extrap_salin.py
```diff
@@ -0,0 +1,110 @@
...
+        The name of the output file name after extrapolation
+
+    """
+    def __init__(self, test_case):
+        """
+        Create a new test case
+
+        Parameters
+        ----------
+        test_case : compass.ocean.tests.utility.extrap_woa.ExtrapWoa
```
xylar, line 25:
```suggestion
        test_case : compass.ocean.tests.utility.create_salin_restoring.CreateSalinRestoring
```

compass/ocean/tests/utility/create_salinity_restoring/__init__.py
```diff
@@ -0,0 +1,25 @@
+from compass.ocean.tests.utility.create_salinity_restoring.salinity_restoring import Salinity
+from compass.ocean.tests.utility.create_salinity_restoring.extrap_salin import ExtrapSalin
+from compass.testcase import TestCase
+
+
+class CreateSalinRestoring(TestCase):
```
xylar, line 6: If this is the name of the class, please name the directory `create_salin_restoring` to match.

compass/ocean/tests/utility/create_salinity_restoring/__init__.py
```diff
@@ -0,0 +1,25 @@
+from compass.ocean.tests.utility.create_salinity_restoring.salinity_restoring import Salinity
+from compass.ocean.tests.utility.create_salinity_restoring.extrap_salin import ExtrapSalin
```
xylar, line 2:
```suggestion
from compass.ocean.tests.utility.create_salin_restoring.salinity_restoring import Salinity
from compass.ocean.tests.utility.create_salin_restoring.extrap_salin import ExtrapSalin
```

compass/ocean/tests/utility/__init__.py
```diff
@@ -1,6 +1,7 @@
 from compass.ocean.tests.utility.combine_topo import CombineTopo
 from compass.ocean.tests.utility.cull_restarts import CullRestarts
 from compass.ocean.tests.utility.extrap_woa import ExtrapWoa
+from compass.ocean.tests.utility.create_salinity_restoring import CreateSalinRestoring
```
xylar, line 4:
```suggestion
from compass.ocean.tests.utility.create_salin_restoring import CreateSalinRestoring
```

compass/ocean/tests/utility/create_salinity_restoring/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, test_group):
+        """
+        Create the test case
+
+        Parameters
+        ----------
+        test_group : compass.ocean.tests.utility.Utility
+            The test group that this test case belongs to
+        """
+        super().__init__(test_group=test_group, name='create_salinity_restoring')
```
xylar, line 22:
```suggestion
        super().__init__(test_group=test_group, name='create_salin_restoring')
```

xylar (commented):
@vanroekel, this looks great and just needs a little clean up and documentation.  Feel free to commit my suggestions from GitHub but please also make some arbitrary changes to whitespace locally, then make sure you source the compass load script and then commit the files.  This should trigger pre-commit which will do some linting that seems to be missing.

compass/ocean/tests/utility/create_salinity_restoring/salinity_restoring.py
vanroekel: what linting tool do you recommend?  I ran these all through pylint and it missed a ton of stuff

xylar: It's built in and gets called automatically when you commit with the conda enviornment activated. We use `flake8`, `isort` and some tools for detecting trailing whitespace:
https://github.com/MPAS-Dev/compass/blob/main/.pre-commit-config.yaml

compass/ocean/tests/utility/extrap_woa/extrap_step.py
```diff
@@ -35,12 +35,16 @@ def __init__(self, test_case):
...
-        self.add_input_file(
-            filename='topography.nc',
-            target='../remap_topography/topography_remapped.nc')
+#        self.add_input_file(
+#            filename='woa.nc',
+#            target='../combine/woa_combined.nc')
+
+#        self.add_input_file(
+#            filename='topography.nc',
+#            target='../remap_topography/topography_remapped.nc')
```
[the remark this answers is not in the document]
vanroekel: oh jeez, yep, that is a mistake to leave in there

compass/ocean/tests/utility/create_salinity_restoring/__init__.py
vanroekel: fixed

compass/ocean/tests/utility/__init__.py
vanroekel: fixed

compass/ocean/tests/utility/create_salin_restoring/__init__.py
```diff
@@ -0,0 +1,28 @@
+from compass.ocean.tests.utility.create_salin_restoring.extrap_salin import (
+    ExtrapSalin,
+)
+from compass.ocean.tests.utility.create_salin_restoring.\
```
vanroekel, line 4: @xylar I'm stuck on this line.  If I use this form, it passes flake8 but fails isort.  If I use the previous form 
```
from compass.ocean.tests.utility.create_salin_restoring.salinity_restoring import (
    Salinity,
)
```
It fails flake8 but passes isort.  I'm at a loss on what to do

xylar: Sometimes, you just need to have a long line, e.g.:
https://github.com/MPAS-Dev/compass/blob/7333744570d0ac174ae626192034aa2bf609b839/compass/landice/tests/ismip6_forcing/ocean_basal/__init__.py#L4

```
from compass.ocean.tests.utility.create_salin_restoring.salinity_restoring import (  # noqa: E501
    Salinity,
)  
```

xylar: In general, a comment with `  # noqa: <CODE>` will bypass the linting on that line for that issue.

vanroekel: Thanks @xylar for that tip!

compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
```diff
@@ -29,11 +29,11 @@ def __init__(self, test_case):
                          ntasks=512, min_tasks=1)
 
         self.add_input_file(
-            filename='PHC2_salx.2004_08_03.filled_double_precision.nc',
-            target='PHC2_salx.2004_08_03.filled_double_precision.nc',
+            filename='woa23_decav_0.25_sss_monthly_extrap.20241031.nc',
+            target='woa23_decav_0.25_sss_monthly_extrap.20241031.nc',
```
xylar, line 33: The actual file in the initial condition database needs to be renamed `woa` --> `woa23`.

xylar: I can't do it because I don't have write permission. In general, if you can chmod all your files in `mpas_standalone` so they are group writable, that would be helpful.

xylar: Interesting!  I can rename it without having write permission.

compass/ocean/tests/utility/create_salin_restoring/extrap_salin.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        Determine the output filename
+        """
+
+        now = datetime.now()
+
+        datestring = now.strftime("%Y%m%d")
+
+        self.woa_filename = f'woa23_decav_0.25_sss_monthly_extrap.\
+                              {datestring}.nc'
```
xylar, line 51: It isn't okay to put a continuation line in the middle of a string.  Here's how to do that properly:
```suggestion
        self.woa_filename = f'woa23_decav_0.25_sss_monthly_extrap.' \
                            f'{datestring}.nc'
```

compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
```diff
@@ -44,9 +44,9 @@ def run(self):
         config = self.config
         ntasks = self.ntasks
 
-        in_filename = 'PHC2_salx.2004_08_03.filled_double_precision.nc'
+        in_filename = 'WOA23_monthly_sfc_salin.20241030.nc'
```
xylar, line 47:
```suggestion
        in_filename = 'woa23_decav_0.25_sss_monthly_extrap.20241031.nc'
```

xylar (commented):
@vanroekel, if you make my suggested changes, I'll test again.

compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
vanroekel: Sorry for the stupid question, but here you mean just make sure group has full rwx permissions after I put the file in the initial_condition_database and not figuring how to do this through the python script?

vanroekel:
@xylar Thanks for the testing.  I've taken on your changes and run the full test top to bottom on a different machine and it worked.  This should be ready for your retesting.  I also changed permissions on the files I added to the initial_conditions_database.

compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
xylar: No, these files should not be executable.  Just group readable and writable.  (Folks add execute permission to files a lot that shouldn't have it and this is not a great idea.)  To change permissions recursively, do something like:
```
chmod -R g+rwX .
```
The capital `X` indicates execute permission only gets added to directories, not files.

xylar: > not figuring how to do this through the python script?

Correct, not in the python script.

xylar:
Perfect, I'll re-test as well, just to be safe.  Sounds like we're likely good to go.

compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
vanroekel: I caught my 'x' and did not add it when I changed permissions.  It should just be 'rw' for groups.

xylar:
I ran both tests just fine.  I'm seeing:
woa23_sss_north_pole

I just want to make sure we're okay with the extrapolation artifacts at the north pole.

xylar:
Presumably, this is partly because we extrapolate in lat/lon space, not in real distance.

xylar (approved):
@vanroekel, this looks great, thanks for your patience and hard work.

I'll merge as soon as I get your reaction to my north pole plot.

vanroekel:
I'm not sure about the artifacts on the N. Pole.  Looks like a fair bit of variability over the pole.  I'd ask what @alicebarthel and @milenaveneziani think.

Does this same kind of artifact show up in the initial condition too?

vanroekel:
and thank you very much for the review and help on this @xylar!

xylar:
I can't imagine the same artifact isn't in the initial condition since the approach is really similar.  I haven't looked.

milenaveneziani:
I agree that the artifacts are likely present in both the initial condition and this new restoring file. This makes me wonder something that I have been thinking for a while though (please don't hate me): should we use PHC3.0 for both, ie migrating from WOA to PHC initial conditions? PHC was designed to represent the polar ocean state (especially the Arctic), so I would be surprised if it led to the artifacts seen here. I know there is a lot work behind this PR, but I wanted to share my 2c.

Another option is to try to understand what led to the artifacts: is it the extrapolation/interpolation process onto the MPAS mesh? But then wouldn't we see them closer to the North Pole? I have looked at WOA and PHC salinities for the month of January, and I believe both data sets should go all the way to 89N, or maybe even 89.5N.

xylar:
> should we use PHC3.0 for both, ie migrating from WOA to PHC initial conditions?

I have not heard of PHC3.0 being better in the Antarctic.  I would be pretty opposed to moving to PHC3.0 without knowing how it behaves there.

> is it the extrapolation/interpolation process onto the MPAS mesh?

No, we are extrapolating on a lat-lon mesh because we need a file that is preprocessed once and just gets remapped to each MPAS mesh.

> I have looked at WOA and PHC salinities for the month of January, and I believe both data sets should go all the way to 89N, or maybe even 89.5N.

The the artifacts must be in WOA23 from the start and not be due to our extrapolation. But it very much looks like it's *their* extrapolation rather than coming from local data.

I don't want to invest work in it but I could imagine someone stitching together PHC3.0 in the Arctic with WOA23 elsewhere.

milenaveneziani:
> The the artifacts must be in WOA23 from the start and not be due to our extrapolation. But it very much looks like it's their extrapolation rather than coming from local data.

This is what I have for January salinity from both data sets: 
Screenshot 2024-10-29 at 1 11 47 PM
Screenshot 2024-10-29 at 1 13 48 PM

They look pretty similar to me, at least at this level of detail.

xylar:
It just doesn't really show up in a lat-lon plot.  We'd need a polar plot.

And, wow, those data sets look really different most places but especially in the Arctic!

xylar:
Ah, sorry, just seeing that they're offset by 180 degrees from each other.  Phew!

milenaveneziani:
yes, sorry, I was being lazy and just plotting the data with a couple of lines of matlab code.. But I could spend some extra time on it, making better plots using python and the polar stereographic projections.
I'll let @vanroekel and @alicebarthel chime in before moving in that direction though.

alicebarthel:
I think discussing the WOA vs. PHC is worthwhile, esp. if we end up spinning up with close SSS restoring. I did not anticipate them to be that different
Screen Shot 2024-11-01 at 5 15 25 PM

I seem to remember that Laurie Padman looked into doing a better interpolation of WOA, if we end up questioning the raw dataset we want to use. But let's see the 
As to the artifacts, would a smoother field near the North pole (away from the coasts) be all we need for it to pass the sanity check? Or are we concerned about the values (and differences in datasets)? I am just trying to articulate what would be a "good enough" field.

milenaveneziani:
WOA is fresher than PHC overall, and that is also visible in the raw plots that I made with matlab (except for the Amazon and other South American rivers, which seem to be missing completely in WOA for some reason). I can definitely make more decent plots focused on the polar regions and for all months (next week).

alicebarthel:
@milenaveneziani Yes, I knew it was fresher but I had not grasped how big the deltaSSS was... @vanroekel is there an updated file on Icos30? I am looking at the ones you shared yesterday, had you updated the date to Jan 1 then?

xylar:
@milenaveneziani and @alicebarthel, isn't PHC 3.0 what we were using long ago before we moved to WOA?  Wasn't it crazy in some places?  It appears to be from 2005 https://psc.apl.washington.edu/nonwp_projects/PHC/Climatology.html and their web design appears to be from 1995.  Do we really see that as the way forward??

alicebarthel:
The features seem to be in the original data
Screen Shot 2024-11-01 at 6 30 22 PM
Screen Shot 2024-11-01 at 6 29 54 PM

milenaveneziani:
Nice plots Alice, thank you.
@xylar (btw, for a second I thought you were in Germany and spending your early Saturday morning working.. :o): I personally think that the very-old time stamp of PHC doesn't really matter, as long as we think it is a superior (smoother) product for our initial conditions. But we need to make sure of that, and also look closely at SO climatologies (where we could compare with SOSE, maybe?).

xylar:
> (where we could compare with SOSE, maybe?).

Oh, no!  Definitely not!  SOSE is much worse than WOA23 in the southern ocean, particularly on the continental shelf.

xylar:
PHC is still available in Compass as an initial condition.  It would just be a question of adding the variants of the test cases that use it back into the list.  But I am very much against using it in its current form as our global initial condition because it really isn't great in the regions I know.

vanroekel:
I'm really not comfortable with PHC (especially what we have when we can't track the history of how it was developed).  As @xylar pointed out the website you download PHC3 from doesn't instill confidence.  Also I recall @proteanplanet mentioning that even Mike Steele doesn't recommend the use of PHC (preference to EN4 actually if I recall right).  So in these delta plots how much confidence do we have that PHC is closer to truth?

I also don't feel highly detailed investigations of ICs and SSS restoring for G-cases are good uses of time and resources.  For decadal+ projections I don't see the details of these diffs to be highly impactful.  

My feeling is the questions raised here don't really affect this PR.  I'm definitely open to discussions of alternate ICs and other SSS options in a different forum.

vanroekel:
Thinking a little more - if we had confidence that the more saline arctic was closer to reality and had a good data file (RARE data?) maybe that would be good to explore since this will affect long G-cases.  But again it's a tension, the focus of E3SM is on coupled cases, so we can't expend a significant chunk of effort on SSS restoring.

xylar:
@vanroekel, I went through this same debate in my head.  I think it's not worth going overboard about the initial condition but I do think the salinity restoring might deserve a bit of extra scrutiny.  If we restore strongly to that weird pattern at the north pole, we might notice it.  If we restore weakly, we probably wont.

xylar:
@vanroekel, I'm assigning this to you.  Please hit the green merge button when you are ready.

vanroekel:
@milenaveneziani can you make the plots for all months you mention above (for the SSS fields)?  Also do you have any sense on best in class products for SSS in the arctic?  what do you think about comparing to a product like RARE for SSS?  or some other reanalysis?

milenaveneziani:
I think Kat worked on making RARE climatologies. I will double check with her.
And yes, I will make the monthly plots for WOA23 and PHC. I will also ask Mike Steel directly about what he thinks of his product.

vanroekel:
Thanks @milenaveneziani!  I think it would still be good to compare to another data source as well (even if I misremembered and Mike does think PHC is best), I'm curious how much variance there is in Arctic SSS.

milenaveneziani:
I have all monthly plots for both PHC and WOA2023. I used pcolormesh in cartopy, so that there is not interpolation whatsoever. I will post Jan, Apr, Jul, and Sep for now.

milenaveneziani:
Jan. PHC on the left, WOA2023 on the right: 
sssm01

milenaveneziani:
Apr: 
sssm04

milenaveneziani:
July: 
sssm07

milenaveneziani:
And Oct: 
sssm10

milenaveneziani:
Kat no longer has the RARE climatologies, so those would have to be recalculated. But of course they wouldn't work for the SO.

vanroekel:
Thanks @milenaveneziani this is intriguing.  For the arctic it looks like since we have weak AMOC we'd probably want to use PHC3.  It still would be nice if we could get a sense of if PHC3 is closer to truth.  Any thoughts on how to do that?

The story seems switched for the S. Ocean.  There are more of those wave like structures in the SSS field and the WOA23 looks smoother.

I'm not sure what the best path is.  I'll need to think on this.  Let me know what you think after looking at these.

vanroekel:
you're right - the structures seem very odd in WOA23.  Could some of the difference be vertical resolution?  I recall WOA23 has higher near surface resolution, but don't know what PHC3 is.

What are your feelings on thinking about blending WOA and PHC?  Seems scary to me, but I don't know how to pick up what we want in each hemisphere.  The other alternative is two sss restoring files that are used for different cases, but I'm not a huge fan of that either.  I'm also curious what @xylar an @irenavankova think too.

proteanplanet:
Just to reiterate @vanroekel's earlier point that not even the authors of PHC recommend it now.

milenaveneziani:
Those patterns in the Arctic in WOA are worrisome to me. I don't think we want to use WOA-SSS restoring for G-cases that have a north polar ocean focus.
I sent an email to Mike and report back when I hear.

About the vertical resolution, Luke: WOA has a 5m resolution in the upper 100m, PHC has 10m res. Are you thinking that the artifacts are related to vertical interpolation?

irenavankova:
Looking at the Weddell and Ross patterns, they look much better to me in WOA. I am not sure what happens in Weddell in October though that the salinity increases suddenly over a large area. While the pattern in Weddell looks reasonable for April and July, the peak salinities in front of western FRIS front are too low. They need to reach 34.9 as that is very consistently observed at the sea floor for three decades now. What I see from the WOA plot is 34.5 instead. 

When I was looking at the PHC earlier in the Southern Ocean I saw some places with a one cycle in a season and other places with two cycles in a season. I wasn't sure whether that was physical or not.

xylar:
@vanroekel, (and everyone), I don't feel comfortable with PHC but I do share @milenaveneziani's concern that WOA23 looks weird enough in the middle of the Arctic that it could cause weird behavior there.  My guess is that data is just sparse there and no product is well constrained.  Even so, some may be more plausible than others.

I continue to feel that salinity restoring is a crude instrument, but particularly anywhere you have salinification (areas of sea ice production and evaporation).  Whereas fresher water stays at the surface, saltier water sinks so there's less of a signature in the SSS.  So my feeling is partly that, if you are restoring hard enough that the details of the SSS restoring field matter especially where you have sea ice cover, even having the correct SSS field doesn't relate to having the right salt fluxes.

milenaveneziani:
@irenavankova, about this:
>They need to reach 34.9 as that is very consistently observed at the sea floor for three decades now. What I see from the WOA plot is 34.5 instead.

Note that those salinities I plotted are SSS, not Sbot. I can also plot the bottom level (I actually thought about it, but I just didn't want to inundate this PR with plots). Maybe I'll just make a one month plot for the SH only.

milenaveneziani:
Here is Mike's response, mingled with my own comments:

> PHC is the best product for initializing a global model when the first year of the run is some time during the period 1950-1990. This is the period over which the classified Russian Soviet data were collected; these are included in PHC. During this time, non-Soviet data coverage in the Arctic was horrible. Also: the Soviet data that we used was "pre-smoothed" and interpolated, which is why it looks so nice.

Luke, Alice: this makes me think that PHC could work well for the new initial condition process that we discussed (1-year G-case, which of course would start in 1958), but perhaps not for the long, standard G-cases.

> However, since 1990 the Arctic T and S fields have of course changed a lot. "Atlantification," freshening in the Beaufort, etc. So PHC is a poor choice for validation of Arctic Ocean simulations in recent years. Also it will drag your model solutions back to pre-1990s conditions if you use it for restoring.

I am less worried about this, and more worried about PHC in the SO.

> My impression is that people in Europe have more money for earth science, and often produce better quality data sets. Specifically, NOAA's WOA is often lacking recent observations in the Arctic Seas, owing to a lack of human resources for collecting and QC'ing the data. Meanwhile, the UK Met Office's hydrographic product called EN4 includes both WOD profiles and other, more recent data, and so I believe it is the best product to use now for Arctic research (and probably global research). Although it is not perfect! We have found static instabilities and other errors in it. EN4 provides T and S profiles (like NOAA's WOD) as well as gridded T and S fields for long-term climatology (like NOAA's WOA) as well has monthly mean fields with interannual variability. The latter likely defaults to climatology often, since we don't have lots of data every month, but still it could be nice.

For some reason, I had forgotten about EN4. I'll see if I can find climatologies easily; I know I have downloaded it a while back, but I can look into it again.

My conclusion in all this is similar to yours, Luke: there is likely no perfect product for the whole globe. One option is to allow (within compass) for a couple of choices, say WOA and EN4, and then the user decides what to do when they are building their cases from scratch.

irenavankova:
> @irenavankova, about this:
> 
> > They need to reach 34.9 as that is very consistently observed at the sea floor for three decades now. What I see from the WOA plot is 34.5 instead.
> 
> Note that those salinities I plotted are SSS, not Sbot. I can also plot the bottom level (I actually thought about it, but I just didn't want to inundate this PR with plots). Maybe I'll just make a one month plot for the SH only.

Yes, I am aware SSS was plotted. But I thought that for 34.9 to be observed  on the sea floor, it must have been at least that high at the surface at some point prior to that.

xylar:
Regarding EN4, I have used it in ISMIP6.  If it has been updated since then to include newer (especially seal-based) Antarctic observations, it might be worth considering.  @vanroekel tried their 1900 climatology as an initial condition and it didn't help whatever the issue was at that time (Compass still supports that).

We can give it a go...

milenaveneziani:
@irenavankova : here is WOA bottom salinity (1500 m) for April (never mind the title, and also note the different colorbar levels than before): 
woa2023_sssSH_month04_pmesh_z-1
values are more 34.7 than 34.9.

xylar:
So it's basically missing the continental shelf?

vanroekel:
I guess we can try EN4 and see if that's reasonable.  What should we use for the climatology?  1993-2023?  Getting that processed will be trickier since the data I'm aware of is by year.  I can't seem to find the climatology Mike mentions.

vanroekel:
Before changing this PR though, Milena can you make some plots of EN4?  This is where I've found data - https://www.metoffice.gov.uk/hadobs/en4/download-en4-2-2.html

I'd prefer to find a product that is acceptable than try have multiple G-case configs.

xylar:
From what I can tell, EN4's gridded products use a climatology from 1970-2000 wherever data is missing. That worries me for creating a climatology that's ostensibly supposed to be more recent.

In general, I worry about creating our own climatology.

vanroekel:
I agree @xylar tinkering around with making our own climo is worrying.  Is it worth exploring other editions of WOA?  CESM uses WOA18.  

As I've thought about this, I'd really prefer to use a single product and not blend or have a bunch of different options.  Maybe the available data precludes desire this though

xylar:
I am in the process of switching from WOA18 to 23 in MPAS-Analysis. The differences seem to be negligible.

alicebarthel:
Thanks, that's helpful to hear Mike's comments about PHC. I agree with @vanroekel and @xylar's concern about making our own dataset. My main concern in bringing it up was to have a sense for the agreed-upon best estimates, or the uncertainty range associated with monthly values in the Arctic. If we are simply concerned about the polar features (and satisfied with the magnitude), would a small degree of smoothing for the SSS restoring fields be an option? SSS restoring is already a model-per-model decision, and it sounds like PHC was smoothed. That would avoid impacting the broader E3SM/B-case methodology. 
The discussion as to which dataset is best for initialization is still valid, particularly if EN4 looks significantly better.

vanroekel:
@alicebarthel I was thinking about smoothing as well.  That seems like a potentially good option.

Agreed on initialization being a valid discussion, but that should be a different issue and not a part of this PR.

milenaveneziani:
Sorry for the delay all; trying to get things done while going through a nasty case of covid.
I was also looking at the EN4 links yesterday. My intention was to download the objective analysis data set for as far back as possible, and then compute my own climatology. I was actually thinking to do two climatologies: one pre-1990 and one 1990-2024. I don't see a problem with doing that, as far as we document what we did. Not sure why you are concerned?

xylar:
Sorry @vanroekel for continuing to hijack this PR. We should move this discussion somewhere else.  Someone feel free to make a space for this and we can move it there.

> I was actually thinking to do two climatologies: one pre-1990 and one 1990-2024. I don't see a problem with doing that, as far as we document what we did. Not sure why you are concerned?

My concern is with the strategy they use to fill in the gaps.  They use a 1970-2000 climatology.  That means if you do a 1990-2024 climatology, rather than having holes where there was no data, you will have 1970-2000 data.  In my view, it would be better to have the holes and interpolate/extrapolate to fill them in rather than fill them in with data that is 20 years older than you think you have.

vanroekel:
How should we finish this PR?  Should we add some smoothing?  @milenaveneziani do you want to plot SSS from EN4?

vanroekel:
my initial feeling is to just try some smoothing. I think where/when we have a more robust IC discussion, we may end up revisiting SSS restoring to be consistent with the IC chosen.

milenaveneziani:
Yes, smoothing of the WOA field sounds good: hopefully it helps.
I would treat the whole EN4 issue as separate discussion.

vanroekel:
Let's plan on trying smoothing in this PR and in the IC discussion if we switch from WOA we definitely want to reconsider SSS

vanroekel:
@xylar I hope to return to this this week, do you have any advice on how to explore smoothing?  I was thinking the smoothing should be in the step where the extrapolated WOA file is made.

xylar:
@vanroekel, yes, I agree.  It should be possible to call `_extrap_with_halo()` a few extra times once you're done extrapolating and let that do the smoothing for you.  Unlike for extrapolation, you would use the full smoothed array, not just the portion that was previously invalid.  Let me know if that's clear or if you need further details.  The number of iterations of smoothing should be a config option.

milenaveneziani:
Just to close the loop on the EN4 question: as per our slack exchange with Luke and Alice, we have decided against using EN4 salinity, since the climatology that I computed (using the same years that they use for filling the gaps: 1971-2000) showed that 1) EN4 SSS seems even fresher than WOA in the Eastern Arctic and 2) smoothing would still be necessary for EN4. Xylar: if you are interested in looking at EN4 SSS (or bottom level salinity) for the Southern Ocean, let me know, and I can send the plots to you and Irena via slack.

vanroekel:
@xylar I took a first pass at smoothing in the northern high latitudes.  if you have a minute could you have a quick look and let me know if this is along the lines of what you envisioned?  Here is SSS in the arctic, no smoothing
image


here is 30 iterations
image

It seems to be working, but 30 doesn't appear enough.  I may try a higher latitude cutoff and smooth further.

vanroekel:
here is 60 iterations

image

the north pole extrapolation artifacts are still visible.

xylar:
@vanroekel, it seems like maybe the approach I suggested isn't very effective.  The smoothing is taking place in lat-lon space, not in physical distance so it's very inefficient at the pole, precisely where the artifact exists.  I think we should brainstorm an alternative.

maltrud:
@vanroekel and @xylar is this really a big deal?  I understand it's annoying but seems like not the best use of your time (IMHO).  if you really do care, maybe just replace values poleward of, say, 65N with the zonal average.  it's not like we really know the salinity up there anyway.  my unsolicited 2 cents worth....

vanroekel:
In a discussion with @xylar we decided to merge this feature as is, without smoothing.  Prior to creating the new initial condition for E3SM, a follow on PR will be issued to use the new smoothing capability when interpolating to the MPAS mesh from @xylar and @bmooremaley 

@xylar I've removed the smoothing commit, do you want me to merge this in?

xylar:
Yep, go for it!

PR #874, opened 2024-12-03
Add python 3.13 support and make it the default

xylar:
This merge also updates mpas_tools to 0.36.0 and geometric_features to 1.6.0, the latest versions.  Thus, it bumps compass to 1.6.0-alpha.2.

Some unused files are removed.

Checklist


altheaden:
@xylar What you have now looks good to me. I think you also want to add 3.13 to the `setup.py`. There is also a mention of the latest python version in `docs/developers_guide/quick_start.rst` which you might want to update as well. I also see a mention of 3.12 in the `conda/default.cfg` but I'm not sure if that is something you'd want to update. I have these changes in a branch right now, so if you'd like I can push any or all of those here.

altheaden:
@xylar just pushed those changes for you!

xylar:
I feel like I should run the `pr` suite on Chrysalis to avoid any surprises now that python 3.13 has become the default.

altheaden:
Let me know if you'd like me to do any testing as well.

xylar:
@altheaden, thanks for your help debugging and reviewing this!

PR #876, opened 2024-12-19
Enable coarse mesh resolution over the open ocean by using geojson file.

andrewdnolan:
This pull request is a work in progress. It's purpose is to produce a coarse but variable resolution MALI mesh to be by used E3SM for testing / debugging purposes.

Checklist


compass/landice/tests/greenland/mesh.py
```diff
@@ -91,8 +95,9 @@ def run(self):
         build_mali_mesh(
             self, cell_width, x1, y1, geom_points, geom_edges,
             mesh_name=self.mesh_filename, section_name=section_name,
-            gridded_dataset=source_gridded_dataset_1km,
-            projection=src_proj, geojson_file=None)
+            gridded_dataset=source_gridded_dataset_1km, projection=src_proj,
+            geojson_file="greenland_only_outline_45km_buffer_latlon_singlepart.geojson",  # noqa: E501
```
matthewhoffman: @trhille , will this line make this new geojson buffer the new behavior?  Is that what we want?

trhille:
Testing
-------
I successfully created a 4–40km mesh with these changes:
image

However, I notice that the geojson file seems to extend beyond our bed topography data set in a few places and we get artifacts with bedTopography = 0 at the edge of the mesh:
Screenshot 2026-02-27 at 1 37 00 PM

I'm not sure what's the best way of dealing with this. We could manually edit the geojson to stay within the bounds of the gridded dataset. Or we could set a mask that is the intersection of the two.

compass/landice/tests/greenland/mesh.py
trhille: To answer your second question, I think we want this to be the new default behavior, but that could still be introduced through the default .cfg file instead of hardcoded in the script.

andrewdnolan: A config option seem like the right call. We can have it on by default if, but also the option turn it off and/or use a different buffer geojson file.

trhille:
My testing confirms that those interpolation artifacts:
image

PR #878, opened 2024-12-22
Update to geometric_features v1.6.1

xylar:
This brings in some bug fixes to Greenland and historical sea-ice regions.

Because of the dependency changes, we also update to compass to v1.6.0-alpha.3

Checklist


xylar:
In draft mode for now because I don't think the new package is available yet.

xylar:
@alexolinhager, are there any compass changes you are aware of that are required because of this change?

xylar:
I'm moving this to #879

PR #881, opened 2025-01-09
Update to 1.7.0 alpha.2

xylar:
This brings in:
* mpas_tools v1.0.0 -- modernizes command-line tools (including renaming them)
* pyremap v2.0.0 -- modernization, refactoring, API-breaking changes, MOAB support

As a result of the `mpas_tools` update, many command-line tools used by `landice` have been renamed.  The relevant documentation has also been updated.

Checklist

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -688,14 +688,14 @@ def build_mali_mesh(self, cell_width, x1, y1, geom_points,
     logger.info('writing grid_converted.nc')
     write_netcdf(dsMesh, 'grid_converted.nc')
     levels = section.get('levels')
-    args = ['create_landice_grid_from_generic_MPAS_grid.py',
+    args = ['create_landice_grid_from_generic_mpas_grid',
             '-i', 'grid_converted.nc',
             '-o', 'grid_preCull.nc',
             '-l', levels, '-v', 'glimmer']
 
     check_call(args, logger=logger)
 
-    args = ['interpolate_to_mpasli_grid.py', '-s',
+    args = ['interpolate_to_mpasli_grid', '-s',
             gridded_dataset, '-d',
             'grid_preCull.nc', '-m', 'b', '-t']
 
@@ -703,7 +703,7 @@ def build_mali_mesh(self, cell_width, x1, y1, geom_points,
 
     cullDistance = section.get('cull_distance')
     if float(cullDistance) > 0.:
-        args = ['define_cullMask.py', '-f',
+        args = ['define_landice_cull_mask', '-f',
                 'grid_preCull.nc', '-m',
                 'distance', '-d', cullDistance]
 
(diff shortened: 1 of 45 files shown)


xylar:
This needs:
https://github.com/MPAS-Dev/MPAS-Tools/pull/596
https://github.com/MPAS-Dev/MPAS-Tools/pull/597
and a new release of MPAS-Tools (v1.0.0) before it will be useful.

xylar:
## Testing

I ran both the `pr` and `full_integration` suites on Chrysalis with this new version of `mpas_tools` and the associated changes in this branch.  I didn't check that thing were BFB however and I need to do that.

xylar:
@trhille and @matthewhoffman, I will need your help to make sure I didn't break anything for you 2.  Can you run tests that go beyond `full_integration` to make sure other affected tests aren't broken?  Then, can you review this and https://github.com/MPAS-Dev/MPAS-Tools/pull/596 together?

compass/landice/mesh.py
```diff
@@ -688,22 +688,22 @@ def build_mali_mesh(self, cell_width, x1, y1, geom_points,
     logger.info('writing grid_converted.nc')
     write_netcdf(dsMesh, 'grid_converted.nc')
     levels = section.get('levels')
-    args = ['create_landice_grid_from_generic_MPAS_grid.py',
+    args = ['create_landice_grid_from_generic_mpas_grid',
```
matthewhoffman, line 691: @xylar , to clarify, we could replace all of these changes with function calls, right?  I'm assuming you retained the `check_call` calls to minimize how invasive this PR is.  @trhille and I are on board with that, but just wanted to make sure we understood what's going on.

xylar: I would have to remind myself of the details.  You should not try to call the entry-point functions because the command-line arguments are still expected for those and they won't work as expected.  But in many cases there are other functions with "proper" argument you could call instead and that might be preferable.

In any case, yes, I tried to change things minimally.

xylar: Looking at this further:
https://github.com/MPAS-Dev/MPAS-Tools/blob/701c8d97dfa733ec7128f32aff5ddd4f48a14e2c/conda_package/mpas_tools/landice/create.py#L10
The function `create_from_generic_mpas_grid()` as written is not really callable because it assumes that `OptionParser.parse_args()` will return the expected command-line arguments (which `compass` obviously wouldn't do).  Elsewhere in `mpas_tools`, we have a function with proper arguments that can be called directly and then we have entry point function that do arg parsing and then call the other function.  It would be great to move in that direction but I didn't try to do that for you all because it was one step too far for me.

matthewhoffman: @xylar , got it - that makes sense.  I'm happy with changing things in this way for now (here and the companion MPAS-Tools PR).

trhille (approved):
Approved based on the testing described here: https://github.com/MPAS-Dev/MPAS-Tools/pull/596

xylar:
@matthewhoffman and @trhille, the versions of pyremap and mpas_tools that this update depends on have been released.  There is a non-zero chance that I've missed a few updates needed for either of these.  We can either do more testing now, before this goes in or we can go ahead and merge this, and fix anything that remains in follow-up PRs.  Basically, it's a question of how much you want to suffer now vs. potentially suffer later...

xylar:
I'll certainly re-test the `pr` and `full_integration` before I merge.

xylar:
## Further testing

`full_integration` and `pr` suites pass on Chrysalis with gnu and intel, respectively.

matthewhoffman (approved):
Approving based on inspection and @trhille's testing

xylar:
Thanks @trhille and @matthewhoffman!!

PR #883, opened 2025-01-13
Add smoothing to remapping of SSS restoring

xylar:
This requires a major reorganization of the step to be more like `remap_topography` after its recent reorganization.

Remapping is performed with `mbtempest`.  The public function `remap_sss()` has been removed.  (It was not mentioned in the docs.)

This merge also adds a remapping step to the utility that generates the SSS field from WOA23.  The data is remapped to a cubed-sphere grid at ne300 (~10 km) resolution.  Remapping from the cubed sphere to MPAS meshes will be *much* more performant than remapping from the WOA original lat-lon grid, particularly when there is smoothing near the poles as we do here.

The smoothing goes linearly from zero at 70 N to 1000 km at 90 N.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
+++ b/compass/ocean/tests/global_ocean/files_for_e3sm/remap_sea_surface_salinity_restoring.py
@@ -1,19 +1,23 @@
 import os
+import pathlib
 
+import numpy as np
 import xarray as xr
 from mpas_tools.io import write_netcdf
-from pyremap import LatLonGridDescriptor, MpasCellMeshDescriptor, Remapper
+from mpas_tools.logging import check_call
+from pyremap import MpasCellMeshDescriptor
 
 from compass.io import symlink
 from compass.ocean.tests.global_ocean.files_for_e3sm.files_for_e3sm_step import (  # noqa: E501
     FilesForE3SMStep,
 )
+from compass.parallel import run_command
 
 
 class RemapSeaSurfaceSalinityRestoring(FilesForE3SMStep):
     """
-    A step for for remapping sea surface salinity (SSS) from the Polar science
-    center Hydrographic Climatology (PHC) to the current MPAS mesh
+    A step for for remapping sea surface salinity (SSS) from WOA23 to the
+    current MPAS mesh
     """
     def __init__(self, test_case):
         """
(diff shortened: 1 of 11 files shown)


xylar:
## Testing

I ran the `remap_sea_surface_salinity` step for the `IcoswISC30E3r5` mesh.  The step took about 3 minutes on 4 Chrysalis nodes (without using the cubed-sphere grid it had not completed after 30 minutes).  With a maximum smoothing length scale of 1000 km (the current default), the result is this field:

sss_jan_with_smoothing

Compare with a similar plot without smoothing:

image

xylar:
Here is what the smoothing field looks like:
expand_dist

The units are meters.

xylar:
@vanroekel, this is now ready for your review.  Feel free to review the code (by inspection) but the main review I would like is whether you are good with the outcome.

xylar:
@maltrud, @milenaveneziani, @alicebarthel and @irenavankova, you all commented on https://github.com/MPAS-Dev/compass/pull/872 so I wanted to make sure you're aware of this follow-up work.

milenaveneziani:
It looks good to me @xylar, thanks for working on this.
If we wanted, could we apply this tool to smooth the WOA data that we use in MPAS-Analysis? This would eliminate the weird patching that we usually see in the bias plot.

xylar:
> If we wanted, could we apply this tool to smooth the WOA data that we use in MPAS-Analysis? This would eliminate the weird patching that we usually see in the bias plot.

It's something we could think about.  It relies on conservative remapping with MOAB and that performs really badly with latitude-longitude grids (because there are so many tiny, overlapping cells near the poles).  I had to switch to putting the SSS fields on a cubed-sphere grid here to make the remapping/smoothing to the MPAS mesh performant.  I'm not sure how feasible it is to take a similar approach in MPAS-Analysis.

milenaveneziani:
I see. Thanks for clarifying that part, I had missed it.

xylar:
> I had missed it.

I see I forgot to add that to the PR description!  My mistake.  It's there now.

vanroekel (approved):
Approving by visual inspection of code and examination of resulting smoothed field.

Thanks @xylar this looks great to me

xylar:
Thanks @vanroekel!

alicebarthel:
Thanks @xylar , this looks great!

PR #884, opened 2025-01-14
Add `pre-commit` autoupdate workflow

altheaden:
This workflow runs weekly on Wednesday to check for updates to `pre-commit` and its dependencies. If available, the updates will be automatically applied and put into a new PR.

Checklist


altheaden:
@xylar As is the case with Polaris, I haven't done any additional testing for this PR, but it will run tomorrow morning and we can observe the behavior then.

xylar (approved):
Yep, this looks great!  thanks for getting to this so quickly!

PR #886, opened 2025-01-15
In CI, use a unique cache directory for each python version

xylar:
We're seeing conflicts when CI runs and I'm hoping this will fix it.

This also turns off fast-fail in the CI matrix build.


xylar:
@altheaden, I think this is working.  Could you review this and then make a similar PR to the other repos?  I think the problem we've been seeing is that all the tests are trying to update files in the same location, whereas we need separate caches for each test.

xylar:
@altheaden, let's make sure we turn off `fast-fail` elsewhere as well when we're making this change.

altheaden (approved):
Looks good to me.

PR #888, opened 2025-01-24
Add a step for smoothing topograpy

xylar:
We remap the topography once without smoothing and then once with smoothing.  The unsmoothed topography is used to determine the land mask, thus allowing many meshes with different levels of smoothing to have the same mapping files in E3SM.

For the time being, only two simple types of smoothing are supported: a smoothing over a fixed distance or by expanding cells by a fixed fraction.  (The two types of smoothing can also be used at the same time.)

In the future, maps with non-uniform smoothing could be added for finer control.

A check use used to bypass the second smoothing pass if, in fact, no smoothing is requested (the default behavior).

Checklist

### DIFF
--- a/compass/ocean/mesh/remap_topography.cfg
+++ b/compass/ocean/mesh/remap_topography.cfg
@@ -14,7 +14,7 @@ description = Bathymetry is from GEBCO 2023, combined with BedMachine
               Antarctica v3 around Antarctica.
 
 # the target and minimum number of MPI tasks to use in remapping
-ntasks = 1280
+ntasks = 640
 min_tasks = 256
 
 # remapping method {'bilinear', 'neareststod', 'conserve'}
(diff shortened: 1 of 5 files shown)


compass/ocean/mesh/remap_topography.cfg
```diff
@@ -30,7 +30,7 @@ ice_density = 910.0
 
 # smoothing parameters
 # no smoothing (required for esmf):
-#     expandDist = 0   [m]
-#     expandFactor = 1 [cell fraction]
-expandDist = 0
-expandFactor = 1
+#     expand_distance = 0   [m]
```
bmooremaley, line 33: We can get rid of the commented parameters here.

xylar:
## Testing

I ran the Icos240 mesh, init and performance tests using a smoothing factor of 2.  I looked at the results in ParaVeiw and the smoothing was visible in the bed_elevation field.  I also saw that the topography had been remapped in the step with smoothing (whereas it should just be symlinked if there is no smoothing).

I'm not yet able to run the `pr` suite because I need to update the cache files.

xylar:
## Testing

I successfully ran the `pr` suite on Chrysalis after caching the required mesh and init files.

PR #891, opened 2025-02-04
CI updates and cleanup

altheaden:
This PR adds a few CI updates we've implemented in other repos recently:

1. I updated the default python version in various places to 3.13, and changed this version from a magic number to a variable in a few places (for consistency). 
2. I enabled `auto-update-conda` which we had disabled previously thinking it was causing trouble, but it wasn't.
3. I removed the caching step for creating a conda environment - it doesn't appear to be saving any time and we don't consider it worth keeping.

Checklist

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -12,6 +12,7 @@ on:
 env:
   CANCEL_OTHERS: false
   PATHS_IGNORE: '["**/README.md", "**/docs/**"]'
+  PYTHON_VERSION: "3.13"
 
 jobs:
   pre-commit-hooks:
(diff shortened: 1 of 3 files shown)


altheaden:
## Testing
I pushed this to my `main` branch on my fork and ran all the CI tests, everything passed and the docs built as expected (as far as I can tell).

@xylar I updated 3.10 -> 3.13 in a few places besides the docs, so let me know if that looks OK to you.

altheaden:
@xylar I don't have permission to re-run jobs but I think the failed job is due to the CI issues we've been experiencing lately.

xylar (approved):
Looks great!

PR #894, opened 2025-02-21
Enable Option for High Res Mesh in Ocean

alexolinhager:
This PR creates the option (`max_res_in_ocn=True`) to maintain the highest resolution used on land ice cells in all ocean cells when generating the GIS MALI mesh. This is done to resolve fjord geometry and coastal bathymetry necessary to translate ocean properties to glacier termini when coupling MALI to MPAS-Ocean (or prescribing ocean forcing to MALI in standalone mode). As this has the potential to drastically increase mesh generation time, so there is an additional option `define_bnds_by_geojson=True` that performs an initial cull of the BedMachine dataset according to the bounds set by `compass/landice/tests/greenland/gis_contShelfExtent_PS.geojson` before generating the mesh in Jigsaw.

Checklist


compass/landice/mesh.py
```diff
@@ -265,12 +265,10 @@ def set_cell_width(self, section_name, thk, bed=None, vx=None, vy=None,
     low_bed = section.getfloat('low_bed')
     high_bed = section.getfloat('high_bed')
 
-    # convert km to m
-    cull_distance = section.getfloat('cull_distance') * 1.e3
```
trhille, line 269: We definitely don't want to remove culling by distance entirely.

trhille (changes requested):

trhille:
@alexolinhager, this is missing an updated `mesh_gen.cfg` file.

compass/landice/mesh.py
```diff
@@ -290,18 +288,30 @@ def set_cell_width(self, section_name, thk, bed=None, vx=None, vy=None,
...
+        # If max_res_in_ocn is true, use the minimum cell spacing for
+        # all ocean cells. Important for ocean-coupled runs where
+        # resolving fjords and ocean bathymetry is necessary for
+        # thermal forcing parameterizations. Otherwize, telescope out
+        # to coarse resolution with distance from the ice front.
+        if section.get('max_res_in_ocn') == 'True':
+            spacing_bed[np.logical_and(bed < 0, thk == 0)] = min_spac
+            print("Maximizing resolution in ocean. {} ocean cells \
+                  ".format(np.sum(np.logical_and(bed < 0, thk == 0))))
+        else:
```
trhille: This should not be within an else-block, or we will not be able to set maximum resolution in the ocean _and_ have bed topography influence resolution for the ice sheet.

PR #899, opened 2025-04-02
github action for running a coarse gis meshgen test

cwsmith:
This PR adds a github action workflow for running a coarse GIS meshgen case on a 'free' github hosted runner.  The run takes about 11 minutes.  It is only triggered when a PR, in any of its commits, changes either of the following files:

https://github.com/SCOREC/compass/blob/368a4a5b0d22344b53c08dbd2c8fbebf6b738032/.github/workflows/gis_coarse_meshgen.yml#L7-L8

```
    - compass/landice/mesh.py
    - compass/landice/tests/greenland/mesh.py
```

Testing was done here by pushing changes to `compass/landice/tests/mesh.py`.  Those changes were reverted, but the action log is available here: https://github.com/MPAS-Dev/compass/actions/runs/14233471393/job/39888454918

As noted below in the checklist, this test is currently downloading the GIS input data from a public RPI box directory.  Is there a better place to host these files?

Checklist
* [X] Document any testing that was used to verify the changes
* [ ] Find a better home for the GIS input data.

### DIFF
--- a/.github/workflows/gis_coarse_meshgen.yml
+++ b/.github/workflows/gis_coarse_meshgen.yml
@@ -23,16 +23,16 @@ jobs:
 
           # create the compass environment
           ./conda/configure_compass_env.py --conda $root/mambaForge --mpi mpich
-          
+
           # enable the compass environment
           source load_dev_compass_1.7.0-alpha.1_mpich.sh
-          
+
           popd
 
           # download the input GIS coarse mesh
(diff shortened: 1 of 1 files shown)


xylar (approved):
Great, thanks @cwsmith!  Let's give this a try!

xylar:
> Find a better home for the GIS input data.

It could potentially be hosted on the LCRC server.  We've tried to be careful not to overload the server but this wouldn't be big and wouldn't run all that often.

PR #900, opened 2025-05-01
Update namelist file to include new&missed options

hollyhan:
This PR updates the MALI namelist file in the `ismip6_run` test case to include options for new model developments (e.g., the Runge-Kutta time integration method) that have been added since the test case was originally merged.

### DIFF
--- a/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/namelist.landice
+++ b/compass/landice/tests/ismip6_run/ismip6_ais_proj2300/namelist.landice
@@ -5,6 +5,10 @@
 
     config_thickness_advection = 'fo'
     config_tracer_advection = 'fo'
+    config_horiz_tracer_adv_order = 3
+    config_advection_coef_3rd_order = 0.25
+    config_restore_thickness_after_advection = .false.
+    config_zero_sfcMassBalApplied_over_bare_land = .true.
 
     config_uplift_method = 'none'
     config_slm_coupling_interval = 5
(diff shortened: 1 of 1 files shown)


hollyhan:
@matthewhoffman & @trhille, I got to visit this while setting up new set of ISMIP6 Antarctica 2300 simulations. Let me know if there are other options to include (or remove) based on recent MALI development that would be relevant to include in this PR.

xylar:
This looks fine to me but I'm not really qualified to review it.  @hollyhan, is there anything specific you wanted from me?  If not, I'll probably remove myself as a reviewer.

matthewhoffman:
@hollyhan , thanks for looking at this.  In general, we only include namelist options for options that are active and we want deliberately set in a way that differs from the default.  I think all of the new options you added are for features that are off in this namelist, and therefore we don't need (or probably want) them introduced to the compass test case namelist.  Did you run into any issues that led you to want to add them?

(the patch ice option for SLM makes sense  -  i was just talking about the MALI namelist options)

hollyhan:
> @hollyhan , thanks for looking at this. In general, we only include namelist options for options that are active and we want deliberately set in a way that differs from the default. I think all of the new options you added are for features that are off in this namelist, and therefore we don't need (or probably want) them introduced to the compass test case namelist. Did you run into any issues that led you to want to add them?
> 
> (the patch ice option for SLM makes sense - i was just talking about the MALI namelist options)

Thanks for clarifying, @matthewhoffman. The extra namelist options I added here were included in my original production runs for the coupled MALI-SLM paper. When setting up new runs recently, I found myself digging through my old-run folders to figure out which namelist settings I had used. Having these included in the default namelist would make reproducibility of our work easier. But if all of options I added here are disabled and thus not needed, it's probably better to clean up the namelist in my run folders rather than here (but only 'config_restore_thickness_after_advection' was turned off?) If you think none of these changes are needed, I'll create a separate PR just handling the sea level namelist.

matthewhoffman (approved):
Thanks @hollyhan , for opening this PR to fix these two important settings!  Sorry it took so long to get back to, but I'm approving it now after fixing up the namelist.landice commit.

PR #902, opened 2025-05-02
Switch CI from miniconda to micromamba

altheaden:
This PR switches CI to use the `setup-micromamba` GitHub action instead of the `setup-miniconda` GitHub action, in an effort to fix CI crashes caused by the latest `miniconda` version. 

To do this, I needed to make a change to `conda.shared`, so that it did not reference the `CONDA_EXE` variable. Apparently miniconda sets this environment var, but micromamba does not. 

Checklist

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -74,15 +74,17 @@ jobs:
 
       - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
         name: Set up Conda Environment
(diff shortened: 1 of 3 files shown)


altheaden:
## Testing
I tested this by pushing it to my fork and letting CI run. I'm waiting for the documentation to deploy to make sure that it looks fine, but it built without errors so I'm not too concerned about it.

xylar (approved):
Looks great! Thanks @altheaden!

PR #918, opened 2025-08-21
Fix inadvertent mixup of variable names in build_mapping_file()

trhille:
Fix bug introduced by commit e51fcd1 that switched the variable names `ds` and `ds_out` in several places.

Checklist


xylar:
The process_smb error looks like a problem -- a bad value that's bigger than the maximum allowed integer -- and seemingly not related to this PR.

trhille:
> The process_smb error looks like a problem -- a bad value that's bigger than the maximum allowed integer -- and seemingly not related to this PR.

Yes, I'm working on tracking that down now, but I agree that it's not related to the bug in `build_mapping_file()`.

PR #927, opened 2025-10-28
github action for running a coarse gis meshgen test: update tarball link

cwsmith:
This PR is a follow up to https://github.com/MPAS-Dev/compass/pull/899 that updates the link for downloading the tarball with the subsampled test GIS mesh.

### DIFF
--- a/.github/workflows/gis_coarse_meshgen.yml
+++ b/.github/workflows/gis_coarse_meshgen.yml
@@ -35,8 +35,9 @@ jobs:
           popd
 
           # download the input GIS coarse mesh
-          tarball=7rhrks110k80tuwrn2t9puzmgy39o3en.tgz
-          wget https://rpi.box.com/shared/static/${tarball}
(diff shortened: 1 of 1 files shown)


.github/workflows/gis_coarse_meshgen.yml
xylar: While we're cleaning things up, can we rename this to something other than `CI/CD Build Workflow`? Can we have it trigger when the workflow itself gets modified?  And can we maybe add the ability to run it manually?
```
name: GIS Coarse Mesh Generation

on:
  pull_request:
    branches: [main]
    paths:
    - compass/landice/mesh.py
    - compass/landice/tests/greenland/mesh.py
    - workflows/gis_coarse_meshgen.yml

  # Allow manual triggering of the workflow
  workflow_dispatch:

xylar:
Looks good to me. Let me know when this is ready to merge.

PR #932, opened 2026-01-14
Several fixes to global ocean meshes with MALI topography 

xylar:
This merge:
* Fixes MALI ice draft calculation for grounded ice. The ice draft is determined by flotation in floating regions but is equal to the bed elevation in grounded regions. We also include `sea_level` (currently zero) in the calculation of the ice draft.
* Fixes renormalization of MALI topography. It is first masked to below sea level, then remapped, then renormalized by the ocean fraction (area below sea level).
* Uses the correct target scrip file for topo remapping. We need to use the appropriate unsmoothed or smoothed scrip file, whereas we were previously creating a new MPAS-Ocean scrip file that was always without smoothing.

Checklist

### DIFF
--- a/compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
+++ b/compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py
@@ -319,6 +319,14 @@ class RemapMaliTopography(RemapTopography):
             ds_mali.oceanFrac +
             (1.0 - alpha) * ds_bedmachine.oceanFracObserved)
 
+        # limit land ice floatation fraction to ocean fraction -- because of
+        # machine-precision noise in the combined ocean fraciton,
+        # land ice floating fraction can exceed ocean fraction by ~1e-15
+        ds_out['landIceFloatingFracObserved'] = np.minimum(
+            ds_out['landIceFloatingFracObserved'],
+            ds_out['oceanFracObserved']
+        )
+
         ds_out['ssh'] = ds_out.landIceDraftObserved
 
         write_netcdf(ds_out, 'topography_remapped.nc')


xylar:
## Testing

I ran the `ocean/global_ocean/SOwISC12to30/MALI_topo_AIS_4to20km/mesh` test case after:
* setting `ocean_includes_grounded=True` in two places in `compass/ocean/tests/global_ocean/mesh/__init__.py`
* setting `expand_factor = 2` in `mesh.cfg`

I now see, as expected:
* the draft is exactly equal to the bed both before and after remapping of the MALI topography in fully grounded regions
* the draft is never below the bed (the water column thickness is non-negative)
* the MALI topography is masked by `oceanFrac = bed < 0`, as expected
* the remapped MALI topography is renormalized by the remapped `oceanFrac` as expected: The topography does not drop toward zero at locations where `oceanFrac < 1`
* the remapped MALI topography in the  `remap_topo_smoothed` step is, in fact, visibly smoother than the remapped topography in the unsmoothed step.
* the combined topography doesn't show any visible jumps at the boundary where `landIceFraction` has fractional values (used to interpolate between the two topography deatasets)
* the culled mesh removes masked areas of zero bathymetry and ice draft (locations where bathymetry was above sea level before masking)
* on the culled mesh, the `0.27 < oceanFracObserved < 1`; areas where `oceanFracObserved < 0.5` are few and likely are explained by topological requirements (e.g. adding cells to the ocean to prevent isolated vertices)

For paraview verification, see `xdmf` files in:
```
/lcrc/group/e3sm/ac.xylar/compass_1.9/chrysalis/test_20260113/fix-mali-draft-with-smoothing3/ocean/global_ocean/SOwISC12to30/MALI_topo_AIS_4to20km/mesh/*/xdmf_*/
```

xylar:
The values before remapping in the masked region aren't meaningful. They get culled so negative values are just irrelevant noise. So I think this is expected. I didn't see any negative results after culling in my testing, so I think this is as expected. Thanks for checking.

cbegeman:
> The values before remapping in the masked region aren't meaningful. They get culled so negative values are just irrelevant noise. So I think this is expected. I didn't see any negative results after culling in my testing, so I think this is as expected. Thanks for checking.

I'm a bit confused by your comment. I agree that the red values in the `topography_remapped.nc`'s (`landIceDraftObserved` minus `bed_elevation`) are noise, but my interpretation is that this was generated during the remapping to `topography_ncremap.nc`. And at least some of these negative values in the ice sheet interior wouldn't be culled, we would just dig the bed to accommodate the lower `landIceDraft`. I don't think it's likely to be too much lower than `bed_elevation` so I think it's practically not an issue. I was just wondering whether we might want to limit the `landIceDraft` to `bed_elevation` here https://github.com/MPAS-Dev/compass/blob/b2535794505c5f69d499317b49213cc956a95b01/compass/ocean/mesh/remap_topography.py#L316 in a similar way as you've done in `compass/ocean/tests/global_ocean/mesh/remap_mali_topography/__init__.py`

cbegeman:
Here's what I mean: this is `topography_culled.nc`

**As above, legend title is not correct, this is actually `landIceDraftObserved` minus `bed_elevation`**

image

cbegeman:
@xylar Thanks!

PR #944, opened 2026-03-21
Update to v2.0.0-alpha.1

xylar:
This pull request updates to `mache.deploy`, which uses the `./deploy.py` script instead of `./conda/configure-compass-env.py`.

It switches to using `pixi` in the background for creating environments with conda packages.

### Updates:
- esmf v8.9.1
- mache v3.6.1 -- brings in `mache.deploy`, `mache.jigsaw` and `mache.parallel` as well as module updates on many machines and several bug fixes
- moab v5.6.0
- albany tag `compass-2026-03-21`
- trilinos tag `compass-2026-02-06`

## Testing

Only testing MALI, as MPAS-Ocean is no longer being tested regularly on Compass.

MALI with `full_integration`:
- [ ] Chrysalis (@xylar)
  - [ ] gnu and openmpi - seeing #946 during MALI build
- [ ] Perlmutter (@xylar)
  - [ ] gnu and mpich - seeing #946 during MALI build
  - [ ] gnugpu and mpich - seeing #946 during MALI build

## Deployed

MALI with `full_integration`:
- [ ] Chrysalis (@xylar)
  - [ ] gnu and openmpi
- [ ] Perlmutter (@xylar)
  - [ ] gnu and mpich
  - [ ] gnugpu and mpich

PR #948, opened 2026-03-26
Speed up landice `gridded_flood_fill` with scipy.ndimage

trhille:
Speed up gridded_flood_fill with scipy.ndimage instead of nested loops. For the antarctica_8km_2024_01_29.nc, this reduced the time from about 7s to 0.005s. This could make it much more manageable to use high-resolution gridded data sets to determine cell spacing for large meshes.

Checklist

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -18,6 +18,7 @@ from mpas_tools.mesh.creation import build_planar_mesh
 from mpas_tools.mesh.creation.sort_mesh import sort_mesh
 from mpas_tools.scrip.from_mpas import scrip_from_mpas
 from netCDF4 import Dataset
+from scipy import ndimage
 from scipy.interpolate import NearestNDInterpolator, interpn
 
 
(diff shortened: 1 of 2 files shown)


trhille:
Here are results for the 3–30km GIS mesh, using the head of `main` (left) and the head of this branch (right). Results Both meshes have `nCells = 54173` and `areaCell` fields are identical between the two.  Using this branch, gridded_flood_fill took 0.0086 seconds; using `main` it took 7.55 s.

image

matthewhoffman (approved):
@trhille , this looks great, and thanks for adding the test showing no answers change.  CI is failing for some unrelated reason, so once I sort that out, I will merge.

matthewhoffman:
@xylar , it looks like the CI is failing when trying to build jigsaw.  I forced the CI to rerun and got the same error.  I'm confused because this branch is based on main, and the CI appeared to pass [on the most recent PR](https://github.com/MPAS-Dev/compass/actions/runs/24350459254).  I can dig deeper, but wanted to check if you had any insight into this, as I know it happens periodically.

xylar:
The way of building jigsaw changes (or can change) in #944.  I don't really want to debug the "old" way of building jigsaw.

matthewhoffman:
Based on group discussion, merging without CI passing.

PR #949, opened 2026-03-26
Speed up landice.mesh.preprocess_ais_data extrapolation

trhille:
Speed up extrapolation step in landice.mesh.preprocess_ais_data by replacing `scipy.interpolate.NearestNDInterpolator` with smarter algorithm using `scipy.ndimage.distance_transform_edt`.

Checklist

### DIFF
--- a/compass/landice/tests/antarctica/mesh.py
+++ b/compass/landice/tests/antarctica/mesh.py
@@ -46,8 +46,8 @@ class Mesh(Step):
         self.add_output_file(
             filename=f'{self.mesh_filename[:-3]}_ismip6_regionMasks.nc')
         self.add_input_file(
-            filename='antarctica_1km_2024_01_29.nc',
-            target='antarctica_1km_2024_01_29.nc',
+            filename='antarctica_8km_2024_01_29.nc',
+            target='antarctica_8km_2024_01_29.nc',
             database='')
 
     # no setup() method is needed
@@ -70,7 +70,7 @@ class Mesh(Step):
         section_name = 'mesh'
 
         # TODO: do we want to add this to the config file?
-        source_gridded_dataset = 'antarctica_1km_2024_01_29.nc'
+        source_gridded_dataset = 'antarctica_8km_2024_01_29.nc'
 
         if bedmachine_dataset is not None:
             bm_updated_gridded_dataset = (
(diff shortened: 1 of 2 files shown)


trhille:
I've confirmed that AIS 8–30km meshes look visually identical with these changes. Maximum differences are mostly minuscule:
```
thickness: 4.547473508864641e-13
bedTopography: 4.547473508864641e-13
iceMask: 8.881784197001252e-16
basalHeatFlux: 0.03157263926767409
observedSurfaceVelocityX: 3.3881317890172014e-21
observedSurfaceVelocityY: 1.3552527156068805e-20
observedSurfaceVelocityUncertainty: 1.0587911840678754e-22
observedThicknessTendency: 8.97833206677805e-08
thicknessUncertainty: 1.1368683772161603e-13
```
However, the maximum differences for basalHeatFlux and observedThicknessTendency are relatively much larger than for other fields. Differences appear to be along seams in the extrapolation, so I don't think this is an issue.

image
image

Left panel in each screenshot is the field on the mesh created using `main`; right panel is the absolute value of the difference between the two meshes.

compass/landice/tests/antarctica/mesh.py
```diff
@@ -46,8 +46,8 @@ def __init__(self, test_case):
         self.add_output_file(
             filename=f'{self.mesh_filename[:-3]}_ismip6_regionMasks.nc')
         self.add_input_file(
-            filename='antarctica_1km_2024_01_29.nc',
```
matthewhoffman, line 49: Did you mean to switch to a coarser dataset?  Is that accounting for any of the speedup you reported?

trhille: Yes, this was intentional, and no I don't think it accounts for the speedup I reported. This was inadvertently switched from 8km to 1km in this recent commit: https://github.com/MPAS-Dev/compass/commit/0b5223893559a321c2fe49a6a0236d86fe3c4785#diff-992c4799a18d79fe0e3d55859f4b5f326104e5af5e6b8a5b0459848aa90a84ca (and was missed because there were far too many changes in that PR). This just resets to the original 8km data set, but I have been careful to be consistent in which resolution I use between `main` and this branch.

trhille: In my most recent testing, in which I definitely used the 8km source dataset, I get `All nearest-fill preprocessing completed in 0.288 seconds` for this branch and `All interpolations completed in 11.316709512029774 seconds.` on `main`. So still a ~2 order of magnitude speedup when using a coarse data set, but not so impactful when just cutting of 11 seconds instead of 4100 seconds :).

matthewhoffman (approved):

PR #955, opened 2026-05-05
Mask SCRIP files to reduce memory footprint of ESMF_RegridWeightGen

trhille:
This merge creates a `grid_imask` field in the SCRIP files for source data sets (BedMachine and MEaSUREs) that masks out cells that do not overlap the target MALI mesh (plus a buffer with default 50km width). This reduces the active cells in BedMachine Greenland v6 by about 25%, from 1.8M to 1.4M, and makes it possible to run the default 1–10km greenland `mesh_gen` case on 8 Perlmutter CPU nodes in about 40 minutes, where it previously required 16 nodes.

Checklist

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -1395,6 +1395,127 @@ def add_grid_imask_from_dst_scrip_hull(source_scrip, dest_scrip,
         ds_out.to_netcdf(masked_source_scrip, mode='w')
 
 
+def plot_hull_diagnostic(hull_path, dest_scrip, source_bbox_files,
+                         transformer, _maybe_deg, active_xc, active_yc,
+                         logger=None):
+    """
+    Save a diagnostic PNG (``convex_hull.png``) showing the masking geometry.
+
+    Parameters
+    ----------
+    hull_path : matplotlib.path.Path
+        Buffered convex hull of the destination mesh.
+
+    dest_scrip : str
+        Destination SCRIP file (used to scatter MALI domain extent).
+
+    source_bbox_files : list of tuple
+        Each entry is ``(filepath, label, edgecolor)`` for a source dataset
+        whose bounding box should be drawn.
+
+    transformer : pyproj.Transformer
+        Transformer from lon/lat to planar coordinates.
+
+    _maybe_deg : callable
+        Helper that converts radian angles to degrees when needed.
+
+    active_xc : numpy.ndarray
+        Projected x-coordinates of all active (masked-in) source cells.
+
+    active_yc : numpy.ndarray
+        Projected y-coordinates of all active (masked-in) source cells.
+
+    logger : logging.Logger, optional
+        Logger for status messages; falls back to print if None.
+    """
+    def _log(msg):
+        if logger is not None:
+            logger.info(msg)
+        else:
+            print(msg)
+
+    try:
+        from matplotlib.patches import Polygon as MplPolygon
+        from matplotlib.patches import Rectangle
+
+        # --- destination SCRIP centers (MALI domain extent) ---
+        with xarray.open_dataset(dest_scrip) as ds_dst:
+            if ('grid_center_x' in ds_dst.variables and
+                    'grid_center_y' in ds_dst.variables):
+                xd = np.asarray(ds_dst['grid_center_x'].values)
+                yd = np.asarray(ds_dst['grid_center_y'].values)
+            else:
+                lon = ds_dst['grid_center_lon'].values
+                lat = ds_dst['grid_center_lat'].values
+                lon, lat = _maybe_deg(lon, lat)
+                xd, yd = transformer.transform(lon, lat)
+                xd, yd = np.asarray(xd), np.asarray(yd)
+
+        rng = np.random.default_rng(seed=0)
+
+        def _subsample(x, y, n=50000):
+            idx = np.where(np.isfinite(x) & np.isfinite(y))[0]
+            if len(idx) > n:
+                idx = rng.choice(idx, size=n, replace=False)
+            return x[idx], y[idx]
+
+        def _src_extent(filepath):
+            """Return (x_min, x_max, y_min, y_max) from a gridded dataset."""
+            with xarray.open_dataset(filepath) as ds:
+                if 'x1' in ds and 'y1' in ds:
+                    x, y = ds['x1'].values, ds['y1'].values
+                elif 'x' in ds and 'y' in ds:
+                    x, y = ds['x'].values, ds['y'].values
+                else:
+                    return None
+            return float(x.min()), float(x.max()), \
+                float(y.min()), float(y.max())
+
+        fig, ax = plt.subplots(figsize=(10, 10))
+
+        # active source cells — ice-sheet extent (tab:blue)
+        xs, ys = _subsample(active_xc, active_yc)
+        ax.scatter(xs, ys, s=0.5, color='tab:blue', alpha=0.4,
+                   label='active source cells', rasterized=True)
+
+        # MALI domain extent (tab:pink)
+        xs, ys = _subsample(xd, yd)
+        ax.scatter(xs, ys, s=1.5, color='tab:pink', alpha=0.6,
+                   label='MALI domain extent', rasterized=True)
+
+        # source bounding boxes
+        for filepath, label, color in source_bbox_files:
+            ext = _src_extent(filepath)
+            if ext is None:
+                continue
+            x0, x1, y0, y1 = ext
+            bbox = Rectangle(
+                (x0, y0), x1 - x0, y1 - y0,
+                linewidth=1.5, edgecolor=color,
+                facecolor='none', label=label)
+            ax.add_patch(bbox)
+
+        # convex hull (tab:green)
+        hull_verts = hull_path.vertices
+        hull_patch = MplPolygon(
+            hull_verts, closed=True,
+            linewidth=2, edgecolor='tab:green',
+            facecolor='none', label='convex hull (buffered)')
+        ax.add_patch(hull_patch)
+
+        ax.set_aspect('equal')
+        ax.set_xlabel('x (m)')
+        ax.set_ylabel('y (m)')
+        ax.legend(loc='best', markerscale=6)
+        ax.set_title('Source SCRIP masking: hull vs. domain')
+        fig.savefig('convex_hull.png', dpi=150,
+                    bbox_inches='tight', facecolor=fig.get_facecolor())
+        plt.close(fig)
+        _log('Saved masking diagnostic plot to convex_hull.png')
+    except Exception as exc:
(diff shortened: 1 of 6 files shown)


trhille:
Testing
-------
With these changes, we can create the 1–10km Greenland mesh in 37:21 and the 4–20km Antarctica mesh in 27:24 on 8 CPU nodes on Perlmutter (1024 tasks). The MALI mesh convex hull caching introduced in https://github.com/MPAS-Dev/compass/pull/955/commits/9064a8c963f20aa2e12c7dca0abcd2f3304f2e4a saves a few minutes: the 1–10km Greenland mesh took 40:09 when calculating the convex hull multiple times, but I only ran each case once. Antarctica can fit on 4 nodes (40:39 run time), but Greenland fails on 4 nodes because ESMF_RegridWeightGen runs out of memory. I haven't tested Greenland with more than 4 but less than 8 nodes.

For Greenland, these changes reduce the number of activate cells in the source SCRIP by about 25%: `active source cells after masking: 140813110 / 187459428`
For Antarctica, the savings is much larger, reducing the number of active cells in the source SCRIP by ~60%: `active source cells after masking: 70964025 / 177768889`

trhille:
Here's an example of the saving on active cells in the SCRIP file from an 8–30km Antarctica mesh (which took 22:17 on 4 Perlmutter nodes):
image

trhille:
For Greenland (3–30km), here's the convex hull approach (top) from [8c1dc31](https://github.com/MPAS-Dev/compass/pull/955/commits/8c1dc31656bfdd9a49615c022b7478ff35bf3780), versus the dilation approach (i.e., buffered boundary) approach (bottom) from 69d8b8e3d6ec1de26af1a9e417a8e0a89bac3131:
image
image

trhille:
With the changes in [69d8b8e](https://github.com/MPAS-Dev/compass/pull/955/commits/69d8b8e3d6ec1de26af1a9e417a8e0a89bac3131) I get `active source cells after masking: 62723040 / 177768889` (compared to 70964025 / 177768889 when using the convex hull approach) for the 4–20km Antarctic mesh, which took 42:24 on 4 Perlmutter CPU nodes:
image

compass/landice/mesh.py
```diff
@@ -1130,8 +1132,438 @@ def _nearest_fill_from_valid(field2d, valid_mask):
...
+            '+proj=stere +lat_ts=70.0 +lat_0=90 +lon_0=315.0 +k_0=1.0 '
+            '+x_0=0.0 +y_0=0.0 +ellps=WGS84'
+        ),
+        'antarctica': (
+            '+proj=stere +lat_ts=-71.0 +lat_0=-90 +lon_0=0.0 +k_0=1.0 '
+            '+x_0=0.0 +y_0=0.0 +ellps=WGS84'
+        ),
+    }
+    projections['gis-gimp'] = projections['greenland']
+    projections['ais-bedmap2'] = projections['antarctica']
```
matthewhoffman: This is repeated with function above.  Let's have a single location for these in a common data structure.  I could imagine them being used in other contexts.  (I know we have these already defined in MPAS-Tools, but we don't need to worry about eliminating that redundancy.)

compass/landice/mesh.py
```diff
@@ -1130,8 +1132,438 @@ def _nearest_fill_from_valid(field2d, valid_mask):
...
+        mesh_crs = projections[domain]
+
+    transformer = Transformer.from_crs(source_crs, mesh_crs, always_xy=True)
+
+    def _maybe_deg(lon, lat):
+        if (np.nanmax(np.abs(lon)) <= 2.0 * np.pi + 1.0e-6 and
+                np.nanmax(np.abs(lat)) <= 0.5 * np.pi + 1.0e-6):
+            lon = np.rad2deg(lon)
+            lat = np.rad2deg(lat)
+        return lon, lat
```
matthewhoffman: This is also duplicate and could be moved to a common location.

compass/landice/mesh.py
```diff
@@ -1130,8 +1132,438 @@ def _nearest_fill_from_valid(field2d, valid_mask):
...
+        Destination SCRIP file (used to scatter MALI domain extent).
+
+    source_bbox_files : list of tuple
+        Each entry is ``(filepath, label, edgecolor)`` for a source dataset
+        whose bounding box should be drawn.
+
+    transformer : pyproj.Transformer
+        Transformer from lon/lat to planar coordinates.
+
+    _maybe_deg : callable
```
matthewhoffman: I don't understand why a function would be passed here.  As per comment above, if `_maybe_deg` is moved to a single common function, it could just be called directly and there would be no need to pass it in here.

compass/landice/mesh.py
```diff
@@ -1130,8 +1132,438 @@ def _nearest_fill_from_valid(field2d, valid_mask):
...
+                    x, y = ds['x'].values, ds['y'].values
+                else:
+                    return None
+            return float(x.min()), float(x.max()), \
+                float(y.min()), float(y.max())
+
+        fig, ax = plt.subplots(figsize=(10, 10))
+
+        # active source cells — ice-sheet extent (tab:blue)
+        xs, ys = _subsample(active_xc, active_yc)
```
matthewhoffman: I don't think these lines are used - `xs,ys` get replaced on the next line.

compass/landice/mesh.py
```diff
@@ -1497,7 +1958,88 @@ def run_optional_interpolation(
...
+                    '+proj=stere +lat_ts=70.0 +lat_0=90 +lon_0=315.0'
+                    ' +k_0=1.0 +x_0=0.0 +y_0=0.0 +ellps=WGS84'
+                ),
+                'antarctica': (
+                    '+proj=stere +lat_ts=-71.0 +lat_0=-90 +lon_0=0.0'
+                    ' +k_0=1.0 +x_0=0.0 +y_0=0.0 +ellps=WGS84'
+                ),
+            }
+            projections['gis-gimp'] = projections['greenland']
+            projections['ais-bedmap2'] = projections['antarctica']
```
matthewhoffman: another instance that can be eliminated

compass/landice/mesh.py
```diff
@@ -1497,7 +1958,88 @@ def run_optional_interpolation(
...
+            mesh_crs = projections.get(src_proj, src_proj)
+            transformer = Transformer.from_crs(
+                'EPSG:4326', mesh_crs, always_xy=True)
+
+            def _maybe_deg_plot(lon, lat):
```
matthewhoffman: and this

compass/landice/mesh.py
```diff
@@ -1497,7 +1958,88 @@ def run_optional_interpolation(
...
+            bm_stem = os.path.splitext(
+                os.path.basename(first_src))[0]
+            # Try to find the masked SCRIP that was written to workdir
+            import re as _re
+            match = _re.search(r'(^.*[_-]v\d*[_-])+', bm_stem)
+            if match:
+                scrip_stem = bm_stem[:match.end() - 1]
+            else:
+                scrip_stem = bm_stem
+            masked_scrip = f'{scrip_stem}.scrip_masked.nc'
```
matthewhoffman: Is there a cleaner way to do this?

compass/landice/mesh.py
```diff
@@ -1497,7 +1958,88 @@ def run_optional_interpolation(
             interp_gridded2mali(self, measures_dataset, dst_scrip_file,
                                 parallel_executable, nProcs,
                                 mesh_filename, src_proj,
-                                variables=measures_vars)
+                                variables=measures_vars,
+                                hull_path=hull_path)
+
+        # Diagnostic plot: show hull, MALI domain, and bounding boxes for
+        # all source datasets that were interpolated.
```
matthewhoffman, line 1969: Would it make sense to move the rest of these changes into the plotting function?  I had the sense `run_optional_interpolation` was meant to be a high level driver calling functions for chunks of work, but the remainder of the additions below this point are in the weeds of setting up the plot.  And given the plotting function is only called in one place, conceptually it would all these details could be pushed into that function instead of living here.

matthewhoffman (changes requested):
@trhille , this looks like a really nice performance optimization.  I read all the test results and skimmed all the changes. The general idea and the specific implementation make sense.  I have requested some changes to avoid code duplication and simplify organization.  I may be missing reasons why things are the way they are, though, so if these suggestions aren't practical, feel free to say so.

compass/landice/mesh.py
trhille: fixed in [77413d4](https://github.com/MPAS-Dev/compass/pull/955/commits/77413d4114e3de1b88cc3201f4ef3db43a6c70fb)

compass/landice/mesh.py
trhille: fixed in [77413d4](https://github.com/MPAS-Dev/compass/pull/955/commits/77413d4114e3de1b88cc3201f4ef3db43a6c70fb)

compass/landice/mesh.py
trhille: fixed in [77413d4](https://github.com/MPAS-Dev/compass/pull/955/commits/77413d4114e3de1b88cc3201f4ef3db43a6c70fb)

compass/landice/mesh.py
trhille: fixed in [77413d4](https://github.com/MPAS-Dev/compass/pull/955/commits/77413d4114e3de1b88cc3201f4ef3db43a6c70fb)

compass/landice/mesh.py
trhille: fixed in [77413d4](https://github.com/MPAS-Dev/compass/pull/955/commits/77413d4114e3de1b88cc3201f4ef3db43a6c70fb)

compass/landice/mesh.py
trhille: fixed in af5d55b84

compass/landice/mesh.py
trhille: Addressed in 5e0a04630.

compass/landice/mesh.py
trhille: Removed in https://github.com/MPAS-Dev/compass/commit/5e0a0463050a87829912a6df3942a8707c79dbd3.

trhille:
Greenland 3–30km mesh took 17:30 on 1 Perlmutter CPU node as of [5e0a046](https://github.com/MPAS-Dev/compass/pull/955/commits/5e0a0463050a87829912a6df3942a8707c79dbd3).
BedMachine: `active source cells after masking: 130882695 / 187459428`
Measures: `active source cells after masking: 11705489 / 16434600`
image

trhille:
docs are at https://portal.nersc.gov/cfs/fanssie/trhille/compass-docs-PR955/html/

matthewhoffman (approved):
@trhille , thanks for updating this.  Looks good to me!

PR #962, opened 2026-05-18
Split string to interpret "srun --label" correctly

trhille:
Split string "srun --label" so that it is interpretted correctly when passed to check_call.

Checklist

### DIFF
--- a/compass/landice/mesh.py
+++ b/compass/landice/mesh.py
@@ -1648,13 +1648,17 @@ def interp_gridded2mali(self, source_file, mali_scrip, parallel_executable,
 
     # Generate remapping weights
     logger.info('generating gridded dataset -> MPAS weights')
-    args = [parallel_executable, '-n', nProcs, 'ESMF_RegridWeightGen',
-            '--source', masked_source_scrip,
-            '--destination', mali_scrip,
-            '--weight', weights_filename,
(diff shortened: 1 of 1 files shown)


trhille:
Testing
-------
After updating to `compass 2.0.0-alpha.1`, I got this error when running the `landice/antarctica/mesh_gen` case:

```
Running: srun --label -n 128 ESMF_RegridWeightGen --source BedMachineAntarctica_2020-07-15_v02.scrip_masked.nc --destination Antarctica_scrip.nc -
-weight gridded_to_MPAS_weights.nc --method conserve --netcdf4 --dst_regional --src_regional --ignore_unmapped
      Failed
Exception raised while running the steps of the test case
...
FileNotFoundError: [Errno 2] No such file or directory: 'srun --label'
```
It seems that `check_call()` interpreted `srun --label` as a single string instead of instead of running `srun` with `--label` as a separate argument.

With these changes, the `srun` command is being interpreted correctly: the 8–40km `antarctica` case finished successfully in 34:24 , and  a `humboldt` case ran successfully in 10.5 minutes.

Note: CI is currently failing for the Greenland mesh_gen test due to maintenance on LCRC server.

matthewhoffman (approved):

PR #965, opened 2026-06-10
Update to v2.0.0-alpha.2 and mache v3.7.1

xylar:
This mache version brings in https://github.com/E3SM-Project/mache/pull/437, which fixes version detection and the location of pixi cache files.

Checklist


xylar:
## Testing

I was able to create load script with this version on my laptop with this version, and tested deployment on Perlmutter with this mache version to prove that the version detection and cache location issues are fixed there.

PR #972, opened 2026-07-25
Fix base mesh plotting with cartopy 0.25 and matplotlib 3.11

xylar:
Creating gridline labels with draw_labels=True and then hiding some of them (gl.top_labels = False, etc.) confuses cartopy's title placement. GeoAxes._update_title_position() scans gl.top_label_artists whenever gl.geo_labels is true (the default), so the hidden top labels are still included.  During a draw, their tight bounding boxes are infinite, so the title ends up at y = inf and the axes tight bounding box becomes non-finite.

The NaN then cascades: tight_layout() gives the axes a NaN position, the "geo" spine path becomes all NaN, and cartopy's gridliner hands NaN vertices to shapely, which raises

    GEOSException: IllegalArgumentException: Points of LinearRing do not
    form a closed linestring

Instead, only request the labels we actually want.  In _plot_cell_width(), that means draw_labels=['left', 'bottom'].  In the arrm10to60 and wc14 mesh plots, all four label sets were hidden, so draw_labels=False.

Also drop the redundant fig.canvas.draw() and plt.tight_layout() from _plot_cell_width(), since bbox_inches='tight' already crops the figure. Removing tight_layout() alone is not sufficient: without the label fix, the title is silently dropped and bbox_inches='tight' crops to a garbage bounding box, saving only the colorbar.

List-valued draw_labels requires cartopy 0.22, so add that constraint.

Along the way, satisfy the pre-commit hooks on the arrm10to60 file: isort reorders its imports and flake8 flags a pre-existing F523 (a '{0} x {0}'.format(dlon, dlat) that never used dlat), now an f-string.

This merge also fixes an unrelated provenance issue caused by the transition from conda to pixi.

Checklist

### DIFF
--- a/compass/mesh/spherical.py
+++ b/compass/mesh/spherical.py
@@ -142,29 +142,29 @@ class SphericalBaseStep(Step):
         image_filename = config.get('spherical_mesh',
                                     'cell_width_image_filename')
         register_sci_viz_colormaps()
-        fig = plt.figure(figsize=[16.0, 8.0])
+        plt.figure(figsize=[16.0, 8.0])
         ax = plt.axes(projection=ccrs.PlateCarree())
         ax.set_global()
         im = ax.imshow(cell_width, origin='lower',
                        transform=ccrs.PlateCarree(),
                        extent=[-180, 180, -90, 90], cmap=cmap, zorder=0)
         ax.add_feature(cartopy.feature.LAND, edgecolor='black', zorder=1)
-        gl = ax.gridlines(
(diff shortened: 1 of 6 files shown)


xylar:
## Testing

I'm running the EC30to60 grid through its paces in:
```
/lcrc/group/e3sm/ac.xylar/compass_2.0/chrysalis/test_20260725/ec30to60_2
```

The `mesh`, `WOA23/init` and `WOA23/performance_test` tests passed.

PR #975, opened 2026-08-13
Add ISMIP7 fracture test case for shelf-collapse pathways

trhille:
Add a 'fracture' test case to the ismip7_forcing test group that processes the ISMIP7 surface-melt-driven ice shelf collapse forcing (AIS) via three pathways, each as its own step:

- process_excess_melt (Path A): remaps the excess meltwater field. Reconstructs x/y coordinates and corrects the y-axis orientation of the source file before conservative remapping.
- process_lake_properties (Path B): remaps supraglacial lake mean depth and area fraction (Grau et al. 2025) via bilinear remapping.
- process_shelf_collapse (Path C): remaps the annual ice shelf collapse mask via neareststod, rounding to 0/1.

Add per-pathway remap-method config options and documentation.

Checklist
* [x] Documentation has been [built locally](https://mpas-dev.github.io/compass/latest/developers_guide/building_docs.html) and changes look as expected: https://portal.nersc.gov/cfs/fanssie/trhille/compass-docs-PR975/html/

### DIFF
--- a/compass/landice/tests/ismip7_forcing/__init__.py
+++ b/compass/landice/tests/ismip7_forcing/__init__.py
@@ -1,4 +1,5 @@
 from compass.landice.tests.ismip7_forcing.atmosphere import Atmosphere
+from compass.landice.tests.ismip7_forcing.fracture import Fracture
 from compass.landice.tests.ismip7_forcing.ocean_thermal import OceanThermal
 from compass.testgroup import TestGroup
 
@@ -22,3 +23,4 @@ class Ismip7Forcing(TestGroup):
 
         self.add_test_case(Atmosphere(test_group=self))
         self.add_test_case(OceanThermal(test_group=self))
+        self.add_test_case(Fracture(test_group=self))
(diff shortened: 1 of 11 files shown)


trhille:
Testing
-------
Processing these files took less than one minute on one Perlmutter node.
CESM2-WACCM ssp5-8.5 shelf collapse mask:

https://github.com/user-attachments/assets/778947d0-d5ba-4d4a-a423-3532bf422900

CESM2-WACCM ssp5-8.5 lake depth and area:

https://github.com/user-attachments/assets/3c5ce9ce-6f34-485a-af45-bf8619bd7b0f

CESM2-WACCM ssp5-8.5 excess surface melt:

https://github.com/user-attachments/assets/b6bb1162-c8a8-4f2a-b1dc-0d15e9704538

matthewhoffman (approved):
I reviewed the docs, cfgs, and all python files by inspection.  I saw a handful of places where perhaps refactoring could be applied, performance improvements might be available, or minor convenience improvements, but I don't think those are worth pursuing for this specialized tool.  One convenience thing worth mentioning is it is not clear to me how the new fracture steps are handled if the ice sheet is set to gis - it looks like it would be necessary to set the methods to 'none' to avoid an error, but it would be convenient if these steps were just ignored for gis.  I'll leave it up to you if you want to bother implementing that or not - this certainly seems like it gets the job done.

PR #982, opened 2026-09-09
Add an ISMIP7 melt-module calibration test group

xylar:
This adds a `landice/ismip7_calibration` test group that calibrates MALI's sub-shelf melt parameterization against the ISMIP7 Antarctic ice-ocean protocol (Reese et al., Sect. 4.2), producing the 5th / 50th / 95th percentiles of the melt parameter and a MALI parameter file for each. `replication` reproduces the published 8 km calibration in about a minute with no MALI run; `ais` does the calibration on a MALI mesh.

Worth a reviewer's attention:
* One MALI run per ocean state, not per parameter value: melt is exactly proportional to the parameter, so the sweep is a scaling of a single run.
* MALI runs with a constant salinity, because the ISMIP7 forcing already processed for the projections carries none. This needs MALI-Dev/E3SM#191, so the `MALI-Dev` submodule moves to that merge.
* Code shared by the ISMIP7 test groups moves out of `ismip7_forcing` into a new `compass/landice/ismip7` framework package. Meant to be behaviour-preserving.
* Compass's first unit tests, under `tests/`, run by `pytest` in CI.

The objective function is vendored verbatim from `ismip7-antarctic-ocean-forcing`, which has no installable package, with its upstream commit and SHA256 recorded and checked.

---

*Posted by Claude Code on @xylar's behalf. The testing, analysis and wording above are AI-authored; please check them accordingly.*


xylar:
## Testing

All of the below was run on Chrysalis, with MALI built from `xylar/E3SM` branch `mali/add-burgard-melt-param` (commit `d57c696`), against the ISMIP7 AIS datasets at `/lcrc/group/e3sm/ac.xylar/ismip7/forcing-data/data/AIS/` and the 4-20 km mesh `ais_4to20km.20250625.nc`. Work directories are under `$COMPASS_SCRATCH/ismip7_calibration`.

### Unit tests

`pytest` — 95 tests, 19 s, all passing. These are what CI will run.

### `landice/ismip7_calibration/replication`

PASS in 58 s on one core. Reproduces the published 8 km percentiles exactly:

| percentile | compass | published |
|---|---|---|
| 5th | 4.75000e-05 | 4.75e-05 |
| 50th | 8.50000e-05 | 8.5e-05 |
| 95th | 1.37500e-04 | 1.375e-04 |

The Antarctic-mean ice-draft slope comes out at sin(theta) = 0.0051117, matching the protocol's stated 0.005.

### `landice/ismip7_calibration/ais`

Full ensemble: 56 single-timestep MALI runs (28 ocean states x 2 melt forms) at 128 tasks, plus 2 extra runs for the linearity check. Each run takes about 2 minutes; the ensemble was run as 8 batches of 7 and completed in about 6 minutes wall clock.

**Masks.** The remapped ISMIP7 basins agree with the mesh's existing `regionCellMasks` for 98.7% of all cells and 99.1% of floating cells. The one basin that stands out is the Antarctic Peninsula at 82.5% in the theirs-to-ours direction, which is a genuine difference between the ISMIP6-2022 and IMBIE2-v3 outlines rather than an error, and matches what the prototype found.

**Verification against an independent Python implementation**, on the present-day climatology, over all 109,412 melting cells:

| check | result |
|---|---|
| melt expression, on MALI's own TFdraft | max relative difference 6.05e-16 |
| vertical interpolation, all four code paths | max difference exactly 0.0 K |
| linearity in the melt parameter, measured in MALI | max relative deviation exactly 0 |

The linearity runs give total melt of 892.7163, 1785.4325 and 3570.8650 Gt/yr at 0.5, 1 and 2 times the reference parameter; dividing by the multiple gives 1785.432511 in every case.

**Calibration result**, 28 ocean states, full weighting, 100,000 samples:

| melt form | 5th | 50th | 95th |
|---|---|---|---|
| ismip7 (K) | 4.75e-05 | 7.75e-05 | 1.15e-04 |
| ismip6 (gamma0, m/yr) | 6750 | 11000 | 19500 |

**dT_b fit.** All 16 basins fitted, with the largest correction 1.0 K in basin 9 (Eastern Amundsen), comfortably inside the +/-2 K bound the protocol recommends.

**Shelf area.** MALI's total ice-shelf area is 1510 against an observed 1538 x 10^3 km^2, a ratio of 0.98, with per-basin ratios between 0.53 and 1.05.

### Comparison against the prototype

The prototype that produced the science result reported K = 5.00e-5 / 7.75e-5 / 1.15e-4 on the same mesh, computed with a spatially varying salinity. The median and 95th percentile here match exactly and the 5th is one grid step lower, which is the size of shift expected from moving to a constant salinity. The melting-cell count, 109,412, matches the prototype exactly.

~~The ISMIP6 numbers do not line up as closely: the prototype reported gamma0 = 7453 / 12200 / 21000 against 6750 / 11000 / 19500 here.~~ **Superseded.** The gamma0 grid stopped at 30,000 m/yr and was truncating a real tail. With the grid widened it gives 7250 / 13000 / 21500 on this mesh, against the prototype's 7453 / 12200 / 21000. See the `More Testing` comment below; no science review is needed for this after all.

### Bugs this testing caught

Three problems were found by running against real data rather than by inspection, all of them in this PR's code rather than in MALI:

1. The Python reference applied the freezing-point depth correction in only one of the two branches where MALI extrapolates downward, disagreeing by up to 3.4 K over 11,000 cells.
2. `pyremap`'s `Remapper` invokes `mpirun` directly, which conflicts with the Slurm allocation; the mask remapping now goes through the framework's `build_mapping_file`, which uses the configured `parallel_executable`.
3. The `cellMask` written to the melt output is post-step, and over one timestep about 5,000 ice-free cells pick up a trace of ice. Treating those as floating compared against values MALI never computed, and added zero-melt area to the denominator of the area-weighted basin means J3 is built from. The contributing cells now come from the initial geometry.

The first and third only became visible once the verification ran against a real ensemble, which is the argument for having the verification step at all.

### Not tested

The `ismip7_forcing` consolidation in the first commit is intended to be behaviour-preserving, but I did not run `ismip7_forcing` before and after to confirm byte-identical output; it needs the CMIP scenario forcing, which is not staged on this machine. The changes are mechanical (moved modules, updated imports, and one `decode_times` argument added so that each existing caller keeps its previous behaviour), but a reviewer who can run that test group should confirm it.

GrIS was not exercised at all.


### Parameter distributions

image

The distribution of the optimal `K` over the 100,000 draws, for the ISMIP7 local form — the one the protocol recommends. It is bimodal: a narrow spike at the mode, `K` = 5.25e-05, and a broad hump centred near 8-9e-05. That structure is a property of the protocol's objective rather than of MALI: the four terms disagree about the best parameter, so draws dominated by one pair of terms land somewhere different from draws dominated by the other. The prototype found the same shape. The whole distribution sits well inside the parameter grid, which runs to 3.0e-04, with no draws at either end, so nothing is being clipped.

image

The same for the ISMIP6 non-local form, which with a constant salinity is the Burgard semi-local form. Bimodal again, with a sharp spike near `gamma0` = 7250 m/yr and a long tail out to about 25,000. Note the small bar hard against the right edge: 104 of the 100,000 draws (0.10%) are pinned at the grid maximum of 30,000 m/yr, so the grid is slightly too narrow for this form. At 0.1% it cannot move the 95th percentile of 19,500, but the upper bound should be raised before these numbers are used for anything, and it is one candidate explanation for why the ISMIP6 result does not reproduce the prototype as closely as the ISMIP7 result reproduces its counterpart.

The two panels are not comparable to each other by the height or position of the objective, only by the parameter each selects — the objective normalises every term by its own median over the parameter ensemble, so its minimised value means nothing across melt forms.


This comment was written by Claude Code.

xylar:
Here's the corresponding branch to target MALI-Dev:
https://github.com/xylar/E3SM/tree/mali/add-burgard-melt-param

See:
https://github.com/MALI-Dev/E3SM/pull/191

xylar:
## More Testing

I reran the calibration on the real initial condition, `AIS_4to20km_r03_20260905.nc` (511,727 cells), on Chrysalis with gnu and openmpi:
```
/lcrc/group/e3sm/ac.xylar/compass_2.0/chrysalis/test_20260907/ismip7_calibration_r03
```

| 5th / 50th / 95th | r03 mesh | inputdata 4-20 km mesh |
|---|---|---|
| `K` | 4.50e-05 / 7.25e-05 / 1.075e-04 | 4.75e-05 / 7.75e-05 / 1.15e-04 |
| `gamma0`, m/yr | 7250 / 13250 / 21000 | 7250 / 13000 / 21500 |

Verification is unchanged on the new mesh: melt matches the Python reference to 6.0e-16, the vertical interpolation matches exactly over all 108,504 melting cells, and melt is exactly proportional to `K`. Modelled ice-shelf area is 0.99 of observed against 0.98 before, and the fitted `dT_b` stays within 1.1 K. The whole test case also passed end to end from a single `compass run` on a 7-state subset, in 6:42.

The `gamma0` grid stopped at 30,000 m/yr, which truncated a real tail. Raising it to 50,000 is what moved the old-mesh `gamma0` from 6750 / 11000 / 19500 to the numbers above, and it accounts for most of the gap against the prototype's 7453 / 12200 / 21000 that I flagged in the first `Testing` comment. The calibrate step now fails when draws pile up at the end of the grid.

Two smaller fixes the new mesh needed: `build_mapping_file` copied the mesh with `shutil.copy`, so a read-only mesh gave a read-only copy and the step failed; and the basin cross-check needs an ISMIP6-era region mask, which a new mesh has none of, so it silently skipped. There is now a check needing no extra input, that Pine Island and Dotson fall in ISMIP7 basin 9, which passes at 100% of 943 cells.

image

`K` on the r03 mesh. Bimodal, as on the old mesh and in the prototype: the four terms disagree about the best parameter, so draws dominated by different pairs of them land in different places.

image

`gamma0` on the r03 mesh, with the grid maximum raised to 50,000 m/yr. The tail now ends near 33,000 instead of piling up against the old 30,000 ceiling.

---

*Posted by Claude Code on @xylar's behalf. The testing, analysis and wording above are AI-authored; please check them accordingly.*

xylar:
Two follow-ups to the `dT_b` fit. It is now fitted separately at the 5th, 50th and 95th percentiles, writing a complete MALI parameter file for each — basins, corrections and the parameter itself — since `dT_b` depends on the parameter and projections run at all three. And with MALI-Dev/E3SM#191 now reading `K` from the parameter stream like `gamma0`, that file is drop-in for either melt form; each melt-diagnostic run writes a small `melt_parameter.nc` rather than setting `K` in the namelist. `validate()` compares the fitted corrections and parameter files against a baseline, and the per-basin search is a plain function with 11 new tests (112 total, passing in CI).

Tested by restaging the ISMIP7 inputs from Globus and rerunning everything on Chrysalis (gnu, openmpi): the replication (PASS, published percentiles exact), the full 58-run r03 ensemble, and the whole test case end to end from one `compass run` on a 7-state subset (PASS, 5:41):
```
/lcrc/group/e3sm/ac.xylar/compass_2.0/chrysalis/test_20260915/ismip7_calibration_r03
/lcrc/group/e3sm/ac.xylar/compass_2.0/chrysalis/test_20260915/ismip7_calib_e2e
```
All 58 melt fields are bit-for-bit identical to last week's, and the percentiles are unchanged: `K` = 4.50e-05 / 7.25e-05 / 1.075e-04, `gamma0` = 7250 / 13250 / 21000.

The three-percentile fit is doing real work: in the Eastern Amundsen, `dT_b` moves by 1.2 K between the 5th and 95th percentile fits, so a single median fit would have been wrong at the others. Every correction stays within the ±2 K bound (largest 1.8 K, at the 5th percentile), and each fit cuts the summed J1 misfit by about 95%, for example 795 to 27 Gt/yr at the median for the ISMIP7 form.

---

*Posted by Claude Code on @xylar's behalf. The testing, analysis and wording above are AI-authored; please check them accordingly.*

trhille:
@xylar, I had a stab at resolving the merge conflicts on my own fork: https://github.com/trhille/compass/tree/add-ismip7-calibration

xylar:
Rebased onto `main` at 448272b, starting from @trhille's resolution (thanks!). One thing needed changing in it: since this branch was written, `main` gave `extrapolate_source` two fixes — a fill value for slabs with no valid cell, and CDF-5 intermediate files so `ncremap` can open large 3-D fields — and the rebase had left seven `ismip7_forcing` callers on the framework copy without them. The framework copy now is `main`'s, the duplicate is gone, and both fixes have tests.

The `MALI-Dev` submodule moves to 013c387, the merge of MALI-Dev/E3SM#191, so `[paths] mpas_model` no longer needs to point anywhere. MALI now stops if there is no stream named `output` (it records `fem_extension_method` there), so the melt diagnostic's output stream is renamed; the file is unchanged.

## Testing

I ran the `ais` test case end to end on a 7-state subset on Chrysalis (gnu, openmpi), with MALI built from the submodule:
```
/lcrc/group/e3sm/ac.xylar/compass_2.0/chrysalis/test_20260916/ismip7_e2e_rebased
```
PASS in 6:08. The remapped masks and forcing, all 9 melt fields and the percentiles are bit-for-bit with yesterday's run on the pre-rebase branch. 114 unit tests pass and the docs build clean.

---

*Posted by Claude Code on @xylar's behalf. The testing, analysis and wording above are AI-authored; please check them accordingly.*

trhille:
Testing
-------
I tested this on Chrysalis with `ocean_state_subset = all` and `melt_forms = ismip7, ismip6`. All steps passed successfully in 37 minutes.
I used the `melt_params_ismip7_p50.nc` file to set up a test OCX run with the current 4–20km MALI mesh (AIS_4to20km_r03_20260915.nc). It runs successfully (when I update namelist and streams files to use the new ISMIP7 melt scheme and parameter). However, the melt rates look quite low compared with observations for most regions, and extremely low for FRIS:
image

These melt rate observations are quite old (Rignot et al., 2008 and 2013), but this discrepancy seems too large to be an issue with the observations, especially at FRIS. @xylar and @matthewhoffman, what do you think?

xylar:
@trhille, can you give me the path to your test results?

trhille:
@xylar, the compass run is on Chrysalis, here: `/lcrc/group/e3sm/ac.trhille/ismip7/test_compass_pr982`
The MALI run is on Perlmutter: `/pscratch/sd/t/trhille/ISMIP7/AIS_forcing/test_COMPASS_PR982/test_ocx_run/landice/ismip7_run/ismip7_ais/ocx`

xylar:
@trhille, thanks for running this. Your calibration reproduces mine exactly (identical parameter files and percentiles), and your OCX melt rates match what the calibration produced *after* the `dT_b` fit to within a median 3% per basin — so `dT_b` is in place and doing its job. The gap is the observations: the protocol's targets are the Paolo/Davison/Adusumilli IMBIE2 mean, which total 1067 Gt/yr against about 1317 for Rignot 2008/2013, and 64 Gt/yr for FRIS against 155. The calibration hit 1070 and cannot also hit Rignot.

| region | ISMIP7 target | calibration, after `dT_b` | your run (read from the figure) | Rignot |
|---|---|---|---|---|
| FRIS | 64 ± 77 | 68 | ~60 | 155 |
| Thwaites/PIG | 244 ± 103 | 247 | ~240 | 290 |
| Ross | 68 ± 119 | 62 | ~60 | 70 |
| Mertz | 22 ± 6 | 23 | ~12 | 23 |
| all 16 basins | 1067 | 1070 | ~1024 | ~1317 |

Your run is within the target uncertainty in 15 of 16 basins; the per-basin numbers are in `fit_delta_t/delta_t_ismip7.nc` in your run directory.

Three questions:
* Mertz is the one basin well off, 12 against 23. Which OCX variant and years did you average, and is its thermal forcing at the Mertz draft much colder than the Zhou climatology the calibration used?
* Could you re-plot against `parameterisations/ocean/meltobs/Melt_Paolo_Davison_Adusumilli_imbie2.csv`, or post the `regionalStats` numbers? Reading your figure by eye is the weakest link in the table above.
* FRIS at 64 versus 155 is a real disagreement between the newer products and Rignot 2013, which the calibration cannot resolve. If you think the target is wrong, that is a question for the focus group rather than for `dT_b`.

---

*Posted by Claude Code on @xylar's behalf. The testing, analysis and wording above are AI-authored; please check them accordingly.*

trhille:
@xylar, 
• I'm using the time-varying (i.e., not time-averaged) OCX_main forcing starting in 2008. I'll have to look at the TF in OCX_main relative to Zhou.
• Here's an ncdump of  `regionalSumFloatingBasalMassBal` in kg yr^-1 from the last time step of 2008.
```
  -52154483850156.3, -10060401594795.8, -21992591993014.9, -63154491103599.5, 
    -86472321491954.2, -12399537052373.7, -12809446213527.5, 
    -59817247101279.4, -136666911301573, -240207464593289, -63984738588218.1, 
    -121975300250256, -31084059213381.6, -14885169060455.8, 
    -59392261877585.9, -33600633014569.4 ;
```

• I have confirmed that Paolo et al and Adusumilli et al report about 75 Gt/yr and 54.8 Gt/yr for FRIS, while Rignot et al. (2013) report 150 Gt/yr. So, while that huge difference is surprising, it is indeed a real difference between the older and newer data sets.

trhille:
Here is the OCX_main forcing (left) and the zhou et al climatology (right) at z=-390m. The Mertz basin is outlined in white and the floating ice is outlined in cyan:
image

I don't understand enough about how the OCX forcing was created to determine why these are different, but it certainly looks like the OCX forcing is cooler on average than the climatology forcing for this basin.
Here is the absolute difference between the two fields (note log scale):
image


This difference is inherited from the source data sets. Here is the difference between the Zhou climatology and the last time level of the OCX_main thermal forcing at z=-390:
image
I used:
`ncks -d time,-1 tf_AIS_OCX_ocean_main_v1_1950-2025.nc tmp.nc; ncdiff tmp.nc /global/cfs/cdirs/m4288/users/trhille/ISMIP7/forcing/AIS/obs/zhou_annual_06_nov/tf/v3/tf_AIS_obs_ocean_climatology_zhou_annual_06_nov_v3_1972-2024.nc diff.nc`

I raised this issue in ISMIP Discussions: https://github.com/orgs/ismip/discussions/48

trhille (approved):
@xylar, thanks for this _amazing_ work. The only issue I can find seems to stem from the source data sets (see my previous comments), so I'm approving this.

matthewhoffman:
@xylar , I added the ability to calculate a local slope, but did it in a different way before that should be less prone to local positive-feedback (smoother slope by fitting a plane to each cell + 3 rings of cells, rather than calculating edge slopes and averaging).  It looks promising: https://github.com/MALI-Dev/E3SM/pull/195#issuecomment-5747339699  But see the comment following that one - it leads to unrealistically large melt rates at TG/PIG and Totten/MU.  Presumably recalibrating the regional deltaT would compensate for that, but we'd still want to see if the melt profiles have better or worse variation with depth and/or distance from GL.  I am not convinced that we *should* use this slope-dependence, but we have it available in our arsenal if we need it to avoid GL advance at ice streams (where the no-slope version tends to have melt rates that are too low).

